PR #56, opened 2017-09-14
Add a BMX-managed creds file

chris-redekop:
An early preview PR.  It's definitely not ready for approval, but I want to get the code out there for feedback.

Currently we just support a 'default' set of creds... I'll be adding app/role sections next, then refactoring and tests.  The file looks like the following (it's auto-serialized YAML).

```
credentials:
  default: {AccessKeyId: AKIA..., Expiration: !!timestamp '2017-09-14
      01:06:43', SecretAccessKey: ..., SessionToken: ...}
```


mtjandra (approved):
Looks good with the understanding that the `--profile` and `--role` options will temporarily stop working when there is a valid `~/.bmx/credentials` file with valid STS keys (since it always defaults to `default`)

```sh
# account and role are ignored
bmx aws --account=Lrn-NimbusTorontoe --role=Lrn-NimbusToronto-User cloudformation describe-stacks
```

PR #182, opened 2019-12-18
GitHub Actions for building & testing

jrbeverly:
This replaces the building, testing and vetting done by TravisCI with GitHub Actions. I am keeping TravisCI in place, and will deprecate it when all the desired actions have been merged in (releases / fmt / etc). 

This PR adds in github actions to replace TravisCI as the continuous integration provider. This allows us to have cross-platform testing on both windows and linux environment, automated github releases, conditional regex execution, and CI on forks.

Switching to GitHub Actions will make it a bit easier for us to cut releases, as well as make forks a lot more contribution friendly (see my own jrbeverly/bmx).  I'll be looking to raise a pull request for releases after this, as well as `go fmt` enforcement.


.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
+name: Build & Test
+on:
+  push:
+    paths:
+      - "*.go"
+      - "*/*.go"
+      - "*/*/*.go"
+      - "*/*/*/*.go"
+      - "*/*/*/*/*.go"
+      - "*/*/*/*/*/*.go"
```
andrewalkema: Could we just do `"**.go"`?

jrbeverly: I'll have to check, but if I recall there was an issue with it not resolving correctly.

PR #185, opened 2019-12-23
Replace TravisCI with GitHub Actions

jrbeverly:
This completes the work for replacing TravisCI with GitHub Actions. There may be some work on the TravisCI end of things to fully break the connection, but this should resolve all of the code work related to it.

This PR removes the files for TravisCI and references to TravisCI. Moving forward, the CI workflows exist in .github/workflows, and the badges are built into the repository itself.


andrewalkema:
Nice job, thanks.

PR #190, opened 2020-06-19
Initial work for rewriting BMX in .NET Core

boarnoah:
For Inspiration June 2020.

Quick overview of reasoning:
Brings code more in line with D2L's core langs (increases chance this project can be maintained going forwards)

It's been through one rewrite to golang already (from python) for valid reasons, hopefully this is a equally valid reason as well.

==

Fairly large undertaking (LoC wise), will be split over a series of PRs into this dotnet branch prior to being placed on master.

PS: Will be creating a project under the projects tab to track further goals for release 2.1

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,104 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = tab
+indent_size = 4
+
+[{*.json,.*.yml}]
+indent_style = space
+indent_size = 2
+
+[*.cs]
+tab_width                = 4
+max_line_length          = off
+
+# Yoinked from LMS and modified
+# Language conventions (https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2017#language-conventions)
+dotnet_style_qualification_for_field                                     = false:none                           # default
+dotnet_style_qualification_for_property                                  = false:none                           # default
+dotnet_style_qualification_for_method                                    = false:none                           # default
+dotnet_style_qualification_for_event                                     = false:none                           # default
+
+dotnet_style_predefined_type_for_locals_parameters_members               = true:warning
+dotnet_style_predefined_type_for_member_access                           = true:warning
+
+dotnet_style_require_accessibility_modifiers                             = for_non_interface_members:warning
+csharp_preferred_modifier_order                                          = public, private, protected, internal, static, extern, new, virtual, abstract, sealed, override, readonly, unsafe, volatile, async:none # default
+dotnet_style_readonly_field                                              = true:error
+
+dotnet_style_parentheses_in_arithmetic_binary_operators                  = always_for_clarity:error
+dotnet_style_parentheses_in_relational_binary_operators                  = always_for_clarity:error
+dotnet_style_parentheses_in_other_binary_operators                       = always_for_clarity:error
+dotnet_style_parentheses_in_other_operators                              = never_if_unnecessary:none            # default
+
+dotnet_style_object_initializer                                          = true:suggestion                      # default
+dotnet_style_collection_initializer                                      = true:suggestion                      # default
+dotnet_style_explicit_tuple_names                                        = true:error                           # don't use Item1 when there's a meaningful name
+dotnet_style_prefer_inferred_tuple_names                                 = true:suggestion                      # default
+dotnet_style_prefer_inferred_anonymous_type_member_names                 = false:warning                        # inferred name changes can lead to serialization errors and unintended API changes
+dotnet_style_prefer_auto_properties                                      = true:none                            # default
+dotnet_style_prefer_is_null_check_over_reference_equality_method         = true:suggestion                      # default
+dotnet_style_prefer_conditional_expression_over_assignment               = true:none                            # default
+dotnet_style_prefer_conditional_expression_over_return                   = true:none                            # default
+
+dotnet_style_coalesce_expression                                         = true:suggestion                      # default
+dotnet_style_null_propagation                                            = true:suggestion                      # default
+
+csharp_style_var_for_built_in_types                                      = false:suggestion
+csharp_style_var_when_type_is_apparent                                   = false:none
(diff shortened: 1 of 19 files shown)


.editorconfig
```diff
@@ -11,3 +11,93 @@ indent_size = 4
 [{*.json,.*.yml}]
 indent_style = space
 indent_size = 2
+
+[*.cs]
```
boarnoah, line 15: I pulled the .editorconfig from mono-repo to keep things to the general D2L style.

LMK if that any of the settings have some less than ideal behaviors for compatibility's sake that can be ignored in a greenfield project such as this.
(Already found one such rule which disabled IDE formatting at line breaks - removed it since its better to have it do so).

andrewalkema: I'm happy with this editor config, nice to be consistent.

boarnoah:
arch:
-  core (Bmx.Core) - defines process for interacting with 
-  identity providers (ex: Bmx.Idp.Okta - interact via okta for auth, storing sessions, getting potential roles
-  service providers (aws) - get access tokens via STS
-  command line app (bmx.commandline) - interacting with user as a CLI utility, taking input lives here

core runs the core process (of interacting with api) and fires events (which command line listens to) for the app to prompt user to enter in new input). 
Core accesses the providers (service and identity) by interface (eventually this will be done via DI).

Bmx.CommandLine/CommandLine.cs
```diff
@@ -0,0 +1,168 @@
+﻿using System;
```
MattPinelli, line 1: This seems like a really big class - is there an opportunity to split this into smaller classes (only if it makes sense)?

MattPinelli: Oh, I now see that they are all private methods. This is probably fine.

neverendingqs:
You may be interested in something like:

```sh
git checkout --orphan dotnetcore-master
git reset .
# git add any files you want to keep
git clean -xdf  # removes all unstaged files
git push origin dotnetcore-master
```


You can then submit PRs against `dotnetcore-master` instead of `master`, which should reduce the size of your PRs.

`git checkout --orphan` creates a new branch with no relationship with any other branches or commits (aka a "root commit"). You can treat that new branch as a pseudo `master` branch, and then we can make the switch later with `git branch -m` and `git push -f`.

boarnoah:
Wouldn't that result in the loss of git history prior to the write? 
I would prefer to keep old git history around (after this particular PR the rest of the work should be small in scale).

neverendingqs:
Yes and no. The Golang history will live in another branch we can protect, but it will have no relationship with the dotnet branch.

It could also reduce the amount of data sent / received with git push / pull, because future contributors probably don't care about the Golang history.

I am not strongly opposed to the current direction, so feel free to disregard :) .

Bmx.CommandLine/CommandLine.cs
```diff
@@ -0,0 +1,168 @@
...
+		public CommandLine() {
+			// TODO: Fix this :| (ref: Program.cs)
+			_bmx = new BmxCore( new OktaClient( "d2l" ) );
+			_cmdLineParser = BuildCommandLine().UseDefaults().Build();
+
+			_bmx.PromptUserName += GetUser;
+			_bmx.PromptUserPassword += GetPassword;
+			_bmx.PromptMfaType += GetMfaType;
+			_bmx.PromptMfaInput += GetMfaInput;
+			_bmx.PromptRoleSelection += GetRoleType;
```
neverendingqs, line 36: Minor: do we want to go with `m_`, similar to the LMS codebase?

Bmx.CommandLine/CommandLine.cs
```diff
@@ -0,0 +1,168 @@
...
+
+		private string GetUser( string provider ) {
+			Console.Write( $"{provider} Username: " );
+			return Console.ReadLine();
+		}
+
+		private string GetPassword( string provider ) {
+			Console.Write( $"{provider} Password: " );
+
+			string pw = "";
```
neverendingqs, line 92: Minor: StringBuilder maybe?

Bmx.Core/BmxCore.cs
```diff
@@ -0,0 +1,76 @@
...
+
+			if( user == null ) {
+				_user = PromptUserName( _identityProvider.Name );
+			}
+
+			// TODO: Handle case creds wrong / user mia
+			var mfaOptions =
+				await _identityProvider.Authenticate( _user, PromptUserPassword( _identityProvider.Name ) );
+
+			// TODO: Identify if non MFA use is possible with current setup for BMX, if so skip MFA steps
```
neverendingqs, line 51: Apologies in advance if I'm misunderstanding this comment.

If you're on VPN on a D2L computer, MFA is skipped with the Golang version of bmx.

Bmx.CommandLine/CommandLine.cs
boarnoah: AFAIK that naming style is a holdover from older C++ code-bases.

Closest thing to a official .net core [guideline](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) says prefix "_" for non static private fields. It does also say "m_" if the code-base already uses that style instead :)

boarnoah:
Convinced on the orphan branch approach, going to try to get that working. 

In the event it gets tricky I might need to close this and reopen the PR again.

Bmx.Core/BmxCore.cs
boarnoah: Got it, the no mfa use case I was not able to see in use. Will build out the path to skip MFA if its not required.

Bmx.CommandLine/CommandLine.cs
cfbao: FYI, there are discrepancies between D2L standard and C# standards out in the world.
https://docs.dev.d2l/index.php/Coding_Standards_-_CSharp#Classes_and_Interfaces

Bmx.CommandLine/CommandLine.cs
boarnoah: String interpolation is a bit more concise so I'd prefer if we use that. String builder makes sense for larger work for performance, although I think nowadays String interpolation is suited for that as well.

PR #195, opened 2020-06-24
Implement fetching list of AWS accounts from Okta

### LINKED ISSUE #193: Implement listing available AWS accounts from Okta IDP
Can filter User's apps by key "amazon_aws" to get all the AWS accounts the user's have access to via Okta

boarnoah:
closes #193


Bmx.Core/BmxCore.cs
```diff
@@ -70,7 +62,25 @@ public class BmxCore {
 				// TODO: Remove need for input here, Push style flows have no user input on app here
 				// ex: https://developer.okta.com/docs/reference/api/authn/#response-example-waiting
 				PromptMfaInput( selectedMfa.Name );
+				throw new NotImplementedException();
 			}
+
+
+			// TODO: Decouple this more, there is an implicit understanding its an Okta AWS app
```
neverendingqs, line 69: Ah because users can manually add their own login that isn't SAML?

Bmx.Core/BmxCore.cs
```diff
@@ -70,7 +62,25 @@ public class BmxCore {
...
+
+			int selectedAccountIndex = -1;
+
+			if( account != null ) {
+				selectedAccountIndex = Array.IndexOf( accounts.Select( s => s.ToLower() ).ToArray(), account );
+			}
+
+			if( account == null || selectedAccountIndex == -1 ) {
+				selectedAccountIndex = PromptAccountSelection( accounts );
+			}
```
neverendingqs, line 81: Not sure if you were planning to do this in another PR, but thoughts on logging a message if an account isn't found?

Bmx.Core/BmxCore.cs
```diff
@@ -70,7 +62,25 @@ public class BmxCore {
...
+
+			// TODO: Decouple this more, there is an implicit understanding its an Okta AWS app
+			var accounts = await _identityProvider.GetAccounts( "amazon_aws" );
+			account = account?.ToLower();
+
+			int selectedAccountIndex = -1;
+
+			if( account != null ) {
+				selectedAccountIndex = Array.IndexOf( accounts.Select( s => s.ToLower() ).ToArray(), account );
+			}
```
neverendingqs, line 77: I think you can use `FirstOrDefault()` here if you want. That might be more succinct.

```cs
account = accounts
  .Select( s => s.ToLower() )
  .FirstOrDefault( s => s == account );
```

Bmx.Idp.Okta/models/SessionOptions.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Bmx.Idp.Okta.models {
```
neverendingqs, line 1: Double-check: did you intentionally not capitalize `models`?

Bmx.CommandLine/CommandLine.cs
```diff
@@ -18,6 +18,10 @@ class StubIdp : IIdentityProvider {
 		public Task<bool> ChallengeMfa( int selectedMfaIndex, string challengeResponse ) {
 			throw new NotImplementedException();
 		}
+
+		public Task<string[]> GetAccounts( string accountType ) {
+			throw new NotImplementedException();
```
neverendingqs, line 23: Double-check: this will be updated in another PR?

Bmx.Idp.Okta/OktaClient.cs
```diff
@@ -13,7 +15,16 @@ public class OktaClient : IIdentityProvider {
 		private readonly JsonSerializerOptions _serializeOptions =
 			new JsonSerializerOptions {PropertyNamingPolicy = JsonNamingPolicy.CamelCase,};
 
-		private static readonly HttpClient HttpClient = new HttpClient();
+		private static readonly CookieContainer CookieContainer = new CookieContainer();
+
+		private static readonly HttpClient HttpClient = new HttpClient( new HttpClientHandler {
+			CookieContainer = CookieContainer
+		} );
```
neverendingqs, line 22: Curious: why static? It seems like every instance of this class should have their own HttpClient.

Bmx.Idp.Okta/OktaClient.cs
```diff
@@ -46,37 +58,62 @@ public class OktaClient : IIdentityProvider {
...
+				new StringContent( JsonSerializer.Serialize( sessionOptions, _serializeOptions ), Encoding.Default,
+					"application/json" ) );
+			return await JsonSerializer.DeserializeAsync<OktaSession>( await resp.Content.ReadAsStreamAsync(),
+				_serializeOptions );
+		}
+
+		private async Task<OktaApp[]> GetAccountsOkta( string userId ) {
+			var resp = await HttpClient.GetAsync( $"users/{userId}/appLinks" );
+			return await JsonSerializer.DeserializeAsync<OktaApp[]>( await resp.Content.ReadAsStreamAsync(),
+				_serializeOptions );
```
neverendingqs, line 116: Minor: In case you haven't considered this already: it looks like `HttpClientJsonExtensions` contains a `GetFromJson()` you can use. Not sure it supports the same serialization options though.

Bmx.Idp.Okta/OktaClient.cs
```diff
@@ -46,37 +58,62 @@ public class OktaClient : IIdentityProvider {
...
 			var resp = await HttpClient.PostAsync( $"authn/factors/{authOptions.FactorId}/verify",
-				new StringContent( JsonSerializer.Serialize( authOptions, _serializeOptions ) ) );
+				new StringContent( JsonSerializer.Serialize( authOptions, _serializeOptions ), Encoding.Default,
+					"application/json" ) );
 			return await JsonSerializer.DeserializeAsync<AuthenticateResponseSuccess>(
 				await resp.Content.ReadAsStreamAsync(), _serializeOptions );
 		}
+
+		private async Task<OktaSession> CreateSessionOkta( SessionOptions sessionOptions ) {
+			var resp = await HttpClient.PostAsync( "sessions",
```
neverendingqs, line 106: Similar: `HttpClientExtensions` has a `PostAsJsonAsync()` in case you haven't considered it already.

Bmx.Idp.Okta/OktaClient.cs
boarnoah: It does, doesn't seem to be present in .NET Core 3.1, added in 5.0+.

Bmx.Idp.Okta/OktaClient.cs
boarnoah: Good catch, there is no reason for it to be static since if there was somehow multiple instances they should have their own isolated context and cookies.

Bmx.Idp.Okta/OktaClient.cs
neverendingqs: Ah gotcha. Sorry forgot to check version numbers of the docs.

Bmx.Idp.Okta/models/SessionOptions.cs
boarnoah: Nope, just force of habit. Addressed this.

Bmx.Core/BmxCore.cs
boarnoah: Needed the index specifically, since within Idp.Okta the Accounts are in an array with more Okta specific information (links, Ids etc...), whereas BMX Core only needs to care about the names of the accounts (it gets a string array).

Consistent to how MFA Factors was handled.

Bmx.Core/BmxCore.cs
boarnoah: Yep, there is a lot more to be done in terms of handling things that aren't on the happy path:
- Handling API errors from Okta (due to invalid values passed for credentials, mfa tokens etc...) or this
- Handling any other type of error, unable to contact Okta or expired sessions.

Thinking of handling them by having the Identity or Service providers throw exceptions that Bmx Core can notify the app (in this case CommandLineApp to notify the user about before exiting).

.NET Core has good handling for easily swappable logging, so an actual log to file might be a good idea as well. Will create an issue for this.

Bmx.CommandLine/CommandLine.cs
boarnoah: For integration tests will need to write a more fleshed out mock identity provider, Think at present its best if I remove this StubIdp all together.

Bmx.Core/BmxCore.cs
boarnoah: Could you elaborate on this, my understanding is current AWS apps within Okta will all follow the same convention (in terms of API response). Is it possible for users to manually add an AWS account on Okta? 

This todo was more in the context that I need to move the "amazon_aws" string out of Bmx Core into a app settings.

neverendingqs: > This todo was more in the context that I need to move the "amazon_aws" string out of Bmx Core into a app settings.

Ah gotcha

Sometimes, Okta offers to save username / password pairs for you as you sign into stuff (e.g. LMS). I thought you were referring to someone that added an AWS account manually. It's an edge case, and maybe it already gets filtered out by the time the code is here?

boarnoah: Did a few tests, as far as I can tell those do not get saved as "amazon_aws" in Okta. 

However I'm not 100% sure on these results, it does seem like an edge case. My understanding isn't supposed BMX to handle manually added credentials, requires further investigation.

Issue #191, opened 2020-06-22
Investigate using Generic Host Builder for DI and config loading


boarnoah:
.NET Core has generic [Host Builders](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostbuilder?view=dotnet-plat-ext-3.1) which adds support for config loading and DI in the same way as ASP.NET Core (i.e using ConfigureServices).

cfbao:
I don't know how bmx is structured internally, but if it doesn't make sense to use a host, you can still use [Microsoft.Extensions.DependencyInjection](https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection/), [Microsoft.Extensions.Configuration](https://www.nuget.org/packages/Microsoft.Extensions.Configuration/) and similar packages outside of hosts.

boarnoah:
Attempting to use DI I realized I have written myself into a corner with the way I have BmxCore depends on other things (in this case CommandLine app) to have subscribe to events.

Requires refactoring.

boarnoah:
Although System.CommandLine has support for use via Microsoft.Extensions.Hosting as far as I can tell it is in its infancy (many issues on github trackers related to design which makes me think how it works may change radically). 

Therefore decided to stick to only try using the other Extensions like Chenfeng suggested. 

Microsoft.Extensions.DependencyInjection:
Decided to opt for injecting dependencies manually at entrypoint, as far as I can tell this extension doesn't add support for auto injection of dependencies (the auto injection behavior seen within ASP.NET Core seems to be part of that project, and not this extension). 
Therefore it only really acts as a ServiceLocator, which as far as I can tell is considered something of an anti pattern. 

For these reasons it wasn't adding much value so I stuck to doing poormans/pure DI.

Microsoft.Extensions.Configuration:
Planning to make use of this to load the INI files used for config, will be a separate PR (since its a bit more involved with how it reads multiple config files).

cfbao:
> Decided to opt for injecting dependencies manually at entrypoint, as far as I can tell this extension doesn't add support for auto injection of dependencies (the auto injection behavior seen within ASP.NET Core seems to be part of that project, and not this extension).
Therefore it only really acts as a ServiceLocator, which as far as I can tell is considered something of an anti pattern.

Microsoft.Extensions.DependencyInjection gives you constructor injection. It's similar to most other DI frameworks, and not meant to be used as a service locator.

[...]

PR #199, opened 2020-06-25
Exchange AWS SAML from Okta for AWS Access Tokens via STS

### LINKED ISSUE #194: Implement getting SAML for AWS Auth from Okta IDP & role selection
Okta integrates with AWS to provide access to AWS over SAML (https://www.okta.com/integrations/aws-account-federation).

Unfortunately this workflow was not built with programmatic access in mind, so the current (and what previous iterations of BMX) do is some spooky HTML scraping for the SAML data. 

The folks at Okta do not consider this to be a workflow they have in mind either (https://devforum.okta.com/t/authenticating-a-saml-request-in-order-to-download-content-from-an-app/340/5), at some point we need to replace this with something more robust

boarnoah:
closes #194 

Add Cloud Provider (in this case AWS) which exchanges Okta provided SAML for short lived AWS tokens (these are what BMX outputs). 

- Parse out SAML from okta to Identify roles in the account,
- Use AWS SDK for .NET to Assume a role via the SAML

### DIFF
--- a/Bmx.CommandLine/CommandLine.cs
+++ b/Bmx.CommandLine/CommandLine.cs
@@ -14,7 +14,7 @@ namespace Bmx.CommandLine {
 
 		public CommandLine() {
 			// TODO: Fix this :| (ref: Program.cs)
-			_bmx = new BmxCore( new OktaClient( "d2l" ) );
+			_bmx = new BmxCore( new OktaClient(), new AwsClient() );
 			_cmdLineParser = BuildCommandLine().UseDefaults().Build();
 
 			_bmx.PromptUserName += GetUser;
--- a/Bmx.Idp.Okta/OktaClient.cs
+++ b/Bmx.Idp.Okta/OktaClient.cs
@@ -24,16 +24,18 @@ namespace Bmx.Idp.Okta {
 		private OktaMfaFactor[] _oktaMfaFactors;
 		private OktaApp[] _oktaApps;
 
-		public OktaClient( string organization ) {
+		public OktaClient() {
 			_cookieContainer = new CookieContainer();
 			_httpClient = new HttpClient( new HttpClientHandler {CookieContainer = _cookieContainer} );
-
-			_httpClient.BaseAddress = new Uri( $"https://{organization}.okta.com/api/v1/" );
 			_httpClient.Timeout = TimeSpan.FromSeconds( 30 );
 			_httpClient.DefaultRequestHeaders.Accept.Add( new MediaTypeWithQualityHeaderValue( "application/json" ) );
 		}
 
 		public string Name => "Okta";
+
+		public void SetOrganization( string organization ) {
+			_httpClient.BaseAddress = new Uri( $"https://{organization}.okta.com/api/v1/" );
+		}
 
 		public async Task<MfaOption[]> Authenticate( string username, string password ) {
 			var authResp = await AuthenticateOkta( new AuthenticateOptions( username, password ) );
--- a/Bmx.Service.Aws/AwsClient.cs
+++ b/Bmx.Service.Aws/AwsClient.cs
@@ -1,17 +1,27 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Web;
 using System.Xml;
+using Amazon.SecurityToken;
+using Amazon.SecurityToken.Model;
 using Bmx.Core;
 using Bmx.Service.Aws.Models;
 
 namespace Bmx.Service.Aws {
 	public class AwsClient : ICloudProvider {
+		private readonly AmazonSecurityTokenServiceClient _stsClient;
+
 		private string _samlString;
 		private XmlDocument _samlToken;
 		private List<AwsRole> _awsRoles;
 
-		public void SetSamlToken( XmlDocument samlToken ) {
-			_samlToken = samlToken;
+		public AwsClient() {
+			_stsClient = new AmazonSecurityTokenServiceClient();
+		}
+
 		public void SetSamlToken( string encodedSaml ) {
 			_samlString = encodedSaml;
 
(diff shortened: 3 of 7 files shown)


Bmx.CommandLine/CommandLine.cs
```diff
@@ -14,7 +15,7 @@ public class CommandLine {
 
 		public CommandLine() {
 			// TODO: Fix this :| (ref: Program.cs)
-			_bmx = new BmxCore( new OktaClient( "d2l" ) );
+			_bmx = new BmxCore( new OktaClient( "d2l" ), new AwsClient() );
```
andrewalkema, line 18: I assume the `TODO` here is to inject these instead of using `new` and maybe move `"d2l"` to some sort of config?

boarnoah: It is a required parameter of bmx (from 2.0) so from launch args rather than config.

Constructor makes it mandatory to be set, my other alternative is to move setting Organization to some where after the OktaClient is initialized, it seems like a bit of a code smell since there is no way to enforce that organization is set prior to use of this class.

cfbao: > It is a required parameter of bmx (from 2.0) so from launch args rather than config.

Not sure I understand you correctly, but I'm pretty sure in bmx 2+ you can still set org in config.

boarnoah: ATM this rewrite doesn't have read from config yet.
When there isn't a config present its a required parameter with BMX 2.0. 

Regardless of how its set and read in it won't be possible to access organization to pass onto the constructor here so I would need to move the logic to set organization elsewhere.

MattPinelli: Also recall that this is an open source project - there shouldn't be any reference to `d2l` in the code.

cfbao: > Regardless of how its set and read in it won't be possible to access organization to pass onto the constructor here so I would need to move the logic to set organization elsewhere.

What I normally do is to read the config in the constructor. The config can be composed of different sources including config files and command line args.
More here: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1

boarnoah: Would like to handle proper loading of configs with #191 so different PR.

Bmx.Idp.Okta/OktaClient.cs
```diff
@@ -75,10 +79,28 @@ public class OktaClient : IIdentityProvider {
...
+		private string ExtractAwsSaml( string htmlResponse ) {
+			// HTML page is fairly malformed, grab just the <input> with the SAML data for further processing
+			var inputRegexPattern = "<input name=\"SAMLResponse\" type=\"hidden\" value=\".*\"/>";
+			var inputRegex = new Regex( inputRegexPattern, RegexOptions.Compiled );
+
+			// Access the SAML data from the parsed <input>
+			var inputXml = new XmlDocument();
+			inputXml.LoadXml( inputRegex.Match( htmlResponse ).Value );
+
+			return inputXml.SelectSingleNode( "//@value" ).InnerText;
```
neverendingqs, line 101: Thoughts on parsing the SAML response earlier, or adjusting the naming? It's not clear beyond this point that we're passing around SAML response statements until we parse it in `SetSamlToken()`.

Bmx.Service.Aws/AwsClient.cs
```diff
@@ -0,0 +1,76 @@
...
+
+				_awsRoles.Add( new AwsRole {
+					PrincipalArn = nodeContents[0],
+					RoleArn = nodeContents[1],
+					RoleName = nodeContents[1].Split( "/" )[1]
+				} );
+			}
+
+			return _awsRoles.Select( role => role.RoleName ).ToArray();
+		}
```
neverendingqs, line 59: The way `_samlToken` and `_awsRoles` is set suggests that we should use a builder pattern maybe? Something something

```cs
class AwsClient {
  public AwsClient ( AwsRoles[] awsRoles ) {
    _awsRoles = awsRoles
    _stsClient = new AmazonSecurityTokenServiceClient();
  }
  ...
  static AwsClient Create( string encodedSaml ) {
    ...
    var samlToken = LoadXml( ... );
    ...
    var awsRoles = ....
    ...
    return new AwsClient( awsRoles );
  }
}
```
?

This will prevent callers from calling `GetRoles()` or `GetTokens()` before the internals are fully set up.

Bmx.Idp.Okta/OktaClient.cs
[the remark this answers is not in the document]
boarnoah: Yes, its not ideal but as far as I can tell they decided to do so as there was no viable alternatives. 

Mentioned in the ticket #194 there might be a different way we could pursue at a later point (unsure as to the scope of work to do so).

Bmx.Service.Aws/AwsClient.cs
boarnoah: Thanks! I really didn't like how I had it setup, this pattern solves that problem very well.

Bmx.Idp.Okta/OktaClient.cs
boarnoah: I should clarify the variable names here, Identity provider at this point is dealing with HTML (inputXML) is an HTML<input> node. 

I used XPath parsing  at this point since it was a straightforward way to get the contents of the <input> node's Value attribute.

As to why Identity Provider is passing the token out as a encoded string, initially I had it setup where it did pass parsed SAML as an XML doc, but it requires writing a lot of additional things to re-encode and export the XMLDocument back to a string that the AWS API accepts.

Bmx.Service.Aws/AwsClient.cs
boarnoah: While trying to implement DI and Unit Tests I've realized this issue extends deeper than just this setting. 

Services for Okta and AWS both are stateful, and there is a very strong temporal coupling (methods needing to be called in a very specific order for the app to work without runtime errors). 

As far as I can see there is no way to remove this coupling without making the services stateless (and by leaking internal implementation details from those classes into BmxCore).

tl;dr its an extensive issue that goes deeper than just having the presence encodedSaml guaranteed via builder pattern, requires investigation on its own IMO.

neverendingqs: Oof gotcha. Makes sense to me.

PR #202, opened 2020-07-15
Implement loading args from config

### LINKED ISSUE #201: Implement config loading
Extensions.Configuration.Ini should work.

Config as defined in bmx v2.0:

```
type UserConfig struct {
	AllowProjectConfigs bool
	Org                 string
	User                string
	Account             string
	Role                string
	Profile             string
}
```

Read from ~/.bmx/config file and optionally .bmx files inside CWD (read recursively, could be in sub-folders).

boarnoah:
Closes #201 

Did some odd stuff for merging values for runtime args to values found in configs, System.CommandLine doesn't have a good solution for this [at the moment](https://github.com/dotnet/command-line-api/issues/857).

Aside:
That lib has a lot of in-draft/wip features so I think it makes sense to do some things manually for now.

### DIFF
--- a/Bmx.CommandLine/IniConfiguration.cs
+++ b/Bmx.CommandLine/IniConfiguration.cs
@@ -3,6 +3,8 @@ using System;
 using System.IO;
 using System.Security;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.FileProviders.Physical;
 
 namespace Bmx.CommandLine {
 	public class IniConfiguration : IBmxConfig {
@@ -24,31 +26,28 @@ namespace Bmx.CommandLine {
 
 		private static IConfiguration GetConfiguration() {
 			// Main config is at ~/.bmx/config
-			var configLocation =
-				Path.Join( Environment.GetFolderPath( Environment.SpecialFolder.UserProfile ), ".bmx", "config" );
+			var configLocation = Path.Join( Environment.GetFolderPath( Environment.SpecialFolder.UserProfile ), ".bmx",
+				"config" );
 
 			// If set, we recursively look up from CWD (all the way to root) for additional bmx config files (labelled as .bmx)
-			var shouldReadProjectConfig = ShouldReadProjectConfigs( configLocation );
+			var config = new ConfigurationBuilder().AddIniFile( configLocation, optional: true );
+
+			bool.TryParse( config.Build()["allow_project_configs"], out bool shouldReadProjectConfig );
+
 			string? projectConfigPath = null;
 
 			if( shouldReadProjectConfig ) {
 				projectConfigPath = GetProjectConfigPath();
 			}
 
-			var config = new ConfigurationBuilder().AddIniFile( configLocation );
-
 			if( shouldReadProjectConfig && projectConfigPath != null ) {
-				config.AddIniFile( projectConfigPath );
+				// Default file provider ignores files prefixed with ".", we need to provide our own as a result
+				var fileProvider =
+					new PhysicalFileProvider( Path.GetDirectoryName( projectConfigPath ), ExclusionFilters.None );
+				config.AddIniFile( fileProvider, Path.GetFileName( projectConfigPath ), optional: false, reloadOnChange: false );
 			}
 
 			return config.Build();
-		}
-
-		private static bool ShouldReadProjectConfigs( string primaryConfigLocation ) {
-			var config = new ConfigurationBuilder()
-				.AddIniFile( primaryConfigLocation ).Build();
-			bool.TryParse( config["allow_project_configs"], out bool allowProjectConfigs );
-			return allowProjectConfigs;
 		}
 
 		// Look from cwd up the directory chain all the way to root for a .bmx file
--- a/Bmx.CommandLine/IniConfiguration.cs
+++ b/Bmx.CommandLine/IniConfiguration.cs
@@ -1,0 +1,81 @@
+﻿#nullable enable
+using System;
+using System.IO;
+using System.Security;
+using Microsoft.Extensions.Configuration;
+
+namespace Bmx.CommandLine {
+	public class IniConfiguration : IBmxConfig {
+		public string? Org { get; }
+		public string? User { get; }
+		public string? Account { get; }
+		public string? Role { get; }
+		public string? Profile { get; }
+
+		public IniConfiguration() {
+			var config = GetConfiguration();
+
+			Org = config["org"];
+			User = config["user"];
+			Account = config["account"];
+			Role = config["role"];
+			Profile = config["profile"];
+		}
+
+		private static IConfiguration GetConfiguration() {
+			// Main config is at ~/.bmx/config
+			var configLocation =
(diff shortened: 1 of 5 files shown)


Bmx.CommandLine/IniConfiguration.cs
```diff
@@ -0,0 +1,81 @@
...
+			string? projectConfigPath = null;
+
+			if( shouldReadProjectConfig ) {
+				projectConfigPath = GetProjectConfigPath();
+			}
+
+			var config = new ConfigurationBuilder().AddIniFile( configLocation );
+
+			if( shouldReadProjectConfig && projectConfigPath != null ) {
+				config.AddIniFile( projectConfigPath );
```
neverendingqs, line 41: Double-check: order of config precedence is in reverse of the files added?

Bmx.CommandLine/IniConfiguration.cs
```diff
@@ -0,0 +1,81 @@
...
+
+		// Look from cwd up the directory chain all the way to root for a .bmx file
+		private static string? GetProjectConfigPath() {
+			DirectoryInfo? currentDirectory = new DirectoryInfo( Directory.GetCurrentDirectory() );
+
+			while( currentDirectory != null ) {
+				try {
+					var bmxConfigFiles = currentDirectory.GetFiles( ".bmx" );
+
+					if( bmxConfigFiles.Length > 1 ) {
```
neverendingqs, line 62: Is this only possible if there is a folder and file named `.bmx`?

Bmx.CommandLine/IniConfiguration.cs
boarnoah: [Yes](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?tabs=basicconfiguration&view=aspnetcore-3.1#default-configuration). 

Files added later (in this case project config) override config values from previous file (bmx overall config). 
And runtime args override both files.

Matches BMX 2.0's order.

Bmx.CommandLine/IniConfiguration.cs
boarnoah: It should work whenever there is a file labelled .bmx anywhere in the tree starting with cwd. So:

Assuming cwd is /home/foo/project/:

/home/foo/project/.bmx
/home/foo/.bmx
/home/.bmx
/.bmx 

are all valid destinations for the .bmx file for project configuration purposes. Bit weird but Implementing this behavior since its preexisting.

boarnoah:
I created a bug ( ~/.bmx/config not being optional), addressing now.

Bmx.CommandLine/IniConfiguration.cs
```diff
@@ -0,0 +1,81 @@
...
+
+			// If set, we recursively look up from CWD (all the way to root) for additional bmx config files (labelled as .bmx)
+			var shouldReadProjectConfig = ShouldReadProjectConfigs( configLocation );
+			string? projectConfigPath = null;
+
+			if( shouldReadProjectConfig ) {
+				projectConfigPath = GetProjectConfigPath();
+			}
+
+			var config = new ConfigurationBuilder().AddIniFile( configLocation );
```
cfbao, line 38: `new ConfigurationBuilder().AddIniFile` is called twice in this block (once inside `ShouldReadProjectConfigs`). I feel it can be avoided with better clarity.
Also, if you use `GetValue<bool>`, I don't find `ShouldReadProjectConfigs` adds much value as a separate method.

```suggestion
			var config = new ConfigurationBuilder().AddIniFile( configLocation, optional: true );
			bool shouldReadProjectConfig = config.Build().GetValue<bool>( "allow_project_configs" );
			string? projectConfigPath = null;

			if( shouldReadProjectConfig ) {
				projectConfigPath = GetProjectConfigPath();
			}
```

Bmx.CommandLine/IniConfiguration.cs
cfbao: @neverendingqs `GetFiles` should only find files not folders. I also don't know of a file system where a file and a folder can have the same path.

cfbao:
> Did some odd stuff for merging values for runtime args to values found in configs, System.CommandLine doesn't have a good solution for this [at the moment](https://github.com/dotnet/command-line-api/issues/857).
> 
> Aside:
> That lib has a lot of in-draft/wip features so I think it makes sense to do some things manually for now.

unfortunate 😞

Bmx.CommandLine/IniConfiguration.cs
boarnoah: Fair enough, removed it. However I still need to build the configuration twice. First to read it in order to see if project configs should be included, and then again including the project config.

Bmx.CommandLine/IniConfiguration.cs
```diff
@@ -0,0 +1,81 @@
...
+				config.AddIniFile( projectConfigPath );
+			}
+
+			return config.Build();
+		}
+
+		private static bool ShouldReadProjectConfigs( string primaryConfigLocation ) {
+			var config = new ConfigurationBuilder()
+				.AddIniFile( primaryConfigLocation ).Build();
+			bool.TryParse( config["allow_project_configs"], out bool allowProjectConfigs );
```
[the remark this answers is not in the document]
boarnoah: Decided to not use that since aside from this one boolean, the rest of the config options are consumed as simple strings.

Did strictly type the config a bit by having IniConfiguration implement IBmxConfig which lists the types it expects config to have.

Bmx.CommandLine/IniConfiguration.cs
```diff
@@ -24,33 +26,30 @@ public class IniConfiguration : IBmxConfig {
...
 
 			if( shouldReadProjectConfig ) {
 				projectConfigPath = GetProjectConfigPath();
 			}
 
-			var config = new ConfigurationBuilder().AddIniFile( configLocation );
-
 			if( shouldReadProjectConfig && projectConfigPath != null ) {
-				config.AddIniFile( projectConfigPath );
+				// Default file provider ignores files prefixed with ".", we need to provide our own as a result
```
cfbao, line 44: 🧐

Bmx.CommandLine/IniConfiguration.cs
```diff
@@ -0,0 +1,80 @@
...
+			Profile = config["profile"];
+		}
+
+		private static IConfiguration GetConfiguration() {
+			// Main config is at ~/.bmx/config
+			var configLocation = Path.Join( Environment.GetFolderPath( Environment.SpecialFolder.UserProfile ), ".bmx",
+				"config" );
+
+			// If set, we recursively look up from CWD (all the way to root) for additional bmx config files (labelled as .bmx)
+			var config = new ConfigurationBuilder().AddIniFile( configLocation, true );
```
cfbao, line 33: nit: name the parameter when passing in a literal for better readability. At least that's what's done in the LMS.
```suggestion
			var config = new ConfigurationBuilder().AddIniFile( configLocation, optional: true );
```

cfbao: similarly on L47.

boarnoah: 👍 that's a pretty cool feature.

PR #205, opened 2020-07-20
Refactor Identity and Cloud provider services to be stateless.

boarnoah:
More of an experimental thing, if it looks too silly / convoluted please feel free to say so, its not essential that it's merged.

BMX is primarily about talking to Okta (beyond a generic interface of course). The Okta Authentication API is stateful with for example a state token during authentication which is then exchanged for a session token. 

Its easy enough to store this state internally to OktaIdentityProvider without leaking implementation data to BMXCore (the project that orchestrates the various calls to Identity and Cloud Providers. [Workflow diagram](https://github.com/Brightspace/bmx/wiki/Architecture).

What this means though is that for example Identity Service has very strong sequential coupling, calls cannot be called out of order or skipped. Which is a code smell + it makes unit testing painful: 
ex:

`
_identityProvider.Authenticate(...);
// This call can't happen till .Authenticate has been called etc..., the state is stored internal to _identityProvider
_identityProvider.GetAccounts(...)
`
Couple of solutions:
Maintain state within BmxCore, passed to service calls as arguments (i.e GetAccounts requires a state object that can only be received from Authenticate call) :
1. Leak implementation details (of Okta state) directly to BmxCore
2. Have BmxCore interact with state as generics (implementations such as OktaIdentityProvider will use more specific types) <- What I've implemented in this PR

An alternate suggestion on SE StackExchange (where I asked this)  was to move the state to a wrapper class maintaining state that can be disposed afterwards. Don't see how that solves the problem of sequential coupling IMO. 

PS:  SE Stackexchange thread for a more general [answer](https://softwareengineering.stackexchange.com/questions/412544/avoiding-sequential-coupling-while-maintaining-separation-of-concerns)

Long PR with a lot of diff >.<

### DIFF
--- a/Bmx.CommandLine/CommandLine.cs
+++ b/Bmx.CommandLine/CommandLine.cs
@@ -5,14 +5,20 @@ using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
 using System.Threading.Tasks;
 using Bmx.Core;
+using Bmx.Core.State;
 
 namespace Bmx.CommandLine {
-	public class CommandLine {
-		private readonly IBmxCore _bmx;
+	public class CommandLine<TAuthenticateState, TAuthenticatedState, TAccountState, TRoleState>
+		where TAuthenticateState : IAuthenticateState
+		where TAuthenticatedState : IAuthenticatedState
+		where TAccountState : IAccountState
+		where TRoleState : IRoleState {
+		private readonly IBmxCore<TAuthenticateState, TAuthenticatedState, TAccountState, TRoleState> _bmx;
 		private readonly IBmxConfig _bmxConfig;
 		private readonly Parser _cmdLineParser;
 
-		public CommandLine( IBmxCore bmx, IBmxConfig bmxConfig ) {
+		public CommandLine( IBmxCore<TAuthenticateState, TAuthenticatedState, TAccountState, TRoleState> bmx,
+			IBmxConfig bmxConfig ) {
 			_bmx = bmx;
 			_bmxConfig = bmxConfig;
 			_cmdLineParser = BuildCommandLine().UseDefaults().Build();
--- a/Bmx.Idp.Okta/State/OktaAccountState.cs
+++ b/Bmx.Idp.Okta/State/OktaAccountState.cs
@@ -7,12 +7,10 @@ namespace Bmx.Idp.Okta.State {
 		public OktaAccountState( OktaApp[] oktaApps, string accountType ) {
 			OktaApps = oktaApps;
 			AccountType = accountType;
+			Accounts = OktaApps.Where( app => app.AppName == AccountType ).Select( app => app.Label ).ToArray();
 		}
 
-		public string[] Accounts {
-			get => OktaApps.Where( app => app.AppName == AccountType ).Select( app => app.Label ).ToArray();
-		}
-
+		public string[] Accounts { get; }
 		internal string AccountType { get; }
 		internal OktaApp[] OktaApps { get; }
 	}
--- a/Bmx.Idp.Okta/State/OktaAccountState.cs
+++ b/Bmx.Idp.Okta/State/OktaAccountState.cs
@@ -1,0 +1,19 @@
+﻿using System.Linq;
+using Bmx.Core.State;
+using Bmx.Idp.Okta.Models;
+
+namespace Bmx.Idp.Okta.State {
+	public class OktaAccountState : IAccountState {
+		public OktaAccountState( OktaApp[] oktaApps, string accountType ) {
+			OktaApps = oktaApps;
+			AccountType = accountType;
+		}
+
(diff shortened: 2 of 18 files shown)


Bmx.CommandLine/CommandLine.cs
```diff
@@ -5,14 +5,20 @@
 using System.CommandLine.Parsing;
 using System.Threading.Tasks;
 using Bmx.Core;
+using Bmx.Core.State;
 
 namespace Bmx.CommandLine {
-	public class CommandLine {
-		private readonly IBmxCore _bmx;
+	public class CommandLine<TAuthenticateState, TAuthenticatedState, TAccountState, TRoleState>
+		where TAuthenticateState : IAuthenticateState
```
MattPinelli, line 12: Is this syntax related to .NET Core DI? Never seen it before.

andrewalkema: This is a generic type constraint, I think it's also in .NET Framework.  Usually these `T` style arguments can be any time, Adipa is saying `TAuthenticateState` has to be an implementation of `IAuthenticateState`.

MattPinelli: Ah cool, thanks.

boarnoah:
Appreciate if could I get some thoughts on this, @andrewalkema (you already did in slack :P ), @neverendingqs @cfbao

In short, reduces sequential coupling of services by requiring state to be passed into method calls as typed objects. That way its more explicit what the order of calls are (call 2 requires an object that call 1 generates) and makes testing simpler.

Bmx.Idp.Okta/State/OktaAccountState.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Bmx.Idp.Okta.State {
+	public class OktaAccountState : IAccountState {
+		public OktaAccountState( OktaApp[] oktaApps, string accountType ) {
+			OktaApps = oktaApps;
+			AccountType = accountType;
+		}
+
+		public string[] Accounts {
+			get => OktaApps.Where( app => app.AppName == AccountType ).Select( app => app.Label ).ToArray();
```
neverendingqs, line 13: Minor: we could do this once in the constructor instead of every time we call the getter.

Bmx.Idp.Okta/State/OktaAuthenticateState.cs
```diff
@@ -0,0 +1,31 @@
...
+				if( factor.FactorType.Contains( "token" ) || factor.FactorType.Contains( "sms" ) ) {
+					return new MfaOption( factor.FactorType, MfaType.Challenge );
+				}
+
+				if( factor.FactorType.Contains( "push" ) ) {
+					return new MfaOption( factor.FactorType, MfaType.Verify );
+				}
+
+				return new MfaOption( factor.FactorType, MfaType.Unknown );
+			} ).ToArray();
```
neverendingqs, line 24: Similar: could we simply do this once in the constructor?

boarnoah:
Agree its not super clear at glance value, however it is being checked at compile time as opposed to runtime down casting (because of how I used generics).

For example it is not possible to instantiate an object of different implementation of IAuthenticateState and pass it in, OktaClient expects it to be of type OktaAuthenticateState (This happens on compile / IDE warning).

The option of using callback events for these other services is intriguing, I'm worried it'll make it harder to reason with BmxCore however.

neverendingqs:
> For example it is not possible to instantiate an object of different implementation of IAuthenticateState and pass it in, OktaClient expects it to be of type OktaAuthenticateState (This happens on compile / IDE warning).

That largely resolves my concern then. Thanks!

> The option of using callback events for these other services is intriguing, I'm worried it'll make it harder to reason with BmxCore however.

Double-check: are we both thinking something like this?

```cs
string accountCredentials = await _identityProvider.GetServiceProviderSaml(
  org: org,
  getUser: (idpName) => user ?? PromptUserName( idpName ),
  getPassword: (idpName) => PromptUserPassword( idpName ),
  getMfaType: (mfaOptionNames) => PromptMfaType( mfaOptionNames );
  getMfa: (selectedMfa) => PromptMfaInput( selectedMfa )
);
```

boarnoah:
Sorry, missed the notification.

I guess doing it that way via callback eliminates need for BmxCore to be unaware of state altogether I guess (whether that was implicitly in sequential coupling or explicit like this PR) - state would instead be isolated to just the Okta project. 

The question is would it be more clearer up front with callbacks or explicit state pass through to methods. 
Callbacks certainly make the code a bit more concise (BmxCore is relegated to a much smaller / vestigial role this way).

boarnoah:
Had some further thought on this, think I'll merge it in later today.

I wrote tests using this approach and it was fairly straightforward to pass in a specific state to a Service method and test it it in isolation this way.

PR #206, opened 2020-08-14
Add Unit tests for BmxCore

boarnoah:
Used NUnit3 + Moq, looked around in the Brightspace org, they seem to be the preferred tools.

### DIFF
--- a/Bmx.Core.Test/EventTests.cs
+++ b/Bmx.Core.Test/EventTests.cs
@@ -1,0 +1,218 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Bmx.Core.State;
+using Bmx.Core.Test.State;
+using Moq;
+using NUnit.Framework;
+
+namespace Bmx.Core.Test {
+	[TestFixture]
+	[Category( "Events" )]
+	public class EventTests {
+		private Mock<IIdentityProvider<IAuthenticateState, IAuthenticatedState, IAccountState>>
+			_mockIdentityProvider;
+
+		private Mock<ICloudProvider<IRoleState>> _mockCloudProvider;
+
+		private IBmxCore<IAuthenticateState, IAuthenticatedState, IAccountState, IRoleState> _bmxCore;
+
+		private const string MockIdentityProviderName = "mockProvider";
+
+		[SetUp]
+		public void Setup() {
+			_mockIdentityProvider =
+				new Mock<IIdentityProvider<IAuthenticateState, IAuthenticatedState, IAccountState>>();
+			_mockCloudProvider = new Mock<ICloudProvider<IRoleState>>();
+
+			_bmxCore = new BmxCore<IAuthenticateState, IAuthenticatedState, IAccountState, IRoleState>(
+				_mockIdentityProvider.Object, _mockCloudProvider.Object );
+
+			_bmxCore.PromptUserName += provider => "";
+			_bmxCore.PromptUserPassword += provider => "";
+			_bmxCore.PromptMfaType += options => 0;
+			_bmxCore.PromptMfaInput += prompt => "";
+			_bmxCore.PromptAccountSelection += accounts => 0;
+			_bmxCore.PromptRoleSelection += roles => 0;
+
+			_mockIdentityProvider.Setup( m => m.Name ).Returns( MockIdentityProviderName );
+			_mockIdentityProvider.Setup( provider => provider.Authenticate( It.IsAny<string>(), It.IsAny<string>() ) )
+				.ReturnsAsync( new MockAuthenticateState() );
+			_mockIdentityProvider
+				.Setup( provider =>
+					provider.ChallengeMfa( It.IsAny<MockAuthenticateState>(), It.IsAny<int>(), It.IsAny<string>() ) )
+				.ReturnsAsync( new MockAuthenticatedState() );
+			_mockIdentityProvider
+				.Setup( provider => provider.GetAccounts( It.IsAny<MockAuthenticatedState>(), It.IsAny<string>() ) )
+				.ReturnsAsync( new MockAccountState() );
+			_mockIdentityProvider
+				.Setup( provider => provider.GetServiceProviderSaml( It.IsAny<MockAccountState>(), It.IsAny<int>() ) )
+				.ReturnsAsync( "" );
+
+			_mockCloudProvider.Setup( provider => provider.GetRoles( It.IsAny<string>() ) )
+				.Returns( new MockRoleState() );
+			_mockCloudProvider.Setup( provider => provider.GetTokens( It.IsAny<MockRoleState>(), It.IsAny<int>() ) )
+				.ReturnsAsync( new Dictionary<string, string>() );
+		}
+
+		/*
+		 * Tests if BMXCore fires input prompt events if it didn't receive values for required input at the start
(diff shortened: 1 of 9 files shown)


boarnoah:
Looking to finish what was left with BMX since last time I looked at it during Inspiration.

This + a branch that adds unit tests for Okta are in flight, would appreciate if I could get some eyes on it.

MattPinelli (approved):
LGTM

Bmx.Core.Test/EventTests.cs
```diff
@@ -0,0 +1,218 @@
...
+		[Test]
+		[Category( "PromptPassword" )]
+		public void PromptUserPasswordCalled() {
+			bool eventRaised = false;
+			_bmxCore.PromptUserPassword += providerName => {
+				eventRaised = true;
+				return "";
+			};
+			_bmxCore.Print( "org" );
+			Assert.IsTrue( eventRaised );
```
neverendingqs, line 95: Is it not possible to assert what `providerName` should be? Or does it not make sense to test that?

Bmx.Core.Test/EventTests.cs
```diff
@@ -0,0 +1,218 @@
...
+		[Test]
+		[Category( "InformUnknownMfaTypes" )]
+		public void InformUnknownMfaTypesCalled() {
+			bool eventRaised = false;
+			_bmxCore.InformUnknownMfaTypes += mfaOptions => {
+				eventRaised = true;
+			};
+			// Default mock state includes a MfaType.Unknown MfaOption
+			_bmxCore.Print( "org" );
+			Assert.IsTrue( eventRaised );
```
neverendingqs, line 107: Similar: Is it not possible to assert what `mfaOptions` should be? Or does it not make sense to test that?

Bmx.Core.Test/State/MockAuthenticateState.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Bmx.Core.State;
+
+namespace Bmx.Core.Test.State {
+	public class MockAuthenticateState : IAuthenticateState {
```
neverendingqs, line 4: Curious: why did you choose a custom implementation for testing instead of mocking via `moq`?

Bmx.Core.Test/EventTests.cs
```diff
@@ -0,0 +1,218 @@
+using System;
```
neverendingqs, line 1: Double-check: are more tests inbound? I noticed we don't verify calls to `_identityProvider`, and it doesn't seem like the conditionals in https://github.com/Brightspace/bmx/blob/46f5346c6b1be35e158d8366bedaf96b43980c07/Bmx.Core/BmxCore.cs#L71-L118 are being covered.

neverendingqs (commented):
Meta: thoughts on setting up CI to run unit tests? Would also be neat to be able to see code coverage (I know we use coveralls for some of our Node stuff).

Bmx.Core.Test/EventTests.cs
```diff
@@ -0,0 +1,218 @@
...
+
+			_bmxCore = new BmxCore<IAuthenticateState, IAuthenticatedState, IAccountState, IRoleState>(
+				_mockIdentityProvider.Object, _mockCloudProvider.Object );
+
+			_bmxCore.PromptUserName += provider => "";
+			_bmxCore.PromptUserPassword += provider => "";
+			_bmxCore.PromptMfaType += options => 0;
+			_bmxCore.PromptMfaInput += prompt => "";
+			_bmxCore.PromptAccountSelection += accounts => 0;
+			_bmxCore.PromptRoleSelection += roles => 0;
```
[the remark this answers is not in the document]
boarnoah: By default if nothing is subscribed to an event C# will not complain.

I found this problematic WRT Bmx.Core, since any clients (such as Bmx.CommandLine, needs to subscribe for Bmx.Core to work as expected).

So I enforced that events are subscribed to using assertions (https://github.com/Brightspace/bmx/blob/dotnet/Bmx.Core/BmxCore.cs#L39). 

What this means however is the unit tests need to subscribe at least once before invoking Bmx.Core at all, else the assertions will fail.

Bmx.Core/BmxCore.cs
```diff
@@ -21,7 +21,7 @@ public class
 		public event PromptMfaInputHandler PromptMfaInput;
 		public event PromptAccountSelectHandler PromptAccountSelection;
 		public event PromptRoleSelectionHandler PromptRoleSelection;
-		public event PromptRoleSelectionHandler InformUnknownMfaTypesHandler;
+		public event InformUnknownMfaTypesHandler InformUnknownMfaTypes;
```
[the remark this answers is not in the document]
boarnoah: Yep

Bmx.Core.Test/EventTests.cs
boarnoah: Doesn't make sense, as org (ex: "d2l") is a user option.

Bmx.Core.Test/EventTests.cs
boarnoah: It does but as part of Bmx.Idp.Okta tests, since we want to ensure that that particular project is parsing a given API response to the appropriate mfaOptions.

Bmx.Core on the other hand doesn't need to be aware of anything past whether or not the event was triggered as part of it's workflow.

Bmx.Core.Test/EventTests.cs
boarnoah: >  I noticed we don't verify calls to _identityProvider
I have tests written for testing the methods inside Bmx.Idp.Okta (separate test project). https://github.com/Brightspace/bmx/pull/208 (its WIP). Is 

that what you meant? Here Bmx.Core is tested in isolation by using a mock Identity Provider.

> it doesn't seem like the conditionals in [Line 71 - 118] are being covered.
Yeah :( tests for these are bit more complicated so I was planning to leave them out till later.

As an aside, it seems like a bit of a warning the code is doing a lot more than it should in one go when testing it is hard.

Bmx.Core.Test/State/MockAuthenticateState.cs
boarnoah: Good point, TBH I haven't really used Moq or NUnit much at all prior to this, investigating this.

Bmx.Core.Test/EventTests.cs
neverendingqs: > > I noticed we don't verify calls to `_identityProvider`
> 
> I have tests written for testing the methods inside Bmx.Idp.Okta (separate test project). #208 (its WIP).
> 
> Is that what you meant? Here Bmx.Core is tested in isolation by using a mock Identity Provider.

Sort of. `_identityProvider` is an implementation of `IIdentityProvider<TAuthenticateState, TAuthenticatedState, TAccountState>`. We can't test the implementation here, but we can test that we are calling the interface methods properly.

For example, we can confirm that `_identityProvider.Authenticate()` gets called with the proper arguments (from https://github.com/Brightspace/bmx/blob/46f5346c6b1be35e158d8366bedaf96b43980c07/Bmx.Core/BmxCore.cs#L54-L55).

If you're convinced to switch to strict mocks and call `_mockIdentityProvider.VerifyAll()` at the end of every test (maybe on `[Teardown]`), that should help verify most of the calls to `_identityProvider` if you only set up for arguments you expect.

PR #208, opened 2021-01-25
Add unit tests for Bmx.Idp.Okta

boarnoah:
Theses tests mostly check if the Idp.Okta methods are able to parse out expected results given mock API response data.


neverendingqs (commented):
Double-check: more tests are en route? I noticed not all methods were covered at this time.

I have a similar comment wrt. strict mocks, but otherwise no concerns.

PR #210, opened 2021-10-07
Add macOS build

apalaniuk:


apalaniuk:
@boarnoah A locally-compiled build was tested, and the release build will be used right after merge to validate it.

PR #211, opened 2022-02-07
default the profile to default

josephsaffron:


AlexBedley (approved):

neverendingqs:
Added @andrewalkema as a reviewer, because it looks like you're the only one on @Brightspace/bmx-maintainers?

neverendingqs:
Oh wait @boarnoah! Should you be on @Brightspace/bmx-maintainers as well?

boarnoah (approved):
We should really update that..
This looks fine to me

cmd/bmx/write.go
```diff
@@ -42,7 +42,7 @@ func init() {
 	}
 
 	if userConfig.Profile == "" {
-		writeCmd.MarkFlagRequired("profile")
+		userConfig.Profile = "default"
 	}
```
NolanKingdon, line 46: The `.Flags().StringVar(...)` call on line 33 gives the option for default values. You could do something like:
```go
writeCmd.Flags().StringVar(&writeOptions.Profile, "profile", "default", "aws profile name")
```

and delete this check altogether I believe. I don't have go on my dev machine, so I can't test.

andrewalkema:
I know Vulcan owns this now, but I really think `default` profile should be avoided whenever possible (I'd suggest never using it).  If `default` profile contains valid credentials, anything running on your machine could inadvertently run against whatever those creds let you.  If you're using `write`, you should put it into a specific named profile for the task you're doing, and configure that profile in vscode or wherever you need it.  I'm curious what the need for `default` is?

cfbao:
We've discussed this change in our team, and decided to close this PR for now.
The security risk introduced with `default` as the default profile is not warranted for the convenience in our opinion.

We'd be happy to reconsider if there's a compelling use case that needs this default behaviour.
Feel free to continue discussion here.

PR #212, opened 2022-03-07
Update BMX readme with more explicit setup steps

adkulas:
Updating the docs to configure and use bmx with aws

This was born out of notes I made during my own struggle to configure and use bmx.
Addition includes:
- pre-requisites to use bmx/aws-cli
- steps to setup bash alias to call bmx with required flags
- details on changing region while using bmx
- Verification of successful aws configuration
- sanity checks to confirm bmx is working correctly

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,3 +93,111 @@ allow_project_configs=true
 org=my_okta_org
 user=my_user
 ```
+
+## How to use bmx and the AWS cli 🚲 - Opinionated Workflow
+
+### Pre-requisites
+- [Install the aws cli](https://aws.amazon.com/cli/)
+- Install and configure the d2l bmx tool
+  - [Guide on how to install bmx](https://github.com/Brightspace/bmx/wiki/Quick-Start)
+
+
+### Step 1 - Add this alias to your bashrc
+
+```bash
+# bashrc
+export OKTA_USER={YOUR_OKTA-USER_HERE}
+alias bmxAwsAdmin='
+    export AWS_PROFILE=dev-profile
+    bmx write --org my_okta_org --user $OKTA_USER --account YOUR-ACCOUNT-HERE --role YOUR-ROLE-HERE --profile dev-profile'
+alias bmxAwsReadOnly='
+    export AWS_PROFILE=dev-profile
+    bmx write --org my_okta_org --user $OKTA_USER --account YOUR-ACCOUNT-HERE --role YOUR-ROLE-HERE-ReadOnly --profile dev-profile'
+```
+
+Note: Be sure to restart your shell or run `source ~/.bashrc` to make the alias available in the shell
+Note: Be sure to replace `--account` and `--role` flags with the account and role you want to have credentials for
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,3 +93,109 @@ allow_project_configs=true
...
+- Install and configure the d2l bmx tool
+  - [bmx repo](https://github.com/Brightspace/bmx)
+  - [Confluence guide on how to install bmx](https://desire2learn.atlassian.net/wiki/spaces/TH/pages/1011745036/Installing+and+using+BMX)
+
+
+### Step 1 - Add this alias to your bashrc
+
+```bash
+# bashrc
+export OKTA_USER={YOUR_OKTA-USER_HERE}
```
boarnoah, line 110: Not sure if this is the norm, how I have it setup is a file in ~/.bmx/config with:
```
org = d2l
user = awijayathilaka
allow_project_configs = true
```
to avoid having to specify org and user everywhere

README.md
```diff
@@ -93,3 +93,109 @@ allow_project_configs=true
...
 user=my_user
 ```
+
+## How to use bmx and the AWS cli 🚲 - Opinionated Workflow
+
+### Pre-requisites
+- [Install the aws cli](https://aws.amazon.com/cli/)
+- Install and configure the d2l bmx tool
+  - [bmx repo](https://github.com/Brightspace/bmx)
+  - [Confluence guide on how to install bmx](https://desire2learn.atlassian.net/wiki/spaces/TH/pages/1011745036/Installing+and+using+BMX)
```
boarnoah, line 103: Consider linking to https://github.com/Brightspace/bmx/wiki/Quick-Start , its more recently updated. Wasn't really aware of that confluence page at all TBH

README.md
adkulas: Yeah I understand that the config file is meant to store the flag values, but I remember when going through the setup, it made it more difficult to understand what it was actually doing and how that would interact with the aws-cli and env variables. Perhaps we would need to just add more detail on the config file works? For example, if I want two different accounts setup, do I do that in the one config file?

adkulas: Ah, I guess this would just be for the flags that would be common to all accounts, i.e. org and user

boarnoah: Yeah, it reads both the file at ~/.bmx/config, and any .bmx (composing them together, with local file taking precedence) files in working directory (or up the chain from there) not the most intuitive behaviour.

adkulas: Updated the setup to suggest using the profile for the two flags
https://github.com/Brightspace/bmx/pull/212/commits/aa54f4207b26faabcdb33d4d62d7678b0a9e3141

README.md
adkulas: Linked to the wiki instead 👍 
https://github.com/Brightspace/bmx/pull/212/commits/aa54f4207b26faabcdb33d4d62d7678b0a9e3141

boarnoah (approved):
Nice update, I especially like the little guide re: AWS CLI, it was definitely one of the more unsaid things re: using AWS that proves to be a stumbling block.

PR #214, opened 2022-05-17
Adding flags to request different session durations.

ktonon:
Session duration will still be restricted by AWS role max session duration.

See https://d2l.slack.com/archives/C728CNC4F/p1652801825925589


ktonon:
Looking into updating the config with a default duration value...

ktonon:
> As far as I can tell the SAML assertion doesn't contain any metadata that gives this information, so its not something we can address without storing some metadata re: roles that we ship with BMX internally etc...

I've left the default as 60 minutes, so the update shouldn't cause any existing usage to suddenly break. We might be able to retrieve the max session duration (in the case that the request fails for this reason). The `Role` type has a `MaxSessionDuration` value

https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#Role

ktonon:
I managed to retrieve the max session duration for the role being assumed, but I had to first re-attempt to assume the role with saml, using the minimum session duration of 900 seconds. Using the temporary credentials we can make an IAM call to "get role" and retrieve the max duration. This assumes that 900 is always safe to use, and that the assumed role has permission to "get role".

If this approach is acceptable. I can cleanup the code and amend this PR

```shell
$ ./bmx print --duration=61
2022/05/17 13:28:52 max session duration for role arn:aws:iam::803669300714:role/Dev-Enterprise-Admin is 3600
2022/05/17 13:28:52 ValidationError: The requested DurationSeconds exceeds the MaxSessionDuration set for this role.
        status code: 400, request id: 290897e4-5df8-49a9-9789-3bf3f6fbf8e7
```

saml/serviceProviders/aws/provider.go
```diff
@@ -76,9 +76,10 @@ func (a AwsServiceProvider) GetCredentials(saml string, desiredRole string) *sts
 	}
 
 	samlInput := &sts.AssumeRoleWithSAMLInput{
-		PrincipalArn:  aws.String(role.Principal),
-		RoleArn:       aws.String(role.ARN),
-		SAMLAssertion: aws.String(saml),
+		DurationSeconds: aws.Int64(int64(durationInMinutes) * 60),
```
neverendingqs, line 79: Double-check: this will just work with any config files as well? https://github.com/Brightspace/bmx#configuration-parameters

If so, could the README be updated as well?

neverendingqs (commented):
Thanks! I'm not experienced with the code or Golang, but LGTM.

saml/serviceProviders/aws/provider.go
ktonon: Yes, I'll update the readme

cfbao:
> I managed to retrieve the max session duration for the role being assumed, but I had to first re-attempt to assume the role with saml, using the minimum session duration of 900 seconds. Using the temporary credentials we can make an IAM call to "get role" and retrieve the max duration. This assumes that 900 is always safe to use, and that the assumed role has permission to "get role".
> 
> If this approach is acceptable. I can cleanup the code and amend this PR
> 
> ```shell
> $ ./bmx print --duration=61
> 2022/05/17 13:28:52 max session duration for role arn:aws:iam::803669300714:role/Dev-Enterprise-Admin is 3600
> 2022/05/17 13:28:52 ValidationError: The requested DurationSeconds exceeds the MaxSessionDuration set for this role.
>         status code: 400, request id: 290897e4-5df8-49a9-9789-3bf3f6fbf8e7
> ```

I think we should keep it simple and not do detection this way.
While it's typically the case at D2L, there's still no guarantee that the role we're assuming even has `iam:GetRole` permission. Also, bmx is an open source project that can be used at other orgs.
Leaving the responsibility to the org & user when they're not using the default max session duration seems like an ok position to take.

cfbao (approved):

ktonon:
How does versioning work? Do I need to make any changes here to trigger a release?

boarnoah (approved):
I believe you just make a release and have it kick off CD

https://github.com/Brightspace/bmx/blob/7c236449635eb76adf93340ec7cb54d9583a66be/.github/workflows/release.yml#L2-L5

PR #223, opened 2023-02-02
dotnet7: create barebones project

gord5500:
### Why

rewrite for dotnet 7

I think I have detached the main branch properly and have it using the same protections as master brranch


.github/workflows/publish.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v1
+      - name: install linux prerequisites
+        if: matrix.platform == "linux"
+        run: |
+          sudo apt-get install clang zlib1g-dev
+          sudo apk add clang build-base zlib-dev
+      - name: publish
+        run: dotnet publish -r ${{ matrix.platform }}-${{ matrix.architecture }} -c Release
```
gord5500, line 39: https://github.com/dotnet/runtime/blob/main/src/coreclr/nativeaot/docs/compiling.md#cross-architecture-compilation


The bottom of that section has this and I'm not sure how to find that directory on a github worker.
> You also need to specify the sysroot directory for Clang using the SysRoot property. For example, assuming you are using one of ARM64-targeting [Docker images](https://github.com/dotnet/runtime/blob/main/docs/workflow/building/coreclr/linux-instructions.md#Docker-Images) employed for cross-compilation by this repo, you may publish for linux-arm64 with the following command:
>dotnet publish -r linux-arm64 -c Release -p:CppCompilerAndLinker=clang-9 -p:SysRoot=/crossrootfs/arm64

.github/workflows/publish.yml
```diff
@@ -0,0 +1,39 @@
...
+          - machine: macos-latest
+            platform: osx
+          - machine: ubuntu-20.04
+            platform: linux
+    runs-on: ${{ matrix.machine }}
+    timeout-minutes: 5
+    steps:
+      - name: checkout
+        uses: actions/checkout@v1
+      - name: install linux prerequisites
```
gord5500, line 33: https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/#prerequisites

src/D2L.Bmx.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<PublishAot>true</PublishAot>
```
gord5500, line 4: https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/#publish-native-aot---cli

src/D2L.Bmx.csproj
```diff
@@ -0,0 +1,18 @@
...
+		<TargetFramework>net7.0</TargetFramework>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<Nullable>enable</Nullable>
+	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
+		<PackageReference Include="runtime.linux-x64.Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
+		<PackageReference Include="runtime.osx-x64.Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
+		<PackageReference Include="runtime.win-x64.Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
```
gord5500, line 15: https://github.com/dotnet/runtime/blob/main/src/coreclr/nativeaot/docs/compiling.md#cross-architecture-compilation

src/Program.cs
```diff
@@ -0,0 +1,3 @@
+
+// See https://aka.ms/new-console-template for more information
+Console.WriteLine( "Hello, World!" );
```
gord5500, line 3: Need some sort of code for it to build

.github/workflows/publish.yml
gord5500: Should we be using those docker files somehow

.github/workflows/publish.yml
```diff
@@ -0,0 +1,39 @@
...
+      matrix:
+        machine: [windows-latest, macos-latest, ubuntu-20.04]
+        architecture: [x64, arm64]
+        include:
+          - machine: windows-latest
+            platform: win
+          - machine: macos-latest
+            platform: osx
+          - machine: ubuntu-20.04
+            platform: linux
```
cfbao, line 27: don't think this is a valid job matrix

.github/workflows/publish.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ${{ matrix.machine }}
+    timeout-minutes: 5
+    steps:
+      - name: checkout
+        uses: actions/checkout@v1
+      - name: install linux prerequisites
+        if: matrix.platform == "linux"
+        run: |
+          sudo apt-get install clang zlib1g-dev
+          sudo apk add clang build-base zlib-dev
```
cfbao, line 37: you're mixing instructions for two different distros here. only the first line applies for Ubuntu

src/D2L.Bmx.sln
```diff
@@ -0,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
cfbao, line 2: let's put the sln file at the repo root

.github/workflows/publish.yml
cfbao: if it's too much hassle we can forego cross-arch compilation for now.
we don't need arm builds for feature parity with the Go version.

.github/workflows/publish.yml
cfbao: this needs to be different on windows runners though?

gord5500: Yea I tried googling on the way to install these and wasn't finding anything
 >Desktop development with C++ workload.
My thought was can I just use what's done in the example docker file to set it up https://github.com/dotnet/samples/blob/main/core/nativeaot/HelloWorld/Dockerfile.windowsservercore-x64#L5-L6

cfbao: the Windows runner might already have this preinstalled

cfbao: (you need an extra empty line after a block quote to exit block quote mode)

gord5500: Hey you're right. Linux doesn't need the prerequisites installed either 🎉

.github/workflows/publish.yml
gord5500: Yea I'm a little confused while reading the docs. Maybe I can look into it again later on

.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
...
+    strategy:
+      matrix:
+        machine: [windows-latest, macos-latest, ubuntu-20.04]
+        architecture: [x64]
+        include:
+          - machine: windows-latest
+            platform: win
+          - machine: macos-latest
+            platform: osx
+          - machine: ubuntu-20.04
```
cfbao, line 26: why not latest for linux?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
...
+            platform: linux
+    runs-on: ${{ matrix.machine }}
+    timeout-minutes: 5
+    steps:
+      - name: checkout
+        uses: actions/checkout@v1
```
cfbao, line 32: we can (and probably should) still use https://github.com/Brightspace/third-party-actions here

.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: publish
+        working-directory: src
+        run: dotnet publish -r ${{ matrix.platform }}-${{ matrix.architecture }} -c Release
```
cfbao, line 35: don't put `${{ ... }}` directly in scripts. remember what we learned about code injection in github actions?
while we do have control over the variables here, it's best to avoid this pattern entirely.

.gitignore
```diff
@@ -0,0 +1,8 @@
+# DotNet7
+**/bin
+**/obj
+**/.vs
+.vs/
+
+# Rider
+.idea/
```
cfbao, line 8: nit: no final new line? that's not in accordance with our .editorconfig

src/D2L.Bmx.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<PublishAot>true</PublishAot>
+		<OutputType>Exe</OutputType>
+		<TargetFramework>net7.0</TargetFramework>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<Nullable>enable</Nullable>
```
cfbao, line 8:
```suggestion
		<Nullable>enable</Nullable>
		<WarningsAsErrors>nullable</WarningsAsErrors>
```

src/D2L.Bmx.csproj
```diff
@@ -0,0 +1,21 @@
...
+	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
+		<PackageReference Include="runtime.linux-x64.Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
+		<PackageReference Include="runtime.linux-arm64.Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
+		<PackageReference Include="runtime.osx-x64.Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
+		<PackageReference Include="runtime.osx-arm64.Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
+		<PackageReference Include="runtime.win-x64.Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
+		<PackageReference Include="runtime.win-arm64.Microsoft.DotNet.ILCompiler" Version="8.0.0-preview.2.23101.9" />
```
cfbao, line 18: do we need the arm ones if we're not cross compiling to arm?

src/D2L.Bmx.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
cfbao, line 1: please put the project in a folder with the same name.
i.e. files should look like
```
src/D2L.Bmx/D2L.Bmx.csproj
src/D2L.Bmx/*.cs
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
...
+  pull_request:
+    paths:
+      - "src/**"
+      - ".github/workflows/publish.yml"
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        machine: [windows-latest, macos-latest, ubuntu-20.04]
```
cfbao, line 19: do we need both this line and the `machine` in `include`? maybe this one can go?

.github/workflows/publish.yml
boarnoah: This is the latest LTS seems like a good choice. Although considering most the other dependencies are experimental 😆

.github/workflows/publish.yml
cfbao: also, since we're running on different OSes here, and different OSes have different default shell, we should explicitly specify `shell:` and pick one.

.github/workflows/publish.yml
gord5500: The version is built on is the minimum required to run. So if someone tried to run on ubuntu 18 they wouldn't be able to. But 20 and ahead they could. That's why I didn't pick latest for linux

.github/workflows/publish.yml
gord5500: How I understand the include part to work is that if the key value pairs in the matrix match the combinations in include the correct platform value is added to the matrix. 

So I think if we removed the `machine` under `matrix` it would always choose the first entry in `include` which is windows

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations

I think if we ever go back and look at arm64 we could use `include` for extra parameters at the end like is required for linux arm64 building. If we wanted to get rid of `include` I think the matrix supports passing it arrays https://github.com/orgs/community/discussions/24981

src/D2L.Bmx.csproj
gord5500: No, I'll get rid of these now

.github/workflows/publish.yml
gord5500: I would have picked ubuntu 18 but when looking at the github workers they said to use either 20 or 22

Looks like it's actually gone now 
https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

.github/workflows/publish.yml
cfbao: > So I think if we removed the `machine` under `matrix` it would always choose the first entry in include which is windows

I don't think this is true. If there's no match, objects in `include` are all added to the list. From the docs:

> For each object in the `include` list, the key:value pairs in the object will be added to each of the matrix combinations if none of the key:value pairs overwrite any of the original matrix values.

We have quite a few examples of having no parameter directly under `matrix` and relying purely on `include`, and all objects in `include` are run.

gord5500: Oh yea I see in slims. You're right then I think we are good to remove the line 19

.github/workflows/publish.yml
cfbao: aaarrrgghh...

We can worry about this later, but eventually we'll definitely need Docker builds. E.g. Alpine definitely needs its own special builds, and likely Amazon Linux too. With Docker builds we can support older Ubuntu versions too...

D2L.Bmx.sln
```diff
@@ -0,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.4.33213.308
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "D2L.Bmx", "src\D2L.Bmx.csproj", "{DFACA1A3-6CD4-412A-9841-50468E3E8EF4}"
```
cfbao, line 6: path needs to be updated

cfbao (approved):

D2L.Bmx.sln
gord5500: Ah, forgot to save in visual studio

cfbao (approved):

gord5500:
Thanks Chenfeng

PR #226, opened 2023-02-17
dotnet7: Recreate dotnet cli

jcard313:
### Why

Recreates the base CLI using dotnet. Does not yet read from config files.

To run the CLI, use the BMX commands as usual, except replace `bmx` with `dotnet run -- `

Ex. `bmx print --org d2l` becomes `dotnet run -- print --org d2l`

### DIFF
--- a/src/D2L.Bmx/Authenticate.cs
+++ b/src/D2L.Bmx/Authenticate.cs
@@ -1,0 +1,25 @@
+public class Authenticator {
+	public static void Authenticate( string? org, string? user, bool nomask ) {
+		// TODO: Add authenticating from cache
+
+		Console.Write( "Okta Password: " );
+		var password = "";
+		if( nomask ) {
+			password = Console.ReadLine();
+		} else {
+			while( true ) {
+				var input = Console.ReadKey( intercept: true );
+				if( input.Key == ConsoleKey.Enter ) {
+					Console.Write( '\n' );
+					break;
+				} else if( input.Key == ConsoleKey.Backspace && password.Length > 0 ) {
+					password = password.Remove( password.Length - 1, 1 );
+				} else if( !char.IsControl( input.KeyChar ) ) {
+					password += input.KeyChar;
+				}
+			}
+		}
+
+		// TODO: Add authenticaion into Okta Client. If error, print and terminate.
+	}
+}
--- a/src/D2L.Bmx/Print.cs
+++ b/src/D2L.Bmx/Print.cs
@@ -1,0 +1,70 @@
+public class Print {
+
+	public static void PrintHandler( string? org, string? user, string? account, string? role, int? duration, bool nomask, string? output ) {
+
+		// TODO: Get values of org, user, account, and role from the config file and assign them locally. This way we can check if
+		// values are null to see if they have been set through the commandline or config file
+
+		// ask user to input org if org flag isn't set
+		if( string.IsNullOrEmpty( org ) ) {
+			// TODO: also print out --help output
+			Console.WriteLine( "Error: required flag(s) \"org\"" );
+			return;
+		};
+
+		// ask user to input username if user flag isn't set
+		if( string.IsNullOrEmpty( user ) ) {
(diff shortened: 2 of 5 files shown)


src/D2L.Bmx/Authenticate.cs
```diff
@@ -0,0 +1,25 @@
...
+				var input = Console.ReadKey( intercept: true );
+				if( input.Key == ConsoleKey.Enter ) {
+					Console.Write( '\n' );
+					break;
+				} else if( input.Key == ConsoleKey.Backspace && password.Length > 0 ) {
+					password = password.Remove( password.Length - 1, 1 );
+				} else if( !char.IsControl( input.KeyChar ) ) {
+					password += input.KeyChar;
+				}
+			}
```
cfbao, line 20: 😬 I'd hope there's a prettier way, but a quick search doesn't give me an obvious winner.

src/D2L.Bmx/Print.cs
```diff
@@ -0,0 +1,70 @@
...
+			if( int.TryParse( Console.ReadLine(), out int index ) ) {
+				if( index > accounts.Length || index < 1 ) {
+					Console.WriteLine( "Error: Invalid selection" );
+					return;
+				}
+				account = accounts[index - 1];
+			} else {
+				Console.WriteLine( "Error: Please enter an integer" );
+				return;
+			}
```
cfbao, line 44: this is fine, but there's a slightly simpler way
```suggestion
			if( !int.TryParse( Console.ReadLine(), out int index ) || index > accounts.Length || index < 1 ) {
				Console.WriteLine( "Error: Invalid selection" );
				return;
			} else {
				account = accounts[index - 1];
			}
```

src/D2L.Bmx/Program.cs
```diff
@@ -1,3 +1,73 @@
...
+		orgOption,
+		printOutputOption,
+		roleOption,
+		userOption,
+	};
+
+printCommand.SetHandler(
+	Print.PrintHandler,
+	orgOption, userOption, accountOption, roleOption, durationOption, nomaskOption, printOutputOption
+ );
```
cfbao, line 49: it bothers me that we have a long list of options/handler arguments here, and most of them are of the same type `string?`. seems very prone to accidental mismatch.
I wonder if there's a better way. but this is fine for now.

src/D2L.Bmx/Write.cs
```diff
@@ -0,0 +1,87 @@
...
+			unsetFlags.Add( "\"org\"" );
+		};
+
+		// check if profile flag has been set
+		if( string.IsNullOrEmpty( profile ) ) {
+			unsetFlags.Add( "\"profile\"" );
+		}
+
+		if( unsetFlags.Count != 0 ) {
+			Console.Write( "Error: required flag(s) " );
```
cfbao, line 20: I assume this is the current behaviour of the Go version? If so, this is fine. But it does seem unintuitive to me that missing some flags results in a prompt for input, whereas others result in error exit. I wonder if we can/should change that in the future.

src/D2L.Bmx/Write.cs
```diff
@@ -0,0 +1,87 @@
...
+		};
+
+		// Asks for user password input, or logs them in through caches
+		Authenticator.Authenticate( org, user, nomask );
+
+		// TODO: replace placeholder values with actual values
+		var accounts = new[] { "Dev-Slims", "Dev-Toolmon", "Int-Dev-NDE" };
+		var roles = new[] { "Dev-Slims-ReadOnly", "Dev-Slims-Admin" };
+
+		if( string.IsNullOrEmpty( account ) ) {
```
cfbao, line 47: the account & role selection code in the print & the write commands is identical, right?
may be suitable to refactor into its own function.

cfbao (approved):
LGTM!
Have some comments, but shouldn't block merge.

We'll probably also want to think about the general code organization here, but right now the app may be too simple for us to decide on a clear direction. Some thinking & research ahead.

PR #227, opened 2023-02-17
ci: fix publish workflow

gord5500:
I think because of that architecture value in the matrix the `include` part wasn't working. Or maybe it's something else. I just noticed it was only being run on linux which is the last entry


gord5500:
Seems like it was as this PR has a workflow running for all OS. Maybe it's not even a big deal since we want to use docker soon

cfbao (approved):

PR #228, opened 2023-02-17
add Roslynator and fix nonconformance

cfbao:
We use Roslynator analyzers in all our team's C# codebase and have some rules (already defined in .editorconfig) we enforce at build time.
Noticed that bmx didn't have them.


boarnoah (approved):

cfbao:
@jcard313 this may introduce merge conflicts with your local dev branches.
I recommend you merge in the updated main branch as early as you can.

PR #232, opened 2023-02-24
dotnet7: Refactor cli experience

jcard313:
### Why

Adds handler classes to C# code, refactors SetHandler calls


src/D2L.Bmx/ConsoleWrapper.cs
```diff
@@ -0,0 +1,21 @@
+namespace D2L.Bmx;
+
+internal interface IConsole {
+	string? ReadLine();
+	void Write( string? value );
+	void WriteLine( string? value );
+}
+
+internal class ConsoleWrapper : IConsole {
```
boarnoah, line 9: Don't think this kind of abstraction is necessary.
It would be one thing if it was an application that can serialize data out in different ways (like files vs to std out etc...).
Did you have something in mind that would benefit from this?

src/D2L.Bmx/Program.cs
```diff
@@ -44,10 +46,18 @@
...
+printCommand.SetHandler( ( InvocationContext context ) => {
+	var handler = new PrintHandler( new ConsoleWrapper() );
+	handler.Handle(
+		org: context.ParseResult.GetValueForOption( orgOption ),
+		user: context.ParseResult.GetValueForOption( userOption ),
+		account: context.ParseResult.GetValueForOption( accountOption ),
+		role: context.ParseResult.GetValueForOption( roleOption ),
+		duration: context.ParseResult.GetValueForOption( durationOption ),
+		nomask: context.ParseResult.GetValueForOption( nomaskOption ),
+		output: context.ParseResult.GetValueForOption( writeOutputOption )
```
boarnoah, line 58:
```suggestion
		output: context.ParseResult.GetValueForOption( printOutputOption )
```

src/D2L.Bmx/ConsoleWrapper.cs
jcard313: Hmm I got that abstraction from @cfbao's branch that was used when we discussed bmx's code structure. I should have followed up on the significance of the wrapper before implementing it though. @cfbao is the wrapper needed?

cfbao: Should've clarified - I was just using that as an example of a mockable dependency, not to suggest that we actually want to do this.

I can see its utility if we want to unit test the CLI experience, but probably not worth it given how simple bmx's workflow is.

jcard313: Ok, I have removed all references to ConsoleWrapper

cfbao (approved):

PR #234, opened 2023-02-28
Add config support dotnet cli

jcard313:

### DIFF
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -1,0 +1,84 @@
+#nullable enable
+using System.Security;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.FileProviders.Physical;
+namespace D2L.Bmx;
+
+internal interface IBmxConfigProvider {
+	BmxConfig GetConfiguration();
+}
+internal class BmxConfigProvider : IBmxConfigProvider {
+
+	public BmxConfig GetConfiguration() {
+		// Main config is at ~/.bmx/config
+		var configLocation = Path.Join( Environment.GetFolderPath( Environment.SpecialFolder.UserProfile ), ".bmx",
+			"config" );
+
+		// If set, we recursively look up from CWD (all the way to root) for additional bmx config files (labelled as .bmx)
+		var configBuilder = new ConfigurationBuilder().AddIniFile( configLocation, optional: true );
+
+		bool.TryParse( configBuilder.Build()["allow_project_configs"], out bool shouldReadProjectConfig );
+
+		string? projectConfigPath = null;
+
+		if( shouldReadProjectConfig ) {
+			projectConfigPath = GetProjectConfigPath();
+		}
+
+		if( shouldReadProjectConfig && !string.IsNullOrEmpty( projectConfigPath ) ) {
+			// Default file provider ignores files prefixed with ".", we need to provide our own as a result
+			var fileProvider =
+				new PhysicalFileProvider( Path.GetDirectoryName( projectConfigPath )!, ExclusionFilters.None );
+			configBuilder.AddIniFile( fileProvider, Path.GetFileName( projectConfigPath ), optional: false,
+				reloadOnChange: false );
+		}
+
+		var config = configBuilder.Build();
+
+		// checking if config["default_duration] is null or empty so that exception is not thrown when default duration is not set in config
+		int? defaultDuration = null;
+		if( !string.IsNullOrEmpty( config["default_duration"] ) ) {
+			if( ( !int.TryParse( config["default_duration"], out int configDuration ) || configDuration < 1 ) ) {
+				throw new BmxException( "Invalid duration in config" );
+			}
+			defaultDuration = configDuration;
+		}
+
+		return new BmxConfig(
+			Org: config["org"],
+			User: config["user"],
+			Account: config["account"],
+			Role: config["role"],
+			Profile: config["profile"],
+			Duration: defaultDuration
+		);
+	}
+
+	// Look from cwd up the directory chain all the way to root for a .bmx file
+	private static string? GetProjectConfigPath() {
+		DirectoryInfo? currentDirectory = new DirectoryInfo( Directory.GetCurrentDirectory() );
+
+		while( currentDirectory is not null ) {
+			try {
+				var bmxConfigFiles = currentDirectory.GetFiles( ".bmx" );
+
+				if( bmxConfigFiles.Length > 1 ) {
+					// TODO: Log this
+					return null;
+				}
+
+				if( bmxConfigFiles.Length == 1 ) {
+					return bmxConfigFiles[0].FullName;
+				}
+
+				currentDirectory = currentDirectory.Parent;
+			} catch( SecurityException ) {
+				// TODO: Log this
+				return null;
+			}
+		}
+
+		return null;
+	}
+}
--- a/src/D2L.Bmx/ConsolePrompter.cs
+++ b/src/D2L.Bmx/ConsolePrompter.cs
@@ -1,0 +1,58 @@
+namespace D2L.Bmx;
+
+internal static class ConsolePrompter {
+
+	public static string PromptOrg() {
+		Console.Write( "Okta org: " );
+		var org = Console.ReadLine();
+		if( string.IsNullOrEmpty( org ) ) {
+			throw new BmxException( "Invalid org input" );
+		}
+
+		return org;
+	}
+	public static string PromptProfile() {
+		Console.Write( "AWS profile: " );
+		var profile = Console.ReadLine();
+		if( string.IsNullOrEmpty( profile ) ) {
+			throw new BmxException( "Invalid profile input" );
+		}
+
+		return profile;
+	}
+
+	public static string PromptUser() {
+		Console.Write( "Okta Username: " );
+		var user = Console.ReadLine();
+		if( string.IsNullOrEmpty( user ) ) {
+			throw new BmxException( "Invalid user input" );
+		}
+
+		return user;
+	}
+
+	public static string PromptAccount( string[] accounts ) {
+		Console.WriteLine( "Available accounts:" );
+		for( int i = 0; i < accounts.Length; i++ ) {
+			Console.WriteLine( $"[{i + 1}] {accounts[i]}" );
+		}
+		Console.Write( "Select an account: " );
+		if( !int.TryParse( Console.ReadLine(), out int index ) || index > accounts.Length || index < 1 ) {
+		}
+
+		return accounts[index - 1];
+	}
+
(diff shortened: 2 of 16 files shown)


src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+			// Default file provider ignores files prefixed with ".", we need to provide our own as a result
+			var fileProvider =
+				new PhysicalFileProvider( Path.GetDirectoryName( projectConfigPath )!, ExclusionFilters.None );
+			configBuilder.AddIniFile( fileProvider, Path.GetFileName( projectConfigPath ), optional: false,
+				reloadOnChange: false );
+		}
+
+		var config = configBuilder.Build();
+
+		// checking if config["default_duration] is null or empty so that exception is not thrown when default duration is not set in config
```
cfbao, line 39: nit: personally think the code is clear enough without the comment
```suggestion
```

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+
+		if( shouldReadProjectConfig ) {
+			projectConfigPath = GetProjectConfigPath();
+		}
+
+		if( shouldReadProjectConfig && !string.IsNullOrEmpty( projectConfigPath ) ) {
+			// Default file provider ignores files prefixed with ".", we need to provide our own as a result
+			var fileProvider =
+				new PhysicalFileProvider( Path.GetDirectoryName( projectConfigPath )!, ExclusionFilters.None );
+			configBuilder.AddIniFile( fileProvider, Path.GetFileName( projectConfigPath ), optional: false,
```
cfbao, line 33: why is it necessary to do this here but not at [L19](https://github.com/Brightspace/bmx/pull/234/files#diff-2eb5576e543a9b7cb2ee5a009fd10a20e4bbeb2a3115370f6d75de9403f7deb1R19)?

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+			try {
+				var bmxConfigFiles = currentDirectory.GetFiles( ".bmx" );
+
+				if( bmxConfigFiles.Length > 1 ) {
+					// TODO: Log this
+					return null;
+				}
+
+				if( bmxConfigFiles.Length == 1 ) {
+					return bmxConfigFiles[0].FullName;
```
cfbao, line 72: We can make this method return `FileInfo?` like this, then the caller can use methods on this object, instead of using static methods from `Path` on a string. It'd be more organized and efficient this way.
```suggestion
					return bmxConfigFiles[0];
```

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+			Account: config["account"],
+			Role: config["role"],
+			Profile: config["profile"],
+			Duration: defaultDuration
+		);
+	}
+
+	// Look from cwd up the directory chain all the way to root for a .bmx file
+	private static string? GetProjectConfigPath() {
+		DirectoryInfo? currentDirectory = new DirectoryInfo( Directory.GetCurrentDirectory() );
```
cfbao, line 60: nit: can be simplified
```suggestion
		DirectoryInfo? currentDirectory = new( Directory.GetCurrentDirectory() );
```

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+					// TODO: Log this
+					return null;
+				}
+
+				if( bmxConfigFiles.Length == 1 ) {
+					return bmxConfigFiles[0].FullName;
+				}
+
+				currentDirectory = currentDirectory.Parent;
+			} catch( SecurityException ) {
```
cfbao, line 76: when/why would this happen? worth explaining in comment.

src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -0,0 +1,58 @@
...
+	public static string PromptOrg() {
+		Console.Write( "Okta org: " );
+		var org = Console.ReadLine();
+		if( string.IsNullOrEmpty( org ) ) {
+			throw new BmxException( "Invalid org input" );
+		}
+
+		return org;
+	}
+	public static string PromptProfile() {
```
cfbao, line 14: nit
```suggestion
	}

	public static string PromptProfile() {
```

src/D2L.Bmx/PrintHandler.cs
```diff
@@ -2,6 +2,11 @@ namespace D2L.Bmx;
 
 internal class PrintHandler {
 
+	private readonly IBmxConfigProvider _configProvider;
+
+	public PrintHandler( IBmxConfigProvider configProvider ) {
+		_configProvider = configProvider;
+	}
 	public void Handle(
```
cfbao, line 10: nit
```suggestion
	}

	public void Handle(
```

src/D2L.Bmx/PrintHandler.cs
```diff
@@ -12,54 +17,56 @@ internal class PrintHandler {
...
+				role = ConsolePrompter.PromptRole( roles );
 			}
-			Console.Write( "Select a role: " );
-			if( !int.TryParse( Console.ReadLine(), out int index ) || index > roles.Length || index < 1 ) {
-				Console.WriteLine( "Error: Invalid selection" );
-				return;
+		}
+
+		// if duration equals zero, then the duration flag is not set. Users cannot set duration to 0 because of flag validation
+		if( duration == 0 ) {
```
cfbao, line 64: Wait... the default value for the duration flag is 0 not null? Our signature `int? duration` is a lie then...

Right... this is a consequence of the ambiguity of `T?` in generics. I'd prefer we make `durationOption` an `Option<int?>` then (not `Option<int>`), then we'd do a null check instead of a zero check here, and the code should be clearer & we won't need the comment.

cfbao: More context:
* In generic code, `T` and `T?` mean the same thing for value types (e.g. `int`) (i.e. the question mark is meaningless), but they do have different meanings for reference types (e.g. `string`).
* This is in part because reference type nullability is only a compile time construct, not a runtime constraint - `string` and `string?` are the same type. On the other hand, value types can never be null, so `int?` is actually an entirely different runtime type [`Nullable<int>`](https://learn.microsoft.com/en-us/dotnet/api/system.nullable-1?view=net-7.0)
* The signature of [`GetValueForOption`](https://learn.microsoft.com/en-us/dotnet/api/system.commandline.parsing.parseresult.getvalueforoption?view=system-commandline) says it should return an object of type `T` and it _may_ be null, which works intuitively for reference types, but for value types, because they can never be null, "may be null" is meaningless. It doesn't return `Nullable<int>` for `Option<int>`, because that's a different type from the type parameter `int`. If we pass it an `Option<int?>` though, it should return `int?` i.e. `Nullable<int>` as expected.

src/D2L.Bmx/WriteHandler.cs
```diff
@@ -1,7 +1,11 @@
 namespace D2L.Bmx;
 
 internal class WriteHandler {
+	private readonly IBmxConfigProvider _configProvider;
 
+	public WriteHandler( IBmxConfigProvider configProvider ) {
+		_configProvider = configProvider;
+	}
 	public void Handle(
```
cfbao, line 9: nit
```suggestion
	}

	public void Handle(
```

cfbao (commented):
some of the comments may be better answered by @boarnoah I suppose

cfbao:
On PR title & commit message - I don't think it worthwhile to add the `dotnet7` prefix to these PRs.
After the rewrite is done, we will switch the main (i.e. dotnet7) branch to be the default branch and hide the Go version entirely, then the `dotnet7` prefix will look redundant.

If desired, we can add a special dotnet7 label to these PRs to differentiate them from any maintenance on the Go version.

src/D2L.Bmx/BmxConfigProvider.cs
boarnoah: IIRC this can happen when walking up the directory tree, in some scenarios.  Since its not immediately obvious in which situations this can occur maybe we get rid of the graceful handling for now.

src/D2L.Bmx/BmxConfigProvider.cs
boarnoah: Do you mean the bit about passing a different fileProvider? Its to handle reading project configs that are in ``.bmx``files, since those are ignored by the Configuration Builder as hidden files.

cfbao: yes, but at L19 aren't we also reading a ".bmx" file? why is it fine to not pass a custom configured file provider there?

boarnoah: L19 is reading the ``~/.bmx/config`` file

jcard313: Ok so from my understanding L19 doesn't need to be changed then. Should the `allow_project_configs` logic just be removed now or does it require more discussion?

cfbao: so... the difference is between .bmx _folder_ & .bmx _file_?
i.e. the default file provider includes folders starting with a dot, but not files starting with a dot?

src/D2L.Bmx/BmxConfigProvider.cs
jcard313: Should I get rid of the try catch entirely then? We can also throw a BmxException instead

src/D2L.Bmx/BmxConfigProvider.cs
boarnoah: > Should the allow_project_configs logic just be removed now or does it require more discussion?

Sorry I meant in the future, for now we should keep this for 1:1 with current BMX

> i.e. the default file provider includes folders starting with a dot, but not files starting with a dot?

yes

cfbao: Let's keep the `allow_project_configs` config for now. I know it's used by some to disable project config lookup to resolve some infinite recursion issue in the Go version on Linux/Mac machines. The current code here could suffer from the same problem. I wouldn't remove it unless we can be sure people won't be upset without it.

src/D2L.Bmx/BmxConfigProvider.cs
cfbao: yes? @boarnoah

cfbao (approved):
FYI, I updated the PR title to get rid of the dotnet7 prefix

src/D2L.Bmx/BmxConfigProvider.cs
boarnoah: Yep, we can add it back if the scenario where this was needed occurs (which I can't replicate ATM)

cfbao (approved):

PR #237, opened 2023-03-07
CI: Use Docker for linux builds

gord5500:
### Why

Github runners don't cover all the possible operating systems we want to build for linux. We can use dockerfiles to do that instead


.github/workflows/publish.yml
```diff
@@ -33,4 +31,4 @@ jobs:
         env:
           PLATFORM: ${{ matrix.platform }}
         working-directory: src/D2L.Bmx
-        run: dotnet publish -r $env:PLATFORM-x64 -c Release
+        run: dotnet publish -r $env:PLATFORM-x64 -c Release -o out
```
gord5500, line 34: This saves the executable to `out` folder to match the docker workflow

gord5500:
There's no official image for older versions of ubuntu / aws linux. Do we want to be creating new images that use those operating systems and install dotnet ourselves and then publish?

gord5500:
Maybe `Dockerfile` could be renamed to something else since it works for both debian and ubuntu. It's just the package manager alpine uses that made me use a separate one

.github/workflows/publish_docker.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: Brightspace/third-party-actions@actions/checkout
+      - name: setup docker buildx
+        uses: docker/setup-buildx-action@v2
+      - name: publish
+        shell: pwsh
+        env:
+          TAG: ${{ matrix.tag }}
+          DOCKERFILE: ${{ matrix.file }}
+        run: |
+          docker buildx build --build-arg "TAG=$env:TAG" -f $env:DOCKERFILE -o out .
```
gord5500, line 41: I wanted to initially try docker compose and not need to create the workflow like this but they don't support the `output` flag

gord5500: the `-o out` saves the published executable into the out folder on the github runner

gord5500: I thought it could be easier to upload them outside of the dockerfile and in the workflow when we get to that point

.github/workflows/publish_docker.yml
```diff
@@ -0,0 +1,41 @@
...
+        include:
+          - tag: "7.0" #Debian 11
+            file: Dockerfile
+          - tag: 7.0-jammy # Ubuntu 22.04
+            file: Dockerfile
+          - tag: 7.0-alpine # Alpine 3.17
+            file: Dockerfile.alpine
+          - tag: 7.0-alpine3.16
+            file: Dockerfile.alpine
+    runs-on: ubuntu-latest
```
scowing, line 28: I think we always want to specify our self hosted runners right? Otherwise is uses a github hosted runner?

gord5500: Yea bmx is public repo so we can't even access our self hosted runners

gord5500: I was waiting for a workflow to finish on the master branch last week and took like 30 minutes before figuring out we can't use self hosted anymore 😅

scowing: Oh yeah I forgot about the public repo thing

gord5500:
For amazon linux 2, the glibc package is too outdated to run the Ubuntu version. Maybe I can try the a dockerfile for it specifically and see what happens? Amazon Linux 2022 has a new enough version but I can't see how to spin one of those up at least in the aws console

.github/workflows/publish_docker.yml
```diff
@@ -0,0 +1,41 @@
+name: publish docker
```
cfbao, line 1: do we need a new workflow? another job within the original publish workflow should be enough I think.

.github/workflows/publish_docker.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        include:
+          - tag: "7.0" #Debian 11
+            file: Dockerfile
+          - tag: 7.0-jammy # Ubuntu 22.04
+            file: Dockerfile
```
cfbao, line 23: Do we need two different builds for them? If they use the same glibc I'd assume one build would be sufficient?
When we release bmx, ideally we release as few binaries as possible to support as many OSes as possible.

.github/workflows/publish_docker.yml
```diff
@@ -0,0 +1,41 @@
...
+            file: Dockerfile.alpine
+          - tag: 7.0-alpine3.16
+            file: Dockerfile.alpine
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: checkout
+        uses: Brightspace/third-party-actions@actions/checkout
+      - name: setup docker buildx
+        uses: docker/setup-buildx-action@v2
```
cfbao, line 34: Do we need docker buildx here? I thought that's only for cross-arch builds.
Cross-arch builds would be nice for sure, I just want to confirm if we are using some other features of buildx, because we aren't building across arch here.

cfbao:
> There's no official image for older versions of ubuntu / aws linux for dotnet 7 sdk. Do we want to be creating new images that use those operating systems and install dotnet ourselves and then publish?

I'd like to. E.g. Ubuntu 22 is fairly new, and there are definitely many older versions of Ubuntu floating around.

> For amazon linux 2, the glibc package is too outdated to run the Ubuntu version. Maybe I can try the a dockerfile for it specifically and see what happens? Amazon Linux 2022 has a new enough version but I can't see how to spin one of those up at least in the aws console

Amazon Linux 2022 is basically irrelevant here. We need to support Amazon Linux 2. Being able to use Amazon Linux 2022 doesn't give us much.

> Maybe `Dockerfile` could be renamed to something else since it works for both debian and ubuntu. It's just the package manager alpine uses that made me use a separate one + the target build is different than just `linux-x64`

Renaming it would be good.

> Think it would be easy to expand on this workflow for arm64 images. I think Linux was the problem previously in regards to doing that. Maybe it can wait though until we get to the point of uploading the binaries

Yeah arm64 can wait. It's not needed for feature parity.

.github/workflows/publish_docker.yml
gord5500: buildx is for the `-o` flag.  On my local machine I could have just `build` work plus setting the $DOCKER_BUILDKIT env var set to 1 but it wasn't saving to the runner's disk.  Thought it would be easier to have the binary on the runner and not do anything with volumes. I think buildx is just to use the buildkit backend instead of the default

.github/workflows/publish_docker.yml
gord5500: I didn't realize that `glibc` package was what mattered and assumed they would need to be separate for each OS. I can spin up an ec2 instance for each and try them

.github/workflows/publish_docker.yml
gord5500: Yea I guess what I did is the equivalent as just doing `buildx ...`. This flag only works when using the buildkit backend https://docs.docker.com/engine/reference/commandline/build/#output

.github/workflows/publish_docker.yml
gord5500: Yea you are probably right. I think I separated just for testing. I'll look to combine them

Dockerfile.ubuntu
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:18.04 AS build
+
+RUN apt-get update && apt-get install -y --no-install-recommends wget ca-certificates
+RUN wget https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb \
+    --no-check-certificate -O packages-microsoft-prod.deb
+RUN dpkg -i packages-microsoft-prod.deb
+RUN rm packages-microsoft-prod.deb
+
```
gord5500, line 8: https://learn.microsoft.com/en-us/dotnet/core/install/linux-ubuntu#1804

bit of a process

Dockerfile.alpine
```diff
@@ -0,0 +1,19 @@
+FROM amd64/alpine:3.15 AS build
+
+RUN apk update \
+    && apk add clang build-base bash icu-libs krb5-libs \
```
gord5500, line 4: https://learn.microsoft.com/en-us/dotnet/core/install/linux-alpine#dependencies

Dockerfile.alpine
```diff
@@ -0,0 +1,19 @@
...
+    libgcc libintl libssl1.1 libstdc++ zlib zlib-dev
```
gord5500, line 5: https://learn.microsoft.com/en-us/dotnet/core/install/linux-alpine#dependencies

Dockerfile.alpine
```diff
@@ -0,0 +1,19 @@
...
+
+RUN mkdir -p /usr/share/dotnet && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
```
gord5500, line 7: https://linuxize.com/post/how-to-create-symbolic-links-in-linux-using-the-ln-command/

If I tried to install on line 11 to `usr/bin/dotnet` when I tried to publish I would get permission denied error. Is there another way to do this?

Dockerfile.alpine
```diff
@@ -0,0 +1,19 @@
...
+
+RUN wget https://dot.net/v1/dotnet-install.sh \
+    && chmod +x ./dotnet-install.sh \
+    && ./dotnet-install.sh --channel 7.0 --install-dir /usr/share/dotnet
+
```
gord5500, line 12: https://learn.microsoft.com/en-us/dotnet/core/install/linux-scripted-manual

Dockerfile.alpine
```diff
@@ -0,0 +1,19 @@
+FROM amd64/alpine:3.15 AS build
```
gord5500, line 1: 3.15 is the oldest version dotnet 7 supports

gord5500:
Are there other linux distros we want to support? Right now it's ubuntu, debian, amazon linux 2, and alpine

cfbao (dismissed):

cfbao:
> Are there other linux distros we want to support? Right now it's ubuntu, debian, amazon linux 2, and alpine. The first three all work on the ubuntu dockerfile build

These are the main ones.
In principle, any other distros that use compatible C libraries should work with the build. Most distros use the same C libraries. It's only Alpine that's special.

I think this is good enough now. We can update later if people report having problems.

cfbao (approved):

gord5500:
Thanks for the help, Chenfeng 🎉

PR #238, opened 2023-03-14
Add aws from dotnet branch

jcard313:


src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -0,0 +1,75 @@
+using System.Text;
+using System.Web;
+using System.Xml;
+using Amazon.SecurityToken;
+using Amazon.SecurityToken.Model;
+
+namespace D2L.Bmx.Aws;
+
+public interface ICloudProvider<TRoleState> where TRoleState : IRoleState {
```
cfbao, line 9: * the heavy use of generics and abstraction is one aspect of the `dotnet` branch that we want to get away from by this new rewrite
* we should default to making types internal rather than public

```suggestion
internal interface IAwsClient {
```

src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -0,0 +1,75 @@
...
+	TRoleState GetRoles( string encodedSaml );
+	Task<Dictionary<string, string>> GetTokensAsync( TRoleState state, int selectedRoleIndex );
+}
+public class AwsClient : ICloudProvider<AwsRoleState> {
```
cfbao, line 13: * nit: blank line between top level blocks
* the class should also be internal

```suggestion
}

internal class AwsClient : IAwsClient {
```

src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -0,0 +1,75 @@
...
+	private readonly IAmazonSecurityTokenService _stsClient;
+
+	public AwsClient( IAmazonSecurityTokenService stsClient ) {
+		_stsClient = stsClient;
+	}
+
+	private XmlDocument ParseSamlToken( string encodedSaml ) {
```
cfbao, line 20: nit: conventional to put private members after public members.

src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -0,0 +1,75 @@
...
+		var samlStatements = encodedSaml.Split( ";" );
+		// Process the B64 Encoded SAML string to get valid XML doc
+		var samlString = new StringBuilder();
+		foreach( var inputValueString in samlStatements ) {
+			samlString.Append( HttpUtility.HtmlDecode( inputValueString ) );
```
cfbao, line 25: `HttpUtility` is from the `System.Web` namespace, which is... legacy, and is even shitted on the .NET team themselves.

We can use [WebUtility](https://learn.microsoft.com/en-us/dotnet/api/system.net.webutility?view=net-7.0) as an alternative.

src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -0,0 +1,75 @@
...
+			PrincipalArn = role.PrincipalArn,
+			RoleArn = role.RoleArn,
+			SAMLAssertion = state.SamlString
+		} );
+
+		return new Dictionary<string, string> {
+				{"AWS_SESSION_TOKEN", authResp.Credentials.SessionToken},
+				{"AWS_ACCESS_KEY_ID", authResp.Credentials.AccessKeyId},
+				{"AWS_SECRET_ACCESS_KEY", authResp.Credentials.SecretAccessKey}
+			};
```
cfbao, line 73: may be worth making a type for this? or is there value in keeping it free form?
depends on how it's used of course.

src/D2L.Bmx/Aws/AwsRole.cs
```diff
@@ -0,0 +1,7 @@
+namespace D2L.Bmx.Aws;
+
+public struct AwsRole {
```
cfbao, line 3: let's make it a record with a primary constructor like the rest of our new C# code.
also - make it internal

src/D2L.Bmx/Aws/AwsRoleState.cs
```diff
@@ -0,0 +1,17 @@
+namespace D2L.Bmx.Aws;
+
+public interface IRoleState {
+	string[] Roles { get; }
+}
+
+public class AwsRoleState : IRoleState {
```
cfbao, line 7: I don't see the need for an interface here. the "role state" is just a data object right? so record with primary constructor?

src/D2L.Bmx/Aws/AwsClient.cs
jcard313: Looking at how it's used in the dotnet branch, there doesn't seem to be any specific reason that it doesn't have a type, so I'll create a type for this

src/D2L.Bmx/Aws/AwsRoleState.cs
```diff
@@ -0,0 +1,13 @@
...
+internal class AwsRoleState {
+	public string?[] Roles { get; }
+	internal List<AwsRole> AwsRoles { get; }
+	internal string SamlString { get; }
+
+	public AwsRoleState( List<AwsRole> awsRoles, string samlString ) {
+		AwsRoles = awsRoles;
+		SamlString = samlString;
+		Roles = AwsRoles.Select( role => role.RoleName ).ToArray();
+	}
```
cfbao, line 12: * All "public" members inside an internal class are effectively "internal", so people typically mark them as "public" instead of "internal"
* If `Roles` are derived from `AwsRoles`, do we really need `Roles`? Can we we simplify this type to just two properties and use record with primary constructor?

cfbao: oh! I see you've actually made that change in your "okta" PR.
would be better if that's in this PR, but this is fine too.

jcard313: So keep AwsRoles as a property and separate Roles out as it's own record?

cfbao: No, I meant deleting the `Roles` property entirely.
So what you did here is good I think:
https://github.com/Brightspace/bmx/pull/239/files#diff-5457c3da07d8bb435dcdf1616bc3b67556b5b83a0454143321babfebdaa61346

src/D2L.Bmx/Aws/AwsRoleState.cs
```diff
@@ -0,0 +1,6 @@
+namespace D2L.Bmx.Aws;
+
+internal record AwsRoleState(
```
boarnoah, line 3: I think based on how hooking this up to the CLI side goes, we can eliminate the need to pass application state around like this, but this is fine for now.

PR #239, opened 2023-03-16
Add okta from dotnet branch

jcard313:


src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -0,0 +1,111 @@
...
+	private readonly HttpClient _httpClient;
+
+	public OktaApi() {
+		_cookieContainer = new CookieContainer();
+		_httpClient = new HttpClient( new HttpClientHandler { CookieContainer = _cookieContainer } );
+		_httpClient.Timeout = TimeSpan.FromSeconds( 30 );
+		_httpClient.DefaultRequestHeaders.Accept.Add( new MediaTypeWithQualityHeaderValue( "application/json" ) );
+	}
+
+	public void SetOrganization( string organization ) {
```
gord5500, line 42:
```suggestion
	public void IOktaApi.SetOrganization( string organization ) {
```

I think [this](https://github.com/Brightspace/instance-metadata-store/pull/643#discussion_r1124738319) could apply here?

src/D2L.Bmx/Okta/OktaClient.cs
gord5500, line 1: Same for this class

src/D2L.Bmx/Okta/OktaApi.cs
gord5500: 🤝

gord5500 (dismissed):

src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
```diff
@@ -0,0 +1,31 @@
...
+
+internal class AuthenticateResponse {
+	public DateTimeOffset ExpiresAt { get; set; }
+
+	[JsonConverter( typeof( JsonStringEnumConverter ) )]
+	public AuthenticationStatus Status { get; set; }
+}
+
+internal class AuthenticateResponseInital : AuthenticateResponse {
+	public string? StateToken { get; set; }
```
cfbao, line 12: I believe source generated json serializer should work with record's primary constructor, so we don't need the setters here, and can use records.
You can set property attributes on a primary constructor like this:
https://github.com/Brightspace/oauth2-iam-grant-libraries/blob/eb2d0b58b873cfd3f91cb8f4a90414473341c54d/src/D2L.IamGrant.Client/IamGrantLambdaResponse.cs

cfbao: Also I'm not sure if we really need the base class/inheritance structure here. Do we actually need these properties? @boarnoah

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -0,0 +1,111 @@
...
+[JsonSourceGenerationOptions( PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase )]
+[JsonSerializable( typeof( AuthenticateOptions ) )]
+[JsonSerializable( typeof( AuthenticateChallengeMfaOptions ) )]
+[JsonSerializable( typeof( SessionOptions ) )]
+[JsonSerializable( typeof( AuthenticateResponseInital ) )]
+[JsonSerializable( typeof( AuthenticateResponseSuccess ) )]
+[JsonSerializable( typeof( OktaSession ) )]
+[JsonSerializable( typeof( OktaApp[] ) )]
+internal partial class SourceGenerationContext : JsonSerializerContext {
+}
```
cfbao, line 28: this seems fit to live in its own file

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -0,0 +1,111 @@
...
+			_cookieContainer.Add( new Cookie( "sid", sessionId, "/", _httpClient.BaseAddress.Host ) );
+		} else {
+			throw new BmxException( "Error adding session: http client base address is not defined" );
+		}
+	}
+
+	async Task<AuthenticateResponseInital> IOktaApi.AuthenticateOktaAsync( AuthenticateOptions authOptions ) {
+		var resp = await _httpClient.PostAsync( "authn",
+			new StringContent(
+				JsonSerializer.Serialize( authOptions!, SourceGenerationContext.Default.AuthenticateOptions ),
```
cfbao, line 57: don't think we need the null forgiving operator here
```suggestion
				JsonSerializer.Serialize( authOptions, SourceGenerationContext.Default.AuthenticateOptions ),
```

src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -0,0 +1,89 @@
...
+	void SetOrganization( string organization );
+	Task<OktaAuthenticateState> AuthenticateAsync( string username, string password );
+
+	Task<OktaAuthenticatedState> ChallengeMfaAsync( OktaAuthenticateState state, int selectedMfaIndex,
+		string challengeResponse );
+
+	Task<OktaAccountState> GetAccountsAsync( OktaAuthenticatedState state, string accountType );
+	Task<string> GetServiceProviderSamlAsync( OktaAccountState state, int selectedAccountIndex );
+}
+internal class OktaClient : IOktaClient {
```
cfbao, line 18: nit
```suggestion
}

internal class OktaClient : IOktaClient {
```

src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -0,0 +1,89 @@
+using System.Text.RegularExpressions;
+using System.Xml;
+using D2L.Bmx.Okta.Models;
+using D2L.Bmx.Okta.State;
+namespace D2L.Bmx.Okta;
+
+internal interface IOktaClient {
+	public string Name { get; }
```
cfbao, line 8: do we still need this at all? @boarnoah 
```suggestion
```

src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -0,0 +1,89 @@
...
+	Task<string> GetServiceProviderSamlAsync( OktaAccountState state, int selectedAccountIndex );
+}
+internal class OktaClient : IOktaClient {
+	private readonly IOktaApi _oktaApi;
+
+	public OktaClient( IOktaApi oktaApi ) {
+		_oktaApi = oktaApi;
+	}
+
+	public string Name => "Okta";
```
cfbao, line 25:
```suggestion
```

src/D2L.Bmx/Okta/State/OktaAccountState.cs
```diff
@@ -0,0 +1,14 @@
...
+	public OktaAccountState( OktaApp[] oktaApps, string accountType ) {
+		OktaApps = oktaApps;
+		AccountType = accountType;
+		Accounts = OktaApps.Where( app => app.AppName == AccountType ).Select( app => app.Label ).ToArray();
+	}
+
+	public string[] Accounts { get; }
+	internal string AccountType { get; }
+	internal OktaApp[] OktaApps { get; }
+}
```
cfbao, line 14: similar to this: https://github.com/Brightspace/bmx/pull/238#discussion_r1140566229

cfbao: if `Accounts` is still nice to have, we can move the logic into the getter? e.g.

```csharp
internal record MyType( string PropA ) {
	public string PropB => PropA.ToLower();
}
```

src/D2L.Bmx/Okta/State/OktaAuthenticateState.cs
```diff
@@ -0,0 +1,25 @@
+using D2L.Bmx.Okta.Models;
+namespace D2L.Bmx.Okta.State;
+
+internal class OktaAuthenticateState {
```
cfbao, line 4: similarly for this class

src/D2L.Bmx/Okta/State/OktaAuthenticatedState.cs
```diff
@@ -0,0 +1,11 @@
+namespace D2L.Bmx.Okta.State;
+
+internal class OktaAuthenticatedState {
```
cfbao, line 3: this can be record too?

src/D2L.Bmx/Okta/OktaClient.cs
boarnoah: No, that was another attribute of me trying to over-generalize the Okta provider bit

src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
jcard313: @boarnoah I got rid of the inheritance structure here but I can add it back in if necessary

boarnoah: We can add it back if we want to use a type discriminator to parse response differently (i.e https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism)  if its asks for MFA vs without, we can get to that after we have the initial implementation working.

jcard313: Ok sounds good then. Would we change the records back into classes in that case?

boarnoah: No, they just used classes in the example in the docs but polymorphic de serialization works just fine with records.

src/D2L.Bmx/Okta/Models/MfaType.cs
```diff
@@ -0,0 +1,13 @@
+namespace D2L.Bmx.Okta.Models;
+
+internal record MfaOption(
+	string Name,
+	MfaType Type
+);
+
+// TODO: More robust / generic way of handling MFA Types
+public enum MfaType {
```
boarnoah, line 9: this can be internal as well
```suggestion
public internal MfaType {
```

src/D2L.Bmx/Okta/Models/OktaError.cs
```diff
@@ -0,0 +1,8 @@
+namespace D2L.Bmx.Okta.Models;
+
+internal record OktaError(
+	string ErrorCode,
+	string ErrorSumamry,
```
boarnoah, line 5:
```suggestion
	string ErrorSummary,
```

src/D2L.Bmx/Okta/SourceGenerationContext.cs
```diff
@@ -0,0 +1,14 @@
...
+
+[JsonSourceGenerationOptions( PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase )]
+[JsonSerializable( typeof( AuthenticateOptions ) )]
+[JsonSerializable( typeof( AuthenticateChallengeMfaOptions ) )]
+[JsonSerializable( typeof( SessionOptions ) )]
+[JsonSerializable( typeof( AuthenticateResponseInital ) )]
+[JsonSerializable( typeof( AuthenticateResponseSuccess ) )]
+[JsonSerializable( typeof( OktaSession ) )]
+[JsonSerializable( typeof( OktaApp[] ) )]
+internal partial class SourceGenerationContext : JsonSerializerContext {
```
boarnoah, line 13: Interesting, https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/#why-do-source-generation 

I guess BMX is a good place to trial this sort of thing.

src/D2L.Bmx/Okta/State/OktaAccountState.cs
```diff
@@ -0,0 +1,10 @@
+using D2L.Bmx.Okta.Models;
+namespace D2L.Bmx.Okta.State;
+
+internal record OktaAccountState( OktaApp[] OktaApps, string AccountType ) {
+
+	public string[] Accounts
+		=> this.OktaApps.Where( app => app.AppName == AccountType ).Select( app => app.Label ).ToArray();
```
boarnoah, line 7: I think this could be hardcoded to "amazon_aws" but again, something to do when refactoring further once its all working.

src/D2L.Bmx/Okta/Models/OktaError.cs
boarnoah: EDIT: lets delete OktaError altogether, we can add anything necessary for handling Okta API errors when we get to that.

boarnoah (approved):

PR #254, opened 2023-04-13
Add write operation

jcard313:
### Why

Adds write operation so users can write to credentials files


src/D2L.Bmx/WriteHandler.cs
```diff
@@ -83,15 +89,19 @@ internal class WriteHandler {
...
+			}
+			credentialsFile = new SharedCredentialsFile( output );
+		}
 
+		var profileOptions = new CredentialProfileOptions {
+			Token = tokens.AwsSessionToken,
+			AccessKey = tokens.AwsAccessKeyId,
+			SecretKey = tokens.AwsSecretAccessKey
+		};
+		credentialsFile.RegisterProfile( new CredentialProfile( profile, profileOptions ) );
```
boarnoah, line 105: that worked out rather neatly compared to custom parsing / writing to files. neat

PR #257, opened 2023-04-18
Add additional MFA support

jcard313:
### Why

Now supports SMS, yubikey, Security Questions, and totp from different providers for feature parity with prod BMX.


src/D2L.Bmx/Authenticate.cs
```diff
@@ -38,23 +38,24 @@ internal class Authenticator {
 		if( authState.OktaStateToken is not null ) {
 
 			var mfaOptions = authState.MfaOptions;
-
 			var selectedMfaIndex = PromptMfa( mfaOptions );
 			var selectedMfa = mfaOptions[selectedMfaIndex - 1];
+			string mfaInput = "";
 
+			// TODO: Handle retry
```
jcard313, line 45: Let me know if I should handle retry in this task as well

boarnoah: Nope, current BMX doesn't support retry so its not necessary for feature parity + better suited for a separate feature PR

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -62,14 +63,27 @@ internal class OktaApi : IOktaApi {
...
 		string challengeResponse ) {
 
-
 		var mfaFactor = state.OktaMfaFactors[selectedMfaIndex];
 
-		var authOptions = new AuthenticateChallengeMfaOptions( mfaFactor.Id, challengeResponse, state.OktaStateToken! );
+		var authOptions = new AuthenticateChallengeMfaOptions(
+			FactorId: mfaFactor.Id,
+			StateToken: state.OktaStateToken!,
+			PassCode: challengeResponse );
```
boarnoah, line 86: The security question MFA uses a field name called ``answer`` instead of ``passcode``, does it also accept the answer as passcode? https://developer.okta.com/docs/reference/api/factors/#request-parameters-23

jcard313: Yup I tested it and it does surprisingly

jcard313: Also currently works on the current BMX as well

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -14,6 +14,7 @@ internal interface IOktaApi {
 	Task<OktaAuthenticateState> AuthenticateOktaAsync( AuthenticateOptions authOptions );
 	Task<OktaAuthenticatedState> AuthenticateChallengeMfaOktaAsync( OktaAuthenticateState state, int selectedMfaIndex,
 		string challengeResponse );
+	Task IssueSmsChallengeMfaOktaAsync( OktaAuthenticateState state, int selectedMfaIndex );
```
boarnoah, line 17: For clarity with what this is in Okta (https://developer.okta.com/docs/reference/api/factors/#factors-that-require-a-challenge-and-verify-operation).

Lets rename it to something that reflects that its issuing an MFA challenge, since its not just SMS that follows this flow.

boarnoah (dismissed):

src/D2L.Bmx/Okta/OktaApi.cs
jcard313: IssueChallengeMfaOktaAsync?

boarnoah: IssuMfaChallengeOktaAsync rolls  off the tongue a wee bit better :P . 

We should at some point drop the word Okta from the interface methods, it seems redundant when the interface has Okta in the name.

jcard313: Oh true

jcard313: I can create a PR to do that once this change is in

boarnoah: Yep, I meant that as an aside, no need to block this PR

boarnoah (approved):

PR #259, opened 2023-04-21
cache okta sessions

gord5500:
### Why

Currently we need to type our okta password in every time. Go version supported caching a session id and reusing it until expiry

Related trello card: https://trello.com/c/GGgGonzC/1216-support-storing-session-and-profiles

### DIFF
--- a/src/D2L.Bmx/Authenticate.cs
+++ b/src/D2L.Bmx/Authenticate.cs
@@ -1,4 +1,3 @@
-using System.Text.Json;
 using D2L.Bmx.Okta;
 using D2L.Bmx.Okta.Models;
 using D2L.Bmx.Okta.State;
@@ -38,7 +37,7 @@ internal class Authenticator {
 		var authState = await oktaApi.AuthenticateOktaAsync( new AuthenticateOptions( user, password ) )
 			.ConfigureAwait( false );
 
-		string sessionToken = "";
+		string? sessionToken = null;
 
 		if( authState.OktaStateToken is not null ) {
 
@@ -67,9 +66,10 @@ internal class Authenticator {
 
 		if( !string.IsNullOrEmpty( sessionToken ) ) {
 			var sessionResp = await oktaApi.CreateSessionOktaAsync( new SessionOptions( sessionToken ) );
+			// TODO: Consider making OktaAPI stateless as well (?)
 			oktaApi.AddSession( sessionResp.Id );
 			CacheOktaSession( user, org, sessionResp.Id, sessionResp.ExpiresAt );
-			return new( true, sessionResp.Id );
+			return new( SuccessfulAuthentication: true, OktaSessionId: sessionResp.Id );
 		}
 
 		throw new BmxException( "Authentication Failed" );
@@ -82,12 +82,15 @@ internal class Authenticator {
 	) {
 		var sessionId = GetCachedOktaSession( user, org );
 		if( string.IsNullOrEmpty( sessionId ) ) {
-			return new( false, "" );
+			return new( SuccessfulAuthentication: false, OktaSessionId: "" );
 		}
 
 		oktaApi.AddSession( sessionId );
 		var userId = await oktaApi.GetMeResponseAsync( sessionId );
-		return new( !string.IsNullOrEmpty( userId ), sessionId );
+		if( !string.IsNullOrEmpty( userId ) ) {
+			return new( SuccessfulAuthentication: true, OktaSessionId: userId );
+		}
+		return new( SuccessfulAuthentication: false, OktaSessionId: "" );
 	}
 
 	private static void CacheOktaSession( string userId, string org, string sessionId, DateTimeOffset expiresAt ) {
--- a/src/D2L.Bmx/Authenticate.cs
+++ b/src/D2L.Bmx/Authenticate.cs
@@ -1,3 +1,4 @@
+using System.Text.Json;
 using D2L.Bmx.Okta;
 using D2L.Bmx.Okta.Models;
 using D2L.Bmx.Okta.State;
@@ -10,7 +11,13 @@ internal class Authenticator {
 		string user,
 		bool nomask,
 		IOktaApi oktaApi ) {
-		// TODO: Add authenticating from cache
+
+		oktaApi.SetOrganization( org );
+
+		var cachedSession = await AuthenticateFromCacheAsync( org, user, oktaApi );
+		if( cachedSession.SuccessfulAuthentication ) {
+			return cachedSession;
+		}
 
 		Console.Write( "Okta Password: " );
 		var password = "";
@@ -28,12 +35,10 @@ internal class Authenticator {
 			}
 		}
 
-		oktaApi.SetOrganization( org );
-
 		var authState = await oktaApi.AuthenticateOktaAsync( new AuthenticateOptions( user, password ) )
 			.ConfigureAwait( false );
 
-		OktaAuthenticatedState? authenticatedState = default;
+		string sessionToken = "";
 
 		if( authState.OktaStateToken is not null ) {
 
@@ -55,17 +60,60 @@ internal class Authenticator {
 				PromptMfaInput( selectedMfa.Name );
 				throw new NotImplementedException();
 			}
-			authenticatedState = await oktaApi.AuthenticateChallengeMfaOktaAsync( authState, selectedMfaIndex - 1, mfaInput );
+			sessionToken = await oktaApi.AuthenticateChallengeMfaOktaAsync( authState, selectedMfaIndex - 1, mfaInput );
 		} else if( authState.OktaSessionToken is not null ) {
-			authenticatedState = new OktaAuthenticatedState( true, authState.OktaSessionToken );
+			sessionToken = authState.OktaSessionToken;
 		}
 
-		if( authenticatedState is not null ) {
-			return authenticatedState;
+		if( !sessionToken.Equals( "" ) ) {
+			var sessionResp = await oktaApi.CreateSessionOktaAsync( new SessionOptions( sessionToken ) );
+			oktaApi.AddSession( sessionResp.Id );
+			CacheOktaSession( user, org, sessionResp.Id, sessionResp.ExpiresAt );
+			return new( true, sessionResp.Id );
 		}
 
 		throw new BmxException( "Authentication Failed" );
 	}
+
+	public static async Task<OktaAuthenticatedState> AuthenticateFromCacheAsync(
+		string org,
+		string user,
+		IOktaApi oktaApi
+	) {
+		var sessionId = GetCachedOktaSession( user, org );
+		if( sessionId.Equals( "" ) ) {
+			return new( false, "" );
+		}
+
+		oktaApi.AddSession( sessionId );
+		var userId = await oktaApi.GetMeResponseAsync( sessionId );
+		return new( !userId.Equals( "" ), sessionId );
+	}
+
+	private static void CacheOktaSession( string userId, string org, string sessionId, DateTimeOffset expiresAt ) {
+		var session = new OktaSessionCache( userId, org, sessionId, expiresAt );
+		var existingSessions = ReadOktaSessionCacheFile();
+		existingSessions.Add( session );
+
+		OktaSessionStorage.SaveSessions( existingSessions );
+	}
+
+	private static string GetCachedOktaSession( string userId, string org ) {
+		var oktaSessions = ReadOktaSessionCacheFile();
+		var session = oktaSessions.Find( session => session.UserId == userId && session.Org == org );
+		return session?.SessionId ?? "";
+	}
+
+	private static List<OktaSessionCache> ReadOktaSessionCacheFile() {
+		var sessions = OktaSessionStorage.Sessions();
+		return RemoveExpiredOktaSessions( sessions );
+	}
+
+	private static List<OktaSessionCache> RemoveExpiredOktaSessions( List<OktaSessionCache> sourceCache ) {
+		var currTime = DateTimeOffset.Now;
+		return sourceCache.Where( session => session.ExpiresAt > currTime ).ToList();
+	}
+
 
 	private static int PromptMfa( MfaOption[] mfaOptions ) {
 
--- a/src/D2L.Bmx/Authenticate.cs
+++ b/src/D2L.Bmx/Authenticate.cs
@@ -65,7 +65,7 @@ internal class Authenticator {
 			sessionToken = authState.OktaSessionToken;
 		}
 
-		if( !sessionToken.Equals( "" ) ) {
+		if( !string.IsNullOrEmpty( sessionToken ) ) {
 			var sessionResp = await oktaApi.CreateSessionOktaAsync( new SessionOptions( sessionToken ) );
 			oktaApi.AddSession( sessionResp.Id );
 			CacheOktaSession( user, org, sessionResp.Id, sessionResp.ExpiresAt );
@@ -81,13 +81,13 @@ internal class Authenticator {
 		IOktaApi oktaApi
 	) {
 		var sessionId = GetCachedOktaSession( user, org );
-		if( sessionId.Equals( "" ) ) {
+		if( string.IsNullOrEmpty( sessionId ) ) {
 			return new( false, "" );
 		}
 
 		oktaApi.AddSession( sessionId );
 		var userId = await oktaApi.GetMeResponseAsync( sessionId );
-		return new( !userId.Equals( "" ), sessionId );
+		return new( !string.IsNullOrEmpty( userId ), sessionId );
 	}
 
 	private static void CacheOktaSession( string userId, string org, string sessionId, DateTimeOffset expiresAt ) {
--- a/src/D2L.Bmx/Authenticate.cs
+++ b/src/D2L.Bmx/Authenticate.cs
@@ -104,7 +104,7 @@ internal class Authenticator {
 	private static string? GetCachedOktaSessionId( string userId, string org ) {
 		var oktaSessions = ReadOktaSessionCacheFile();
 		var session = oktaSessions.Find( session => session.UserId == userId && session.Org == org );
-		return session?.SessionId ?? null;
+		return session?.SessionId;
 	}
 
 	private static List<OktaSessionCache> ReadOktaSessionCacheFile() {
--- a/src/D2L.Bmx/Authenticate.cs
+++ b/src/D2L.Bmx/Authenticate.cs
@@ -105,15 +105,10 @@ internal class Authenticator {
 	}
 
 	private static List<OktaSessionCache> ReadOktaSessionCacheFile() {
-		var sessions = OktaSessionStorage.Sessions();
-		return RemoveExpiredOktaSessions( sessions );
-	}
-
-	private static List<OktaSessionCache> RemoveExpiredOktaSessions( List<OktaSessionCache> sourceCache ) {
+		var sourceCache = OktaSessionStorage.Sessions();
 		var currTime = DateTimeOffset.Now;
 		return sourceCache.Where( session => session.ExpiresAt > currTime ).ToList();
 	}
-
 
 	private static int PromptMfa( MfaOption[] mfaOptions ) {
 
--- a/src/D2L.Bmx/Authenticate.cs
+++ b/src/D2L.Bmx/Authenticate.cs
@@ -101,7 +101,7 @@ internal class Authenticator {
 		OktaSessionStorage.SaveSessions( existingSessions );
 	}
 
-	private static string GetCachedOktaSession( string userId, string org ) {
+	private static string? GetCachedOktaSessionId( string userId, string org ) {
 		var oktaSessions = ReadOktaSessionCacheFile();
 		var session = oktaSessions.Find( session => session.UserId == userId && session.Org == org );
 		return session?.SessionId ?? "";
--- a/src/D2L.Bmx/Authenticate.cs
+++ b/src/D2L.Bmx/Authenticate.cs
@@ -34,7 +34,7 @@ internal class Authenticator {
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


src/D2L.Bmx/Okta/State/OktaAuthenticatedState.cs
```diff
@@ -2,5 +2,5 @@ namespace D2L.Bmx.Okta.State;
 
 internal record OktaAuthenticatedState(
 	bool SuccessfulAuthentication,
-	string OktaSessionToken
+	string OktaSessionId
```
gord5500, line 5: Think it's better to be passing around a session id instead. This value is currently what gets stored in session file

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	internal static string BmxDir() {
+		return Path.Join( UserHomeDir(), ".bmx3" );
+	}
+
+	internal static string UserHomeDir() {
+
+		string osPlatform = System.Runtime.InteropServices.RuntimeInformation.OSDescription;
+		string pattern = @"(?i)\bwindows\b";
```
gord5500, line 56: The `osPlatform` value gives a bit more text than just windows

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -0,0 +1,67 @@
...
+		}
+		return new();
+	}
+
+	internal static string SessionsFileName() {
+		return Path.Join( UserHomeDir(), ".bmx3", "sessions" );
+	}
+
+	internal static string BmxDir() {
+		return Path.Join( UserHomeDir(), ".bmx3" );
```
gord5500, line 50: named it bmx3 just to test out. This can go back to bmx

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -0,0 +1,67 @@
...
+		var bmxDirectory = BmxDir();
+		var sessionsFileName = SessionsFileName();
+
+		if( !Directory.Exists( bmxDirectory ) ) {
+			Directory.CreateDirectory( bmxDirectory );
+		}
+
+		if( !File.Exists( sessionsFileName ) ) {
+			File.WriteAllText( sessionsFileName, "[]" );
+		}
```
gord5500, line 31: We talked about this in pair programming. Wanting to create the sessions file in case it wasn't created. Did the directory as well to be safe?

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -113,12 +121,10 @@ internal class OktaApi : IOktaApi {
 	}
 
 	async Task<OktaAccountState> IOktaApi.GetAccountsOktaAsync( OktaAuthenticatedState state, string accountType ) {
-		// TODO: Use existing session if it exists in ~/.bmx and isn't expired
-		var sessionResp = await ( this as IOktaApi ).CreateSessionOktaAsync( new SessionOptions( state.OktaSessionToken ) );
 		// TODO: Consider making OktaAPI stateless as well (?)
-		( this as IOktaApi ).AddSession( sessionResp.Id );
+		var resp = await _httpClient.GetAsync(
+			"users/me/appLinks" );
```
gord5500, line 126: `me` works here since we set the session id during authentication

gord5500:
Might have been better of me to do multiple commits and make it a bit easier to read. Sorry. I kind of just kept writing and assumed it would all work :rippinelli:

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -0,0 +1,64 @@
...
+		return new();
+	}
+
+	internal static string SessionsFileName() {
+		return Path.Join( UserHomeDir(), ".bmx3", "sessions" );
+	}
+
+	internal static string BmxDir() {
+		return Path.Join( UserHomeDir(), ".bmx3" );
```
gord5500, line 47:
```suggestion
		return Path.Join( UserHomeDir(), ".bmx" );
```

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -0,0 +1,64 @@
...
+				return sessions.ToList();
+			}
+		} catch( JsonException ) {
+			return new();
+		}
+		return new();
+	}
+
+	internal static string SessionsFileName() {
+		return Path.Join( UserHomeDir(), ".bmx3", "sessions" );
```
gord5500, line 43:
```suggestion
		return Path.Join( UserHomeDir(), ".bmx", "sessions" );
```

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -142,6 +148,26 @@ internal class OktaApi : IOktaApi {
 		throw new BmxException( "Account could not be found" );
 	}
 
+	async Task<string> IOktaApi.GetMeResponseAsync( string sessionId ) {
+
+		try {
+			using var meRequest = new HttpRequestMessage( HttpMethod.Get, "users/me" );
+			using var meResponse = await _httpClient.SendAsync( meRequest );
```
gord5500, line 155:
```suggestion
			using var meResponse = await _httpClient.GetAsync( "users/me" );
```

Had it like that because I wanted to look at some things in the request object. This fine with me too since we do it that way above

src/D2L.Bmx/Authenticate.cs
```diff
@@ -10,7 +11,13 @@ internal class Authenticator {
 		string user,
 		bool nomask,
 		IOktaApi oktaApi ) {
-		// TODO: Add authenticating from cache
+
+		oktaApi.SetOrganization( org );
+
+		var cachedSession = await AuthenticateFromCacheAsync( org, user, oktaApi );
```
gord5500, line 17:
```suggestion
		OktaAuthenticatedState cachedSession = await AuthenticateFromCacheAsync( org, user, oktaApi );
```

Maybe that or I can rename to `cachedState`

boarnoah: for non built in types, especially ones with long names like this, var is better. Its just nice to use type names for primitives like ints/bools to improve readability.

src/D2L.Bmx/Okta/OktaSessionStorage.cs
boarnoah: Looks necessary for if the directory doesn't exist, WriteAllText throws an exception otherwise.

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	internal static string BmxDir() {
+		return Path.Join( UserHomeDir(), ".bmx3" );
+	}
+
+	internal static string UserHomeDir() {
```
boarnoah, line 50: Lets consider using SpecfialFolder's UserProfile instead https://learn.microsoft.com/en-us/dotnet/api/system.environment.specialfolder  

A glance through the dotnet source suggests they already put in the work to correctly get the home folder on linux, but it wouldn't doing a quick test, its not very readable code to me 😆

gord5500: Oh yea nice. I didn't know about that. Let's use that instead

boarnoah: No concerns from me aside from it being a departure from current BMX which we wanted to avoid at least in this initial release.

> Oh yea nice. I didn't know about that. Let's use that instead

Best to check on a common Linux distro to make sure it finds /home/<user> properly, the code I see here is a little :suspect: https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Environment.SpecialFolder.cs,7d2f1469d916fc63, not quite sure how to interpret its ability to identify home folder correctly https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Environment.GetFolderPathCore.Unix.cs,7d2f1469d916fc63

gord5500: Yea I am not strongly for appdata. I'll check in docker where that special folder goes in linux

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -1,23 +1,27 @@
...
 using System.Xml;
+using Amazon.Runtime.Internal.Endpoints.StandardLibrary;
 using D2L.Bmx.Okta.Models;
 using D2L.Bmx.Okta.State;
 namespace D2L.Bmx.Okta;
 
 internal interface IOktaApi {
 	void SetOrganization( string organization );
 	void AddSession( string sessionId );
+	Uri GetOrganizationUri();
```
gord5500, line 16:
```suggestion
```

Don't need this anymore

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -36,6 +40,10 @@ internal class OktaApi : IOktaApi {
 		_httpClient.BaseAddress = new Uri( $"https://{organization}.okta.com/api/v1/" );
 	}
 
+	Uri IOktaApi.GetOrganizationUri() {
+		return _httpClient.BaseAddress!;
+	}
```
gord5500, line 45:
```suggestion
```

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -0,0 +1,54 @@
...
+			if( sessions is not null ) {
+				return sessions.ToList();
+			}
+		} catch( JsonException ) {
+			return new();
+		}
+		return new();
+	}
+
+	internal static string SessionsFileName() {
```
boarnoah, line 42: For consistency's sake could you refactor all uses of .bmx directory ex: https://github.com/Brightspace/bmx/blob/4d2e39358594fca24b30f9af28f3a068c408e9f1/src/D2L.Bmx/BmxConfigProvider.cs#L14

To either 1) All pull from one place that defines the folder paths or 2) do it inline in the BmxConfigProvider

src/D2L.Bmx/Okta/OktaApi.cs
boarnoah: Looks fine eliminating that variable

src/D2L.Bmx/Okta/OktaSessionStorage.cs
gord5500: Seems like the only other use is the one you mentioned

src/D2L.Bmx/Authenticate.cs
```diff
@@ -55,18 +60,61 @@ internal class Authenticator {
...
-			authenticatedState = await oktaApi.AuthenticateChallengeMfaOktaAsync( authState, selectedMfaIndex - 1, mfaInput );
+			sessionToken = await oktaApi.AuthenticateChallengeMfaOktaAsync( authState, selectedMfaIndex - 1, mfaInput );
 		} else if( authState.OktaSessionToken is not null ) {
-			authenticatedState = new OktaAuthenticatedState( true, authState.OktaSessionToken );
+			sessionToken = authState.OktaSessionToken;
 		}
 
-		if( authenticatedState is not null ) {
-			return authenticatedState;
+		if( !sessionToken.Equals( "" ) ) {
```
boarnoah, line 68: nit: prefer ``string.IsNullOrEmpty`` for these kind of checks to see if a string is set

src/D2L.Bmx/Authenticate.cs
```diff
@@ -55,18 +60,61 @@ internal class Authenticator {
...
+		var session = oktaSessions.Find( session => session.UserId == userId && session.Org == org );
+		return session?.SessionId ?? "";
+	}
+
+	private static List<OktaSessionCache> ReadOktaSessionCacheFile() {
+		var sessions = OktaSessionStorage.Sessions();
+		return RemoveExpiredOktaSessions( sessions );
+	}
+
+	private static List<OktaSessionCache> RemoveExpiredOktaSessions( List<OktaSessionCache> sourceCache ) {
```
boarnoah, line 112: Isn't this more about getting non expired sessions?

src/D2L.Bmx/Authenticate.cs
gord5500: I don't think we'll have a scenario where it's null but I've swapped to that and for the other times it's used

boarnoah: Yep, its mostly for consistency with how we write empty string checks with the other projects in the team

src/D2L.Bmx/Okta/OktaSessionStorage.cs
cfbao: I'd strongly recommend using SpecialFolder.

Not 100% on AppData, but I'd argue that if we switch, this initial release is exactly when we'd want to do it.
This will be a big release in any case and comes with a major version bump, so people will expect incompatible changes. We'd want to avoid breaking changes in the future (after the initial .NET release) without a good reason.

src/D2L.Bmx/Authenticate.cs
```diff
@@ -55,18 +60,56 @@ internal class Authenticator {
...
+			sessionToken = authState.OktaSessionToken;
 		}
 
-		if( authenticatedState is not null ) {
-			return authenticatedState;
+		if( !string.IsNullOrEmpty( sessionToken ) ) {
+			var sessionResp = await oktaApi.CreateSessionOktaAsync( new SessionOptions( sessionToken ) );
+			oktaApi.AddSession( sessionResp.Id );
+			CacheOktaSession( user, org, sessionResp.Id, sessionResp.ExpiresAt );
+			return new( true, sessionResp.Id );
```
cfbao, line 72: could be nice to name the argument here and other calls to this constructor.
literal values are almost always good candidates for named arguments

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -113,12 +116,10 @@ internal class OktaApi : IOktaApi {
 	}
 
 	async Task<OktaAccountState> IOktaApi.GetAccountsOktaAsync( OktaAuthenticatedState state, string accountType ) {
-		// TODO: Use existing session if it exists in ~/.bmx and isn't expired
-		var sessionResp = await ( this as IOktaApi ).CreateSessionOktaAsync( new SessionOptions( state.OktaSessionToken ) );
 		// TODO: Consider making OktaAPI stateless as well (?)
```
cfbao, line 119: ha! this "todo" echos my comments above.
but also this comment should go with the `AddSession` call. now that the call is moved, this comment no longer makes sense here.

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -142,6 +143,25 @@ internal class OktaApi : IOktaApi {
...
+
+		try {
+			using var meResponse = await _httpClient.GetAsync( "users/me" );
+
+			if( !meResponse.IsSuccessStatusCode ) {
+				return "";
+			}
+
+			var meJson = await meResponse.Content.ReadAsStringAsync();
+			var me = JsonSerializer.Deserialize( meJson, SourceGenerationContext.Default.OktaMeResponse )!;
```
cfbao, line 156: `ReadFromJsonAsync` exists:
https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.httpcontentjsonextensions.readfromjsonasync?view=net-7.0

cfbao: also, not a big concern, but it's theoretically possible for the deserialization to return null. so not the ideal to suppress the null warning.

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -142,6 +143,25 @@ internal class OktaApi : IOktaApi {
 		throw new BmxException( "Account could not be found" );
 	}
 
+	async Task<string> IOktaApi.GetMeResponseAsync( string sessionId ) {
+
+		try {
+			using var meResponse = await _httpClient.GetAsync( "users/me" );
+
+			if( !meResponse.IsSuccessStatusCode ) {
+				return "";
```
cfbao, line 152: similarly here (and below), I'd return null rather than an empty string.
the signature of the method could be
```csharp
Task<string?> IOktaApi.GetMeUserIdOrNullAsync( string sessionId )
```
which better illustrate the behaviours of this method.

cfbao: When you return empty strings, the method signature doesn't communicate that behaviour to the caller. The caller can't know that they should check for empty strings unless they read the docs or code of your method.

When you return nulls, the caller would be forced to check for nulls, so this is less error prone and a better dev experience.

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -0,0 +1,54 @@
...
+using D2L.Bmx.Okta.Models;
+
+namespace D2L.Bmx.Okta;
+
+internal class OktaSessionStorage {
+
+	internal static void SaveSessions( List<OktaSessionCache> sessions ) {
+
+		string jsonString = JsonSerializer.Serialize(
+			sessions.ToArray(),
```
cfbao, line 12: do we have to convert to array? list doesn't work with source generator?

gord5500: Looks like it does work 😄 . I originally was just passing around array but realized list would be easier and then forgot to try changing the serializable type

src/D2L.Bmx/Authenticate.cs
```diff
@@ -55,18 +59,60 @@ internal class Authenticator {
...
+
+	private static void CacheOktaSession( string userId, string org, string sessionId, DateTimeOffset expiresAt ) {
+		var session = new OktaSessionCache( userId, org, sessionId, expiresAt );
+		var existingSessions = ReadOktaSessionCacheFile();
+		existingSessions.Add( session );
+
+		OktaSessionStorage.SaveSessions( existingSessions );
+	}
+
+	private static string GetCachedOktaSession( string userId, string org ) {
```
cfbao, line 104: Looking at a name like "get session", I'd expect it to return a session object, but this isn't the case here.
Better be explicit here. There's also the session token that can add to the confusion.
```suggestion
	private static string? GetCachedOktaSessionId( string userId, string org ) {
```

src/D2L.Bmx/Authenticate.cs
```diff
@@ -55,18 +59,60 @@ internal class Authenticator {
...
+		var oktaSessions = ReadOktaSessionCacheFile();
+		var session = oktaSessions.Find( session => session.UserId == userId && session.Org == org );
+		return session?.SessionId ?? "";
```
cfbao, line 107: here also I think returning null would be better

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -142,6 +143,24 @@ internal class OktaApi : IOktaApi {
...
+
+		try {
+			using var meResponse = await _httpClient.GetAsync( "users/me" );
+			if( !meResponse.IsSuccessStatusCode ) {
+				return null;
+			}
+
+			var meJson = await meResponse.Content.ReadAsStringAsync();
+			var me = await meResponse.Content.ReadFromJsonAsync( SourceGenerationContext.Default.OktaMeResponse );
+			return me?.Id ?? null;
```
cfbao, line 156: this should be sufficient
```suggestion
			return me?.Id;
```

cfbao (dismissed):
I still have a few more naming nits, but many of these were really introduced before this PR.

I'm up for a pair programming session to go over some naming & API design issues, and give bmx a tune up. Later of course.

gord5500:
Just had to do some renaming since the main branch got rid of `Okta` in most of the function names.

src/D2L.Bmx/Authenticate.cs
```diff
@@ -55,18 +60,56 @@ internal class Authenticator {
...
 		} else if( authState.OktaSessionToken is not null ) {
-			authenticatedState = new OktaAuthenticatedState( true, authState.OktaSessionToken );
+			sessionToken = authState.OktaSessionToken;
 		}
 
-		if( authenticatedState is not null ) {
-			return authenticatedState;
+		if( !string.IsNullOrEmpty( sessionToken ) ) {
+			var sessionResp = await oktaApi.CreateSessionOktaAsync( new SessionOptions( sessionToken ) );
+			oktaApi.AddSession( sessionResp.Id );
```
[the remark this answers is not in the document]
gord5500: By stateless does this mean we'd want to hold on to the session id and organization somewhere else and then just pass those in when needed?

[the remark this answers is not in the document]
gord5500: Yea for sure, just wanted to make sure I understood. We can talk about it then. Thanks Chenfeng

src/D2L.Bmx/Authenticate.cs
```diff
@@ -55,18 +59,60 @@ internal class Authenticator {
...
+		var existingSessions = ReadOktaSessionCacheFile();
+		existingSessions.Add( session );
+
+		OktaSessionStorage.SaveSessions( existingSessions );
+	}
+
+	private static string? GetCachedOktaSessionId( string userId, string org ) {
+		var oktaSessions = ReadOktaSessionCacheFile();
+		var session = oktaSessions.Find( session => session.UserId == userId && session.Org == org );
+		return session?.SessionId ?? null;
```
cfbao, line 107: `null ?? null == null`

```suggestion
		return session?.SessionId;
```

cfbao (approved):

src/D2L.Bmx/Authenticate.cs
gord5500: Ahh sorry 😓

PR #260, opened 2023-04-24
CI: Implement release creation

gord5500:
### Why

We want a way to build and publish new versions for non d2lers. On a repository dispatch trigger, this will create a new release, compile BMX for all supported operating systems and upload as assets to the release

### How

The repository dispatch trigger will be invoked from a private brightspace repo. We use the github cli to create a release with the version specified in the payload and then on each OS build, add a step to upload the built project

https://github.com/Brightspace/bmx/releases/tag/untagged-30a277809d529269ae8b example of what a release would look like. I wish the names could all be what they are meant to be. Either just bmx or D2L.Bmx. Think they all need unique names and couldn't find a way around that. Unless we zipped each OS up and changed the name on the zip file? 

I did not invoke this trigger to have it create the release linked. It was done by just setting the release tag to v3.0.0 in code. So will need to test that and confirm it still works the intended way

Related Trello card: https://trello.com/c/k8YEuX06/1226-implement-release-strategy


.github/actions/create-release/action.yml
```diff
@@ -0,0 +1,31 @@
...
+        else
+          NEXT_RELEASE_VERSION=$(echo "${LATEST_RELEASE_VERSION} + 0.0.1" | bc)
+        fi
+
+        RELEASE_TAG="v${NEXT_RELEASE_VERSION}"
+        echo "RELEASE_TAG=${RELEASE_TAG}" >> $GITHUB_ENV
+
+        gh release create "${RELEASE_TAG}" \
+          --title "Release ${RELEASE_TAG}" \
+          --notes "Release notes for version ${RELEASE_TAG}"
```
gord5500, line 28: We can edit the notes after if there's anything important we want to mention

.github/workflows/publish.yml
```diff
@@ -11,34 +11,73 @@ on:
...
+      PLATFORM: ${{ matrix.platform }}
     steps:
       - name: checkout
         uses: Brightspace/third-party-actions@actions/checkout
       - name: publish
         shell: pwsh
+        working-directory: src/D2L.Bmx
+        run: |
+          dotnet publish -r $env:PLATFORM-x64 -c Release -o build
+      - name: upload build
```
gord5500, line 57: I thought to have this step as a reusable action but it just ended up being even more code lol

.github/workflows/publish.yml
```diff
@@ -11,34 +11,73 @@ on:
...
+    if: ${{ github.event_name == 'repository_dispatch' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: Brightspace/third-party-actions@actions/checkout
+      - name: Create release
+        env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+        run: |
+          gh release create "$RELEASE_TAG" --draft `
```
gord5500, line 31: `--draft` because I accidentally broke someone's workflow process by creating non-drafted releases 😓

.github/workflows/publish.yml
```diff
@@ -11,34 +11,73 @@ on:
...
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
           PLATFORM: ${{ matrix.platform }}
+          FILE_NAME: ${{ matrix.file_name }}
+          GH_TOKEN: ${{ github.token }}
         working-directory: src/D2L.Bmx
-        run: dotnet publish -r $env:PLATFORM-x64 -c Release -o out
+        run: |
+          cd build
+          mv "$FILE_NAME" "$PLATFORM-$FILE_NAME"
+          gh release upload "$RELEASE_TAG" "$PLATFORM-$FILE_NAME"
```
gord5500, line 69: There kind of is a worry that the create release step fails and then this tries to upload to a tag that's not created. Should just fail the workflow rather than do anything bad. Wasn't sure how to resolve this because the step has to be in this job since the file is getting built here. And we don't want to use build artifacts to upload / download at a later time

.github/workflows/publish.yml
gord5500: Yea for the jumpboxes they use the latest version release apparently and me creating releases while testing was making them pull the dotnet version

.github/workflows/publish.yml
```diff
@@ -11,34 +11,73 @@ on:
...
 
   build_docker:
+    if: ${{ always() }}
+    needs: create_release
```
gord5500, line 73:
```suggestion
```

This is meant to go but related to [this ](https://github.com/Brightspace/bmx/pull/260/files#r1176898984) comment and what I was doing before

.github/workflows/publish.yml
cfbao: we can make the `build` job depend on the `create_release` job.

in the `if` condition of the build job, specify `!failure()` so the build job would run even when create_release is skipped.
by default, `success()` is assumed, which would be false when a job is skipped.

.github/workflows/publish.yml
```diff
@@ -11,34 +11,73 @@ on:
...
         env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
           PLATFORM: ${{ matrix.platform }}
+          FILE_NAME: ${{ matrix.file_name }}
+          GH_TOKEN: ${{ github.token }}
         working-directory: src/D2L.Bmx
-        run: dotnet publish -r $env:PLATFORM-x64 -c Release -o out
+        run: |
+          cd build
+          mv "$FILE_NAME" "$PLATFORM-$FILE_NAME"
```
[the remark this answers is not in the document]
cfbao: there should be some configs available in csproj that can make the output file just "bmx"

.github/actions/create-release/action.yml
cfbao: `--generate-notes` might be better?

.github/workflows/publish.yml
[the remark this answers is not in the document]
cfbao: zipping is also a good idea. so we'd have `bmx-{os-name}.zip` which contains `bmx[.exe]`

.github/workflows/publish.yml
cfbao: yeah this is not surprising. one should not change the "latest" release of a GitHub repo lightly

.github/workflows/publish.yml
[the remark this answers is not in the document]
cfbao: try `<TargetName>bmx</TargetName>` in csproj

.github/actions/create-release/action.yml
gord5500: Yep I saw this when looking for the draft flag and wanted to add it in too

.github/workflows/publish.yml
```diff
@@ -11,34 +11,78 @@ on:
     paths:
       - "src/**"
       - ".github/workflows/publish.yml"
+  repository_dispatch:
+    types: [create-release]
+    client_payload:
+      release_tag:
+        required: false
+        default: 3.0.0
+  workflow_dispatch:
```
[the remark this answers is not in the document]
gord5500: I could have maybe branched off actually and tried it there

.github/workflows/publish.yml
```diff
@@ -11,34 +11,78 @@ on:
...
       - ".github/workflows/publish.yml"
+  repository_dispatch:
+    types: [create-release]
+    client_payload:
+      release_tag:
+        required: true
 
 jobs:
+  create_release:
+    if: ${{ github.event_name == 'repository_dispatch' }}
```
cfbao, line 22: nit: `${{ ... }}` can be omitted in an `if`

.github/workflows/publish.yml
```diff
@@ -11,34 +11,78 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: Brightspace/third-party-actions@actions/checkout
```
cfbao, line 26: do we even need to check out the repo in this job?

.github/workflows/publish.yml
```diff
@@ -11,34 +11,78 @@ on:
...
     timeout-minutes: 10
+    env:
+      PLATFORM: ${{ matrix.platform }}
     steps:
       - name: checkout
         uses: Brightspace/third-party-actions@actions/checkout
+      - name: install 7zip
+        if: matrix.platform == 'win'
+        run: choco install 7zip -y
+        shell: bash
```
cfbao, line 56: powershell should meet our zipping needs in all OSes
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.archive/compress-archive?view=powershell-7.3

.github/workflows/publish.yml
```diff
@@ -11,34 +11,78 @@ on:
...
         env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
           PLATFORM: ${{ matrix.platform }}
+          FILE_NAME: ${{ matrix.file_name }}
+          GH_TOKEN: ${{ github.token }}
         working-directory: src/D2L.Bmx
-        run: dotnet publish -r $env:PLATFORM-x64 -c Release -o out
+        run: |
+          cd build
```
cfbao, line 72: might as well add `/build` to `working-directory` then

.github/workflows/publish.yml
```diff
@@ -11,34 +11,78 @@ on:
...
+          7z a "$PLATFORM-$FILE_NAME.zip" "$FILE_NAME"
+          gh release upload "$RELEASE_TAG" "$PLATFORM-$FILE_NAME.zip"
```
cfbao, line 74: nit: personally I'd prefer `bmx-{platform}.zip` rather than `{platform}-bmx.zip`.

When I download a tool onto my machine, the first thing I'm going to search for in my filesystem is the tool's name, e.g. "bmx". I won't be thinking about my "platform" - all my tools will be for my computer's platform. So putting the tool name "bmx" first would have more distinguishing power, e.g. good for auto complete on in the terminal.

cfbao: also, I don't think we should put ".exe" in the windows artifacts' name. we can just hardcode "bmx" here.

.github/workflows/publish.yml
gord5500: Hmm I don't think so. It might have been needed when I had it as a separate action in a different file

cfbao (dismissed):

.github/workflows/publish.yml
```diff
@@ -51,4 +88,15 @@ jobs:
         env:
           DOCKERFILE: ${{ matrix.file }}
         run: |
-          docker buildx build -f $env:DOCKERFILE -o out .
+          docker buildx build -f $env:DOCKERFILE -o build .
+      - name: upload build
+        if: github.event_name == 'repository_dispatch'
+        shell: bash
```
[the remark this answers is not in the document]
gord5500: Oh yea missed this. Thanks Chenfeng

cfbao (approved):

PR #265, opened 2023-05-02
Set partial trimming

gord5500:
### Why

The published build of bmx right now will crash with the error `No parameterless constructor defined for type 'Amazon.Util.Internal.PlatformServices.NetworkReachability'`

https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-self-contained

When building bmx with aot, we reduce the deployment size by getting rid of unused parts of libraries. Theres a [problem ](https://github.com/aws/aws-sdk-net/issues/2531) though with the aws sdk right now where it gets rid of some parts we actually need. I think it's to do with [this](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/incompatibilities#dynamic-assembly-loading-and-execution)

### How

The github issue page mentioned a temporary solution to set the trim mode to partial for now. The build size would be 17.5MB compared to 13 so it seems like not a big issue?


src/D2L.Bmx/D2L.Bmx.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
cfbao, line 1: You're adding a BOM here. Did you use VS to edit this file? Can you set the encoding to be without BOM?

cfbao: our editorconfig does set utf-8 (which means without BOM), but VS just doesn't respect that... 😠

cfbao: the linter we use is unfortunately not able to catch charset violations
https://github.com/editorconfig-checker/editorconfig-checker/issues/85

cfbao:
I believe there's a way to scope what we trim/don't trim to specific assemblies via a config file. It might be able to reduce binary size a bit without touching the "fragile" assemblies (e.g. AWS ones).
Not a priority as 17.5 MB isn't a big deal.

Issue #204, opened 2020-07-16
Support output format expected by aws credentials_process


dougmoscrop:
Something like:

> bmx login

Would prompt you for:

- Org (default read from bmx confing)
- Username (default read from bmx config)
- Password

And just setup/refresh the Okta session.

Then in `~/.aws/credentials`:

```ini
[dev-my-account]
credential_process = bmx aws-credentials --account=Dev-MyAccount --role=Dev-MyAccount-Owner
```

No opinion on the actual command, could also be `bmx print --aws` or `bmx print --json` and so on.

I think this has the same "failure mode" as when the `bmx write` credentials are expired (credentials process would just return the expired credentials if it can't renew them) but it avoids having to type in profile, account and role names as often.

boarnoah:
Implemented via: #443

The workflow is:

Setup credential process with a role (once):
     ``bmx write --use-credential-process --account Dev-MyAccount --role Dev-MyAccount-Owner --profile dev-my-account``
 
Then in the future you need to do ``bmx login`` in order to refresh Okta credentials.

There is caching for the generated AWS credentials #384 
- With ``--duration`` set for > 20 mins (default is 60), credentials will be refreshed when < 15 mins to expiry
- With shorter duration than that it operates on a sliding window, at a minimum when < 10 mins to expiry

PR #267, opened 2023-05-05
print: support json / aws credentials output

gord5500:
### Why
In support of this request from a forked version of bmx https://github.com/Brightspace/bmx/issues/204
This allows us to add entries to our ~.aws/credentials files where we can call bmx to retrieve credentials for us. 

### How
I'm adding an extra option value for --output flag as json but also want a new flag --headless .

My thought was maybe someone could want json output outside of the automated process? With the headless flag on it would just throw an error anytime a needed value wasn't provided. Although it can't actually be a headless process because if the okta session is expired we can't get credentials. Right now with the headless flag and we don't have credentials a non-zero exit code is returned and aws knows bmx failed

Do we want to have it split like this or just anytime the output is json (or name it aws or something else) we don't bother looking for inputs. Right now if you try to use an aws command without a session and without throwing an error you don't even get prompted to enter the okta password it just stalls for awhile
This is what an entry in the credentials file looks like
> [bmxtest]
credential_process = bmx print --headless --output json --account Dev-Slims --role Dev-Slims-Readonlythey should add the following to their credentials file

Related trello card: https://trello.com/c/nOA2khxF/1230-support-awscredentials


cfbao:
> My thought was maybe someone could want json output outside of the automated process? With the headless flag on it would just throw an error anytime a needed value wasn't provided. Although it can't actually be a headless process because if the okta session is expired we can't get credentials. Right now with the headless flag and we don't have credentials a non-zero exit code is returned and aws knows bmx failed

I like the idea of the flag, but I'd like to propose a different name: `--non-interactive`. I think it's easier to understand than "headless". It's longer but you will never type it in an interactive console.

Also, have you confirmed that AWS knows the process failed when a non-zero exit code is returned? How does, say the AWS CLI, behave when this happens? When you invoke the AWS CLI repeatedly with such a profile, are the credentials cached? i.e. is bmx invoked repeatedly every time or only when it's needed to refresh the creds?

gord5500:
> I like the idea of the flag, but I'd like to propose a different name: --non-interactive. I think it's easier to understand than "headless". It's longer but you will never type it in an interactive console.

Yep I think that works good.

> Also, have you confirmed that AWS knows the process failed when a non-zero exit code is returned? How does, say the AWS CLI, behave when this happens?

Yep when there is no valid session we throw a BmxException which gives us non-zero and this is what it looks like in the cli 
image

> When you invoke the AWS CLI repeatedly with such a profile, are the credentials cached? i.e. is bmx invoked repeatedly every time or only when it's needed to refresh the creds?

The aws [docs ](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html) mentions that they don't get cached so bmx should get invoked repeatedly. If I understand correctly.

>The AWS CLI does not cache external process credentials the way it does assume-role credentials. If caching is required, you must implement it in the external process.

cfbao:
> The aws [docs ](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html) mentions that they don't get cached so bmx should get invoked repeatedly. If I understand correctly.
> 
> > The AWS CLI does not cache external process credentials the way it does assume-role credentials. If caching is required, you must implement it in the external process.

😬 I have my concerns then... Imagine this being used in a script that makes a bunch of AWS API calls... Okta API will probably not like this.
I think we should implement caching in bmx for this, potentially controlled with a new flag. But that can be done in a different PR or a new Trello card entirely.

cfbao (approved):

PR #271, opened 2023-05-05
ci: publish workflow uses tar for mac / linux

gord5500:
### Why

A bit more native than zip files for these OS's


.github/workflows/publish.yml
gord5500, line 1: Wasn't 100% sure if doing it like this is okay or if we'd rather two separate steps that only run if the machine is mac / windows

gord5500: Maybe we should get rid of the $env:PLATFORM and just do that manually now?

.github/workflows/publish.yml
```diff
@@ -63,8 +63,14 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         working-directory: src/D2L.Bmx/build
         run: |
-          Compress-Archive -Path "$env:FILE_NAME" -DestinationPath "bmx-$env:PLATFORM.zip"
-          gh release upload "$env:RELEASE_TAG" "bmx-$env:PLATFORM.zip"
+          if ($env:PLATFORM -eq 'win') {
+            Compress-Archive -Path "$env:FILE_NAME" -DestinationPath "bmx-$env:PLATFORM.zip"
+            gh release upload "$env:RELEASE_TAG" "bmx-$env:PLATFORM.zip"
```
gord5500, line 68:
```suggestion
            Compress-Archive -Path "$env:FILE_NAME" -DestinationPath "bmx-win.zip"
            gh release upload "$env:RELEASE_TAG" "bmx-win.zip"
```

Might as well just do this then if we are okay with this if statement here?

gord5500: and the equivalent for mac

cfbao: might as well

.github/workflows/publish.yml
cfbao: not the prettiest but this is ok

.github/workflows/publish.yml
```diff
@@ -63,8 +63,14 @@ jobs:
...
+          } else if ($env:PLATFORM -eq 'osx') {
+            chmod u+x "bmx"
+            tar -cvf "bmx-$env:PLATFORM.tar" "$env:FILE_NAME"
```
cfbao, line 71: might as well compress the file too - add the `-z` flag

cfbao: also customary to use the `.tar.gz` extension when compressing

.github/workflows/publish.yml
```diff
@@ -63,8 +63,14 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         working-directory: src/D2L.Bmx/build
         run: |
-          Compress-Archive -Path "$env:FILE_NAME" -DestinationPath "bmx-$env:PLATFORM.zip"
-          gh release upload "$env:RELEASE_TAG" "bmx-$env:PLATFORM.zip"
+          if ($env:PLATFORM -eq 'win') {
+            Compress-Archive -Path "$env:FILE_NAME" -DestinationPath "bmx-$env:PLATFORM.zip"
+            gh release upload "$env:RELEASE_TAG" "bmx-$env:PLATFORM.zip"
+          } else if ($env:PLATFORM -eq 'osx') {
+            chmod u+x "bmx"
```
cfbao, line 70: not sure if `u+x` is enough, or if `+x` is better. I vaguely remember there was a case of `u+x` not doing enough for us. I think it was a lambda issue @boarnoah had.

.github/workflows/publish.yml
```diff
@@ -98,5 +104,6 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         working-directory: build
         run: |
-          Compress-Archive -Path "bmx" -DestinationPath "bmx-$env:PLATFORM.zip"
-          gh release upload "$env:RELEASE_TAG" "bmx-$env:PLATFORM.zip"
+          chmod u+x "bmx"
+          tar -cvf "bmx-$env:PLATFORM.tar" "bmx"
+          gh release upload "$env:RELEASE_TAG" "bmx-$env:PLATFORM.tar"
```
cfbao, line 109: when PowerShell isn't needed, there's merit to using bash due to its more reliable fail fast behaviour

.github/workflows/publish.yml
boarnoah: Don't recall, from what I understand re: tar, it preserves the bits for user, group and others. What it can't do is preserve ownership. So if we create a file with rw for user, but at the other end the current user is different there could be issues? 

Not sure I understood that properly. It seems like a safe bet to to ``+x`` in this case.

cfbao: > it preserves the bits for user, group and others. What it can't do is preserve ownership

That's my understanding too.

> So if we create a file with rw for user, but at the other end the current user is different there could be issues?

Not sure how a user switch would behave. I suspect tar would set the user execution flag on the other side appropriately. My concern is that there'd be a need to to have +x on the group and root too. It seems customary for "installed" executables to get +x on all three.

So yeah, agreed that `+x` is what we want.

.github/workflows/publish.yml
```diff
@@ -63,8 +63,14 @@ jobs:
...
         run: |
-          Compress-Archive -Path "$env:FILE_NAME" -DestinationPath "bmx-$env:PLATFORM.zip"
-          gh release upload "$env:RELEASE_TAG" "bmx-$env:PLATFORM.zip"
+          if ($env:PLATFORM -eq 'win') {
+            Compress-Archive -Path "bmx.exe" -DestinationPath "bmx-win.zip"
+            gh release upload "$env:RELEASE_TAG" "bmx-win.zip"
+          } else if ($env:PLATFORM -eq 'osx') {
+            chmod +x "bmx"
+            tar -czvf "bmx-osx.tar.gz" "bmx"
+            gh release upload "$RELEASE_TAG" "bmx-osx.tar.gz"
```
cfbao, line 72: you're still in powershell here
```suggestion
            gh release upload "$env:RELEASE_TAG" "bmx-osx.tar.gz"
```

.github/workflows/publish.yml
```diff
@@ -98,5 +104,6 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         working-directory: build
         run: |
-          Compress-Archive -Path "bmx" -DestinationPath "bmx-$env:PLATFORM.zip"
-          gh release upload "$env:RELEASE_TAG" "bmx-$env:PLATFORM.zip"
+          chmod +x "bmx"
+          tar -czvf "bmx-$PLATFORM.tar.gz" "bmx"
+          gh release upload "$RELEASE_TAG" "bmx-$PLATFORM.tar.gz"
```
cfbao, line 109: need to change `shell: ` to really change the shell
https://github.com/Brightspace/bmx/pull/271/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R100

cfbao (approved):

.github/workflows/publish.yml
cfbao: there's also the option of doing the `if` at the GitHub Actions step level, if you want to use different shells for the two conditions
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

.github/workflows/publish.yml
```diff
@@ -97,7 +97,7 @@ jobs:
           docker buildx build -f $env:DOCKERFILE -o build .
       - name: upload build
         if: github.event_name == 'repository_dispatch'
-        shell: pwsh
+        shell: shell
```
cfbao:
```suggestion
        shell: bash
```

gord5500: I think this already got changed

cfbao (approved):

.github/workflows/publish.yml
gord5500: Haha maybe I was a bit tired

PR #272, opened 2023-05-08
Dont save session if no config

gord5500:
### Why

It's a concern for the jumpboxes where people could have their sessions being saved. We'll guide non-d2l users to run configure command and automatically run it during install script for d2l users


src/D2L.Bmx/Okta/OktaSessionStorage.cs
gord5500, line 1: Thought it would be better to add this blocking here instead of in the other files

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -15,6 +21,10 @@ internal class OktaSessionStorage {
 
 	internal static List<OktaSessionCache> Sessions() {
 
+		if( !ConfigFileExists() ) {
+			return new();
+		}
```
gord5500, line 26: Thinking it would be good to ignore any existing sessions file just in case

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -7,6 +7,12 @@ internal class OktaSessionStorage {
 
 	internal static void SaveSessions( List<OktaSessionCache> sessions ) {
 
+		if( !ConfigFileExists() ) {
+			Console.WriteLine( "No config file found. Your session will not be saved. " +
```
cfbao, line 11: nit - I think it's clearer this way
```suggestion
			Console.WriteLine( "No config file found. Your Okta session will not be cached. " +
```

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -7,6 +7,12 @@ internal class OktaSessionStorage {
 
 	internal static void SaveSessions( List<OktaSessionCache> sessions ) {
 
+		if( !ConfigFileExists() ) {
```
cfbao, line 10: not a fool proof check but I think this is enough 99% of the time

src/D2L.Bmx/Okta/OktaSessionStorage.cs
cfbao: Not 100% sold on this. `OktaSessionStorage` is a pretty low level storage class that normally we'd avoid putting in any "business logic" or UI feature (like `Console.WriteLine( ... )`).

`Authenticator.CacheOktaSession` is the only code calling `SaveSessions` and is closer to the business logic. I think it makes sense to put the blocking logic there.

src/D2L.Bmx/Okta/OktaSessionStorage.cs
gord5500: Probably for the jumpboxes we'll have to remove the config files on them right? I don't know the setup process or if it's like new images spin up every time

src/D2L.Bmx/Okta/OktaSessionStorage.cs
gord5500: What about changing the return type to bool and returning false on file not existing and true on save? Then the authenticator class can tell the user it did not cache.

cfbao: Hmmm... still not my preference, but I can live with it if you also change the method name to `TrySaveSessions`.

By "business logic" I'm also including the logic that decides when we save or don't save a session. Normally I'd expect a low level storage class to just do what they're told without adding more logic themselves. They handle things like serialization and talking to external systems. Any other logic I'd expect that to be left to higher level code.

This clear separation can be useful when it comes time when we're debugging/troubleshooting issues like "why isn't my session cached" or when we need to tweak our "business logic" (including safety rules). Having such logic inside the storage layer can be unexpected when writing new code and may lead to future bugs. It makes it harder to tweak the logic because it's intertwined with storage.

I get the incentive to put safeguards at lower levels to prevent people from forgetting adding similar checks, but I think the `Authenticator` is low enough for this purpose. To me, it makes sense that "Okta session storage" should only be used by "Okta authenticator" and any other code should use "Okta authenticator" to do authentication. And the "authenticator" should know all the rules.

src/D2L.Bmx/Okta/OktaSessionStorage.cs
cfbao: The org-wide jumpboxes are either fresh machines or come with user-specific profiles. So it shouldn't be a problem there.
On our own bastion machines there's no separation of any sort.

Ideally we'd check if the username in the config file exists & matches that of the current bmx run, but I'm ok with not doing that for simplicity. Really people just shouldn't set up bmx config file if they don't own the machine (or at least a profile).

src/D2L.Bmx/Okta/OktaSessionStorage.cs
gord5500: I think I understand the reasoning now and good to move it to authenticator class. It would probably be a good time to move the `UserHomeDir` you [mentioned ](https://github.com/Brightspace/bmx/pull/273#discussion_r1189034603)in the configure pr. Is there a specific way you want me to do that?

cfbao: I don't have a good idea of the overall shape of bmx right now to suggest a specific way, but at a glance it'd just be a generic static class that provides constant values.

src/D2L.Bmx/Utilities.cs
```diff
@@ -0,0 +1,9 @@
+
+namespace D2L.Bmx;
+internal class Utilities {
```
cfbao, line 3: we can do better here... something named "utilities" can often end up as a dumping ground.
maybe `BmxPaths`?

gord5500: Sure. That sounds fine

cfbao (approved):

PR #273, opened 2023-05-08
Add configure command

gord5500:
### Why

We don't want to create config files by default for the small jumpbox use case. Instead of doing this automatically in an install script we add this command so non-d2l users can create a config file without manually writing it themselves. This also gets run after our d2l install script is run


src/D2L.Bmx/Program.cs
```diff
@@ -51,7 +51,31 @@
 
 var rootCommand = new RootCommand();
 
-var printCommand = new Command( "print", "Print the long stuff to screen" )
```
gord5500, line 54: Thought this was funny

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -53,6 +56,26 @@ internal class BmxConfigProvider : IBmxConfigProvider {
 		);
 	}
 
+	public void SaveConfiguration( BmxConfig config ) {
```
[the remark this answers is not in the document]
gord5500: So I can change it back to not use reflection if anyone has an issue with it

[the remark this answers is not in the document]
cfbao: Please no reflection...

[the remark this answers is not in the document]
cfbao: We need to do mapping anyway, e.g. `DefaultDuration` <-> `default_duration`, so the reflection code is technically wrong here.
Reflection isn't trimming or AOT friendly in general, so we want to avoid that especially in bmx unless there's a very compelling reason otherwise.

src/D2L.Bmx/ConfigureHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+	private readonly IBmxConfigProvider _configProvider;
+
+	public ConfigureHandler( IBmxConfigProvider configProvider ) {
+		_configProvider = configProvider;
+	}
+
+	public void Handle(
+		string? org,
+		string? user,
+		int? duration
```
cfbao, line 14: for clarity
```suggestion
		int? defaultDuration
```

src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -32,6 +32,16 @@ internal static class ConsolePrompter {
 		return user;
 	}
 
+	public static int PromptDefaultDuration() {
+		Console.Write( "Default duration of session in minutes (recommended: 60): " );
```
cfbao, line 36: instead of "recommended: 60", I wonder if we should just say "default 60", and allow the user to not enter anything (and we don't write anything to the config file in that case)

src/D2L.Bmx/BmxConfigProvider.cs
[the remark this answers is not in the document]
gord5500: Haha okay. I will just use WriteLine for each one. I didn't know about the trimming worries and just saw when publishing to aot it was working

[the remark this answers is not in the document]
gord5500: I originally thought to save it as json thinking that'd be easiest but then we'd break existing setups

cfbao:
`allow_project_configs` is also set in the user-level config and often used by people.
we should support setting it in this "configure" command.

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -53,6 +54,28 @@ internal class BmxConfigProvider : IBmxConfigProvider {
...
+		using StreamWriter writer = new StreamWriter( ConfigFileName(), false );
```
cfbao, line 58: avoid redundant type name, and please name the literal argument `false`.
```suggestion
		using var writer = new StreamWriter( ConfigFileName(), false );
```

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -53,6 +54,28 @@ internal class BmxConfigProvider : IBmxConfigProvider {
...
+		}
+		if( !string.IsNullOrEmpty( config.User ) ) {
+			writer.WriteLine( $"user={config.User}" );
+		}
+		if( config.DefaultDuration.HasValue ) {
+			writer.WriteLine( $"default_duration={config.DefaultDuration}" );
+		}
+	}
+
+	internal static string ConfigFileName() {
```
cfbao, line 71: avoid making the scope bigger than necessary.
```suggestion
	private static string ConfigFileName() {
```

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -53,6 +54,28 @@ internal class BmxConfigProvider : IBmxConfigProvider {
...
+		return Path.Join( UserHomeDir(), ".bmx", "config" );
+	}
+
+	internal static string UserHomeDir() {
```
cfbao, line 75:
```suggestion
	private static string UserHomeDir() {
```

cfbao: I don't think this is the only place where we have this value? we might consider consolidating them.
can do this later.

cfbao (dismissed):

gord5500:
> `allow_project_configs` is also set in the user-level config and often used by people. we should support setting it in this "configure" command.
> 
> can be in a follow up PR

I did see this in my own config file but not in the config record so wasn't sure if we weren't planning to use it. Do we want to add it to `BmxConfig` or should I just change the parameters to pass to include that option. I feel like it would make sense to `BmxConfig` record to have everything that would exist in a config file?

cfbao (approved):

cfbao:
> > `allow_project_configs` is also set in the user-level config and often used by people. we should support setting it in this "configure" command.
> > can be in a follow up PR
> 
> I did see this in my own config file but not in the config record so wasn't sure if we weren't planning to use it. Do we want to add it to `BmxConfig` or should I just change the parameters to pass to include that option. I feel like it would make sense to `BmxConfig` record to have everything that would exist in a config file?

I'm fine with either way. This can be done in a separate PR too.

PR #287, opened 2023-05-18
ci: publish workflow uses draft tag id and finalizes release

gord5500:
### Why

Creating a drafted release doesn't set the tag as what our input is. It's of some format untagged-xxxxxxxxxx . So we need to extract that value from when we create the release and then pass that to remaining steps when uploading.

Also need a new step for removing the draft flag once all the builds have been uploaded.
We're adding a new input for the dispatch called `prerelease`. This will set the created release as a prerelease so we can test the entire workflow without updating the `latest` release tag

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -122,7 +122,7 @@ jobs:
   publish_release:
     needs: [create_release, build, build_docker]
     if: |
-      ${{ !failure() && github.event_name == 'workflow_dispatch' && github.event.inputs.dry_run == 'false' }}
+      ${{ !failure() && github.event_name == 'workflow_dispatch' && github.event.inputs.dry_run == false }}
     runs-on: ubuntu-latest
     timeout-minutes: 5
     env:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,23 +11,28 @@ on:
     paths:
       - "src/**"
       - ".github/workflows/publish.yml"
-  repository_dispatch:
-    types: [create-release]
-    client_payload:
+  workflow_dispatch:
+    inputs:
       release_tag:
         required: true
+        description: 'Tag for the new release'
 
 jobs:
   create_release:
-    if: github.event_name == 'repository_dispatch'
+    if: github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
+    outputs:
+      release_id: ${{ steps.create.outputs.release_id }}
     steps:
       - name: Create release
         env:
           RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
           GH_TOKEN: ${{ github.token }}
+        shell: pwsh
         run: |
-          gh release create "$RELEASE_TAG" --draft --generate-notes --title "Release $RELEASE_TAG"
+          $releaseUrl = gh release create "$RELEASE_TAG" --draft --generate-notes --title "Release $RELEASE_TAG"
+          $releaseId = $releaseUrl -split '/' | Select-Object -Last 1
+          Write-Output "::set-output name=release_id::$releaseId"
 
   build:
     needs: create_release
@@ -54,10 +59,10 @@ jobs:
         run: |
           dotnet publish -r $env:PLATFORM-x64 -c Release -o build
       - name: upload build
-        if: github.event_name == 'repository_dispatch'
+        if: github.event_name == 'workflow_dispatch'
         shell: pwsh
         env:
-          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+          RELEASE_TAG: ${{ needs.create_release.outputs.release_id }}
           PLATFORM: ${{ matrix.platform }}
           FILE_NAME: ${{ matrix.file_name }}
           GH_TOKEN: ${{ github.token }}
@@ -96,10 +101,10 @@ jobs:
         run: |
           docker buildx build -f $env:DOCKERFILE -o build .
       - name: upload build
-        if: github.event_name == 'repository_dispatch'
+        if: github.event_name == 'workflow_dispatch'
         shell: bash
         env:
-          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+          RELEASE_TAG: ${{ needs.create_release.outputs.release_id }}
           PLATFORM: ${{ matrix.platform }}
           GH_TOKEN: ${{ github.token }}
         working-directory: build
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -31,6 +31,7 @@ jobs:
       release_id: ${{ steps.create.outputs.release_id }}
     steps:
       - name: Create release
+        id: create
         env:
           RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
           PRERELEASE: ${{ github.events.input.prerelease }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,6 +16,12 @@ on:
       release_tag:
         required: true
         description: 'Tag for the new release'
+    dry_run:
+      description: 'If true, the draft release will not be finalized'
+      type: boolean
+      required: true
+      default: false
+
 
 jobs:
   create_release:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -11,23 +11,34 @@ on:
...
+      release_id: ${{ steps.create.outputs.release_id }}
     steps:
       - name: Create release
         env:
           RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
           GH_TOKEN: ${{ github.token }}
+        shell: pwsh
         run: |
-          gh release create "$RELEASE_TAG" --draft --generate-notes --title "Release $RELEASE_TAG"
+          $releaseUrl = gh release create "$RELEASE_TAG" --draft --generate-notes --title "Release $RELEASE_TAG"
```
cfbao, line 39: this is PowerShell
```suggestion
          $releaseUrl = gh release create "$env:RELEASE_TAG" --draft --generate-notes --title "Release $RELEASE_TAG"
```

also could be nice to check the exit code
```suggestion
          $releaseUrl = gh release create "$env:RELEASE_TAG" --draft --generate-notes --title "Release $RELEASE_TAG"
          if ( $LASTEXITCODE -ne 0 ) { throw "Failed to create draft release" }
```
PowerShell's failfast mode doesn't automatically check exist code from external executables (unlike bash).

.github/workflows/publish.yml
```diff
@@ -96,14 +107,29 @@ jobs:
...
+      ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.dry_run == false }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    env:
+      RELEASE_TAG: ${{ needs.create_release.outputs.release_id }}
+      GH_TOKEN: ${{ github.token }}
+    steps:
+      - name: finalize release
+        run: |
+          # gh release edit "$env:RELEASE_ID" --draft=false
```
cfbao, line 134: not powershell (I know this is commented out for now)
```suggestion
          # gh release edit "$RELEASE_ID" --draft=false
```

cfbao:
also good to include:
if we're a making a release (not a prerelease or a draft-only release), restrict the workflow/jobs to only the main branch

gord5500:
> 

I like this idea. I'll look into prereleases

.github/workflows/publish.yml
```diff
@@ -11,23 +11,34 @@ on:
...
       - name: Create release
         env:
           RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
           GH_TOKEN: ${{ github.token }}
+        shell: pwsh
         run: |
-          gh release create "$RELEASE_TAG" --draft --generate-notes --title "Release $RELEASE_TAG"
+          $releaseUrl = gh release create "$RELEASE_TAG" --draft --generate-notes --title "Release $RELEASE_TAG"
+          $releaseId = $releaseUrl -split '/' | Select-Object -Last 1
+          Write-Output "::set-output name=release_id::$releaseId"
```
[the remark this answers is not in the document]
gord5500: Huh funny I got that from the github docs. Maybe I was on an outdated page

.github/workflows/publish.yml
gord5500: Oops sorry about those. I think everything has the right syntax now

.github/workflows/publish.yml
```diff
@@ -11,23 +11,38 @@ on:
...
       - name: Create release
         env:
           RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+          PRERELEASE: ${{ github.events.input.prerelease }}
           GH_TOKEN: ${{ github.token }}
+        shell: pwsh
         run: |
-          gh release create "$RELEASE_TAG" --draft --generate-notes --title "Release $RELEASE_TAG"
+          $prereleaseFlag = if ($env:PRERELEASE -eq $true) { '--prerelease' } else { '' }
+          $releaseUrl = gh release create "$env:RELEASE_TAG" $prereleaseFlag --draft --generate-notes --title "Release $env:RELEASE_TAG"
```
cfbao, line 41: This won't work as you expected. When you pass an empty string to a program on the command line, it's still recognized as an extra argument, which will usually mess up argument passing (when you really want the program to ignore the argument).

PowerShell has some hacks though:
```suggestion
          $prereleaseFlag = if ($env:PRERELEASE -eq $true) { '--prerelease' } else { '' }
          $releaseUrl = gh release create "$env:RELEASE_TAG" @prereleaseFlag --draft --generate-notes --title "Release $env:RELEASE_TAG"
```
`@variable` split the variable into separate arguments and pass them into the program. It's really meant for use with arrays and hash maps, but it can also work for strings because of weak typing. When `$variable` is empty, `@variable` becomes an empty array i.e. no argument at all.
See docs on splatting: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_splatting

To do this properly though, best to construct dynamic arguments as an array rather than a string.
Also, GitHub Actions inputs and environment variables are all ultimately strings. `-eq $true` kinda works because PowerShell has weak typing, but still best to use `-eq 'true'`
```suggestion
          $prereleaseFlag = if ($env:PRERELEASE -eq 'true') { @('--prerelease') } else { @() }
          $releaseUrl = gh release create "$env:RELEASE_TAG" @prereleaseFlag --draft --generate-notes --title "Release $env:RELEASE_TAG"
```

.github/workflows/publish.yml
```diff
@@ -11,23 +11,38 @@ on:
...
+      description: 'If true, the release will be set as a prerelease and not impact the latest tag'
+      type: boolean
+      required: true
+      default: false
+
 
 jobs:
   create_release:
-    if: github.event_name == 'repository_dispatch'
+    if: github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main'
```
cfbao, line 28: how about
```suggestion
    if: github.event_name == 'workflow_dispatch' && (github.ref == 'refs/heads/main' || github.events.input.prerelease == 'true')
```

seems fine to create prereleases even when it's not main
(`== 'true'` is needed because all GitHub Actions inputs are ultimately just strings)

.github/workflows/publish.yml
```diff
@@ -96,14 +111,29 @@ jobs:
...
         working-directory: build
         run: |
           chmod +x "bmx"
           tar -czvf "bmx-$PLATFORM.tar.gz" "bmx"
           gh release upload "$RELEASE_TAG" "bmx-$PLATFORM.tar.gz"
+
+  publish_release:
+    needs: [create_release, build, build_docker]
+    if: |
+      github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main'
```
cfbao, line 129: if `create_release` already has the correct conditions, we can omit it here
```suggestion
```

.github/workflows/publish.yml
gord5500: Hey this is a cool trick. I've looked at the docs and tried this in my own powershell and can't get it working. 

Using `$a = if ($PRERELEASE -eq 'true') { @('--prerelease') } else { @() }` I get an error `The error message is: CreateFile -: The system cannot find the file specified.` which makes me think for some reason the github command thinks that argument is a file path? 

The weird thing though is if I do `$a = @('--prerelease')` that works fine and creates the release. I'm confused that somehow the if statement is causing a problem for the github command. 

The line I'm using to try creating a release is ` $releaseUrl = gh release create "vtest" --draft @a`

gord5500: I tried using a hashtable to get around this but then the formatting is bad `bad flag syntax: --prerelease:True`

gord5500: Thought I could do some flashy line but maybe it's just easier if I do a standard if else block to check the value of `$env:PRERELEASE`

.github/workflows/publish.yml
gord5500: Yea you're right this will help with testing too

.github/workflows/publish.yml
gord5500: Good point on that

.github/workflows/publish.yml
cfbao: yeah regular if else block please

gord5500: 🤝

PR #288, opened 2023-05-18
choose MFA option if there is only one

DannyDaoBoYang:
### Why
- Skip MFAoption selection if only 1 option is available.

### DIFF
--- a/src/D2L.Bmx/Authenticate.cs
+++ b/src/D2L.Bmx/Authenticate.cs
@@ -126,17 +126,23 @@ internal class Authenticator {
 	}
 
 	private static int PromptMfa( MfaOption[] mfaOptions ) {
-
 		Console.WriteLine( "MFA Required" );
-		for( int i = 0; i < mfaOptions.Length; i++ ) {
-			Console.WriteLine( $"[{i + 1}] {mfaOptions[i].Provider}: {mfaOptions[i].Name}" );
-		}
-		Console.Write( "Select an available MFA option: " );
-		if( !int.TryParse( Console.ReadLine(), out int index ) || index > mfaOptions.Length || index < 1 ) {
-			throw new BmxException( "Invalid account selection" );
+		if( mfaOptions.Length > 1 ) {
+			for( int i = 0; i < mfaOptions.Length; i++ ) {
+				Console.WriteLine( $"[{i + 1}] {mfaOptions[i].Provider}: {mfaOptions[i].Name}" );
+			}
+			Console.Write( "Select an available MFA option: " );
+			if( !int.TryParse( Console.ReadLine(), out int index ) || index > mfaOptions.Length || index < 1 ) {
+				throw new BmxException( "Invalid account selection" );
+			}
+			return index;
+		} else if( mfaOptions.Length == 0 ) {//idk, is mfaOptions' length gaurenteed to be >= 1?
+			throw new BmxException( "No MFA method have been set up for the current user." );
+		} else {
+			Console.WriteLine( $"MFA method: {mfaOptions[0].Provider}: {mfaOptions[0].Name}" );
+			return 1;
 		}
 
-		return index;
 	}
 
 	private static string PromptMfaInput( string mfaInputPrompt ) {


src/D2L.Bmx/Authenticate.cs
```diff
@@ -124,19 +126,39 @@ IOktaApi oktaApi
...
-
-	private static int PromptMfa( MfaOption[] mfaOptions ) {
-
-		Console.WriteLine( "MFA Required" );
+	private static int MatchDefaultMFAoption( MfaOption[] mfaOptions, string DefautMFAProvider, string DefaultMFAmethod ) {
 		for( int i = 0; i < mfaOptions.Length; i++ ) {
-			Console.WriteLine( $"[{i + 1}] {mfaOptions[i].Provider}: {mfaOptions[i].Name}" );
+			if( ( mfaOptions[i].Provider == DefautMFAProvider ) && ( mfaOptions[i].Name == DefaultMFAmethod ) ) {
+				return i;
+			}
```
gord5500, line 133:
```suggestion
			return Array.IndexOf(mfaOptions, mfaOptions.FirstOrDefault(option => option.Provider == DefautMFAProvider && option.Name == DefaultMFAmethod))
```

[LINQ ](https://learn.microsoft.com/en-us/dotnet/standard/linq/)is pretty op once you start using it a bit. Not saying you should change what you have though

DannyDaoBoYang: sure, will use that

src/D2L.Bmx/Authenticate.cs
```diff
@@ -124,19 +126,38 @@ IOktaApi oktaApi
...
-	private static int PromptMfa( MfaOption[] mfaOptions ) {
-
+	private static int MatchDefaultMFAoption( MfaOption[] mfaOptions, string DefautMFAProvider, string DefaultMFAmethod ) {
+		return Array.IndexOf( mfaOptions,
+			Array.Find( mfaOptions, option =>
+				option.Provider == DefautMFAProvider && option.Name == DefaultMFAmethod
+)
+		);
+	}
+	private static int PromptMfa( MfaOption[] mfaOptions, string DefautMFAProvider, string DefaultMFAmethod ) {
```
gord5500, line 136:
```suggestion
	private static int PromptMfa( MfaOption[] mfaOptions, string defaultMfaProvider, string defaultMfaMethod ) {
```
Looks like the other areas of code have `Mfa` instead of `MFA` all caps

gord5500: Also want to use camelCase

DannyDaoBoYang: Sure will change that

src/D2L.Bmx/Program.cs
```diff
@@ -55,6 +55,9 @@
 var profileOption = new Option<string>(
 	name: "--profile",
 	description: "aws profile name" );
+var MFAOption = new Option<string>(
+	name: "--dftMFA",
+	description: "the defaultMFA in the provider_method format (eg. OKTA_question)" );
```
cfbao, line 60: Looking at the implementation in this PR and [Okta's doc on MFA API](https://developer.okta.com/docs/reference/api/factors/), I have doubt on how worthwhile this "default MFA option" feature really is.
The description here is cryptic, and hard to make it understandable too. Even if users understand this, it's still hard for users to know what to enter here (or in the config). Who would've thought of entering `GOOGLE_token:software:totp` for example?

Selecting the only MFA option when appropriate is a nice-to-have and an easy win, but I'm not convinced we should implement default MFA selection in the config or on the command line.

boarnoah: Agreed, the way Okta exposes factors, names how the names are written is really no good.
Since the MFA case is already an edge case for Okta authn the value is also lessened.

DannyDaoBoYang:
Updated branch. So the only change here will be to choose MFA option if there is only one

cfbao:
> Updated branch. So the only change here will be to choose MFA option if there is only one

Thanks! Can you update the PR title appropriately too?

DannyDaoBoYang:
> > Updated branch. So the only change here will be to choose MFA option if there is only one
> 
> Thanks! Can you update the PR title & description accordingly too?

Yes. Updated.

cfbao (approved):

PR #294, opened 2023-05-19
make classes non-static where appropriate

cfbao:
### Why

Code that directly interacts with external systems should generally be put behind interfaces and not be in static classes/methods. This is necessary to have any sort of unit tests. The .NET bmx doesn't have unit tests right now, but we should still follow this general principle as a best practice.

Main changes:

* Consolidate all user console prompts into `ConsolePrompter` and put it behind an interface.
* Write all console prompts to stderr instead of stdout to make `bmx print` usable.
* Put `OktaSessionStorage` behind an interface too
* Rename `Authenticator` to `OktaAuthenticator` and move it to the Okta namespace

Note: I'm not putting all classes behind interfaces (e.g. `OktaAuthenticator`). Even when we want to unit test the code, code that doesn't directly interact with external systems doesn't necessarily need to be mocked. We can introduce interfaces when/if such a need arises.

### DIFF
--- a/src/D2L.Bmx/ConfigureHandler.cs
+++ b/src/D2L.Bmx/ConfigureHandler.cs
@@ -18,11 +18,11 @@ internal class ConfigureHandler {
 
 		if( string.IsNullOrEmpty( org ) ) {
 			org = _consolePrompter.PromptOrg();
-		};
+		}
 
 		if( string.IsNullOrEmpty( user ) ) {
 			user = _consolePrompter.PromptUser();
-		};
+		}
 
 		if( defaultDuration is null ) {
 			defaultDuration = _consolePrompter.PromptDefaultDuration();
--- a/src/D2L.Bmx/ConfigureHandler.cs
+++ b/src/D2L.Bmx/ConfigureHandler.cs
@@ -3,9 +3,11 @@ namespace D2L.Bmx;
 internal class ConfigureHandler {
 
 	private readonly IBmxConfigProvider _configProvider;
+	private readonly IConsolePrompter _consolePrompter;
 
-	public ConfigureHandler( IBmxConfigProvider configProvider ) {
+	public ConfigureHandler( IBmxConfigProvider configProvider, IConsolePrompter consolePrompter ) {
 		_configProvider = configProvider;
+		_consolePrompter = consolePrompter;
 	}
 
 	public void Handle(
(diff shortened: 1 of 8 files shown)


src/D2L.Bmx/ConfigureHandler.cs
```diff
@@ -15,15 +17,15 @@ internal class ConfigureHandler {
 	) {
 
 		if( string.IsNullOrEmpty( org ) ) {
-			org = ConsolePrompter.PromptOrg();
+			org = _consolePrompter.PromptOrg();
 		};
```
boarnoah, line 21: didnt notice this before, but the ``;`` here and line 23

src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -32,8 +46,27 @@ internal static class ConsolePrompter {
 		return user;
 	}
 
-	public static int? PromptDefaultDuration() {
-		Console.Write( "Default duration of session in minutes (optional, default: 60): " );
+	string IConsolePrompter.PromptPassword() {
+		Console.Error.Write( "Okta Password: " );
+		string password = "";
+
+		// TODO: remove nomask
```
boarnoah, line 53: not relevant anymore

src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -32,8 +46,27 @@ internal static class ConsolePrompter {
...
+		while( true ) {
+			var input = Console.ReadKey( intercept: true );
+			if( input.Key == ConsoleKey.Enter ) {
+				Console.Error.Write( '\n' );
+				break;
+			} else if( input.Key == ConsoleKey.Backspace && password.Length > 0 ) {
```
boarnoah, line 58: IDE isn't a fan of this if else chain here, nor the one in ``PromptMfa``.

It does feel like a clearer read to me if it is separate, or alternatively like:
https://github.com/Brightspace/bmx/blob/46f5346c6b1be35e158d8366bedaf96b43980c07/Bmx.CommandLine/CommandLine.cs#L102-L112

WDYT?

boarnoah (approved):

src/D2L.Bmx/ConsolePrompter.cs
cfbao: What's the IDE's suggestion here?
I pretty often their suggestions on if else vs expression for example, because I often find them just bad...

boarnoah:
```csharp
if( input.Key == ConsoleKey.Enter ) {
	Console.Error.Write( '\n' );
	break;
}

if( input.Key == ConsoleKey.Backspace && password.Length > 0 ) {
	password = password.Remove( password.Length - 1, 1 );
} else if( !char.IsControl( input.KeyChar ) ) {
	password += input.KeyChar;
}
```

I do like it in the context of loop like this to have exit conditions separate from the rest. 

> I'm a little reluctant to do more cursor manipulation here...

Just the do while loop over the input to be clear (move the exit condition out of the loop body)

cfbao: oh I see. I do like early exit.
there's quite a few places in bmx that can be updated for early exit.
I'll update these later.

I have another change lined up to change the whole password input thing. constructing a new string on every key stroke is bad form in itself...

PR #303, opened 2023-05-23
consolidate code shared between print & write commands

cfbao:
### Why

The print and write commands are identical for the most part except their final outputs, so it makes sense to consolidate their code too.

The process is split into three parts
1. authenticate to Okta (now all put in `OktaAuthenticator.AuthenticateAsync`)
2. use Okta session to obtain AWS creds (now all put in `AwsCredsCreator.CreateAwsCredsAsync`)
3. output the AWS creds (remain separate in print & write command handlers)

Step 1 can also be utilized/reused by the future `bmx login` command we talked about.

This also makes some future refactoring I'd like to do easier.

### DIFF
--- a/src/D2L.Bmx/AwsCredsCreator.cs
+++ b/src/D2L.Bmx/AwsCredsCreator.cs
@@ -1,0 +1,65 @@
+using D2L.Bmx.Aws;
+using D2L.Bmx.Okta;
+
+namespace D2L.Bmx;
+
+internal class AwsCredsCreator {
+	private readonly IAwsClient _awsClient;
+	private readonly IConsolePrompter _consolePrompter;
+	private readonly BmxConfig _config;
+
+	public AwsCredsCreator(
+		IAwsClient awsClient,
+		IConsolePrompter consolePrompter,
+		BmxConfig config
+	) {
+		_awsClient = awsClient;
+		_consolePrompter = consolePrompter;
+		_config = config;
+	}
+
+	public async Task<AwsCredentials> CreateAwsCredsAsync(
+		IOktaApi oktaApi,
+		string? account,
+		string? role,
+		int? duration,
+		bool nonInteractive
+	) {
+		var accountState = await oktaApi.GetAccountsAsync( "amazon_aws" );
+		string[] accounts = accountState.Accounts;
+
+		if( string.IsNullOrEmpty( account ) ) {
+			if( !string.IsNullOrEmpty( _config.Account ) ) {
+				account = _config.Account;
+			} else if( !nonInteractive ) {
+				account = _consolePrompter.PromptAccount( accounts );
+			} else {
+				throw new BmxException( "Account value was not provided" );
+			}
+		}
+
+		string accountCredentials = await oktaApi.GetAccountAsync( accountState, account );
+		var roleState = _awsClient.GetRoles( accountCredentials );
+		string[] roles = roleState.Roles;
+
+		if( string.IsNullOrEmpty( role ) ) {
+			if( !string.IsNullOrEmpty( _config.Role ) ) {
+				role = _config.Role;
+			} else if( !nonInteractive ) {
+				role = _consolePrompter.PromptRole( roles );
+			} else {
+				throw new BmxException( "Role value was not provided" );
+			}
+		}
+
+		if( duration is null or 0 ) {
+			if( _config.DefaultDuration is not ( null or 0 ) ) {
+				duration = _config.DefaultDuration;
+			} else {
+				duration = 60;
+			}
+		}
+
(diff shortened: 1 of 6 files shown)


src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -1,37 +1,64 @@
...
+			return _oktaApi;
 		} else if( nonInteractive ) {
 			throw new BmxException( "Authentication failed. No cached session" );
 		}
 
 		string password = _consolePrompter.PromptPassword();
 
-		var authState = await oktaApi.AuthenticateAsync( new AuthenticateOptions( user, password ) )
-			.ConfigureAwait( false );
+		var authState = await _oktaApi.AuthenticateAsync( new AuthenticateOptions( user, password ) );
```
cfbao, line 61: `.ConfigureAwait( false )` isn't meaningful in a console (or ASP.NET Core) app

src/D2L.Bmx/PrintHandler.cs
```diff
@@ -34,82 +23,20 @@ IConsolePrompter consolePrompter
...
-			}
-		}
-
-		var tokens = await _awsClient.GetTokensAsync( roleState, role, duration.GetValueOrDefault() );
+		var oktaApi = await _oktaAuth.AuthenticateAsync( org, user, nonInteractive );
+		var awsCreds = await _awsCreds.CreateAwsCredsAsync( oktaApi, account, role, duration, nonInteractive );
 
 		if( string.Equals( output, "bash", StringComparison.OrdinalIgnoreCase ) ) {
-			PrintBash( tokens );
+			PrintBash( awsCreds );
```
cfbao, line 30: naming nit

src/D2L.Bmx/Program.cs
```diff
@@ -116,32 +120,41 @@
...
-	};
+var writeCommand = new Command( "write", "Write to AWS credentials file" ) {
+	accountOption,
+	durationOption,
+	orgOption,
+	writeOutputOption,
+	profileOption,
+	roleOption,
+	userOption,
+	nonInteractiveOption,
```
cfbao, line 131: the write command doesn't currently accept the `--non-interactive` flag, but I think it should for consistency.

src/D2L.Bmx/Program.cs
```diff
@@ -116,32 +120,41 @@
...
+};
 
 writeCommand.SetHandler( async ( InvocationContext context ) => {
 	var consolePrompter = new ConsolePrompter();
+	var config = new BmxConfigProvider().GetConfiguration();
```
cfbao, line 136: it's a lot easier to inject the config itself (rather than the config provider) into all objects that need it, so we can avoid reading the config files multiple times.

src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	public async Task<AwsCredentials> CreateAwsCredsAsync(
+		IOktaApi oktaApi,
+		string? account,
+		string? role,
+		int? duration,
+		bool nonInteractive
+	) {
+		var accountState = await oktaApi.GetAccountsAsync( "amazon_aws" );
```
boarnoah, line 28: In the future should we just hardcode this deeper in the API, in the context of BMX no other app is ever useful to filter for

src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -0,0 +1,65 @@
...
+				role = _config.Role;
+			} else if( !nonInteractive ) {
+				role = _consolePrompter.PromptRole( roles );
+			} else {
+				throw new BmxException( "Role value was not provided" );
+			}
+		}
+
+		if( duration is null or 0 ) {
+			if( _config.DefaultDuration is not ( null or 0 ) ) {
```
boarnoah, line 56: the lower bound is 15 minutes (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use.html) we check against that as the lower bound instead of 0

boarnoah (approved):

src/D2L.Bmx/AwsCredsCreator.cs
cfbao: This is a check against some sort of "default" value that might exist when the user didn't explicitly specify anything, both null & 0 are plausible values for that, and we fall back to 60 in such cases.

If a user specifies 5 (or something crazier like -5), I'd think letting the process error out (and making the user aware) would be a more reasonable behaviour than implicitly increasing the value to 15.

We can opt to add explicit checks and error behaviour here, but out of my intended scope here.

PR #305, opened 2023-05-23
configure: create .bmx folder and config file if needed

gord5500:
### Why

First time users will crash on finishing `bmx configure` command because the .bmx folder and config file does not exist

### DIFF
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -56,15 +56,8 @@ internal class BmxConfigProvider : IBmxConfigProvider {
 
 	public void SaveConfiguration( BmxConfig config ) {
 
-		string bmxDirectory = BmxPaths.BMX_DIR;
-		string configFileName = BmxPaths.CONFIG_FILE_NAME;
-
-		if( !Directory.Exists( bmxDirectory ) ) {
-			Directory.CreateDirectory( bmxDirectory );
-		}
-
-		if( !File.Exists( configFileName ) ) {
-			File.WriteAllText( configFileName, "" );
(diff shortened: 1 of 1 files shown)


src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -55,6 +55,18 @@ internal class BmxConfigProvider : IBmxConfigProvider {
 	}
 
 	public void SaveConfiguration( BmxConfig config ) {
+
+		string bmxDirectory = BmxPaths.BMX_DIR;
+		string configFileName = BmxPaths.CONFIG_FILE_NAME;
```
cfbao, line 60: on [L70](https://github.com/Brightspace/bmx/pull/305/files#diff-2eb5576e543a9b7cb2ee5a009fd10a20e4bbeb2a3115370f6d75de9403f7deb1R70) `BmxPaths.CONFIG_FILE_NAME` is used directly without the local variable. seems good to me to always do that unless there are other concerns (e.g. perf, race conditions (neither applies here))

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -55,6 +55,18 @@ internal class BmxConfigProvider : IBmxConfigProvider {
...
+
+		if( !Directory.Exists( bmxDirectory ) ) {
+			Directory.CreateDirectory( bmxDirectory );
+		}
+
+		if( !File.Exists( configFileName ) ) {
+			File.WriteAllText( configFileName, "" );
+		}
```
cfbao, line 68: We need to create the directory, but I don't think this is needed. StreamWriter should create it automatically.
(also using `File.WriteAllText` just to create an empty file feels bad)

cfbao (approved):

PR #306, opened 2023-05-23
clean up print output by using raw string literals

cfbao:
there's (almost) no reason to do ugly string escaping when C# 11 has the excellent [raw string literal][1] feature.

[1]: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/raw-string-literal

### DIFF
--- a/src/D2L.Bmx/Aws/AwsClient.cs
+++ b/src/D2L.Bmx/Aws/AwsClient.cs
@@ -62,7 +62,7 @@ internal class AwsClient : IAwsClient {
 				SessionToken: authResp.Credentials.SessionToken,
 				AccessKeyId: authResp.Credentials.AccessKeyId,
 				SecretAccessKey: authResp.Credentials.SecretAccessKey,
-				Expiration: authResp.Credentials.Expiration
+				Expiration: authResp.Credentials.Expiration.ToUniversalTime()
 			);
 		}
 		throw new BmxException( "Invalid role selection" );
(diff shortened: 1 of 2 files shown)


src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -62,7 +62,7 @@ internal class AwsClient : IAwsClient {
 				SessionToken: authResp.Credentials.SessionToken,
 				AccessKeyId: authResp.Credentials.AccessKeyId,
 				SecretAccessKey: authResp.Credentials.SecretAccessKey,
-				Expiration: authResp.Credentials.Expiration
+				Expiration: authResp.Credentials.Expiration.ToUniversalTime()
```
cfbao, line 65: I have trust issue with local time

src/D2L.Bmx/PrintHandler.cs
```diff
@@ -106,36 +106,39 @@ IConsolePrompter consolePrompter
...
-		Console.WriteLine( "}" );
+
+	private static void PrintJson( AwsCredentials credentials ) {
+		Console.WriteLine( $$"""
+		{
+			"Version": 1,
+			"AccessKeyId": "{{credentials.AccessKeyId}}",
+			"SecretAccessKey": "{{credentials.SecretAccessKey}}",
+			"SessionToken": "{{credentials.SessionToken}}",
+			"Expiration": "{{credentials.Expiration:o}}"
```
cfbao, line 140: we can use the standard format specifier
https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#the-round-trip-o-o-format-specifier

gord5500 (approved):

PR #307, opened 2023-05-23
refactor MFA code

cfbao:
### Why

Mostly to remove bloat like `MfaType` and `MfaOptions`. Also to handle MFA in a more generic fashion.


src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -112,27 +112,29 @@ internal class ConsolePrompter : IConsolePrompter {
...
-				Console.Error.WriteLine( $"[{i + 1}] {mfaOptions[i].Provider}: {mfaOptions[i].Name}" );
-			}
-			Console.Error.Write( "Select an available MFA option: " );
-			if( !int.TryParse( Console.ReadLine(), out int index ) || index > mfaOptions.Length || index < 1 ) {
-				throw new BmxException( "Invalid account selection" );
-			}
-			return index;
-		} else if( mfaOptions.Length == 0 ) {//idk, is mfaOptions' length guaranteed to be >= 1?
+
+		if( mfaOptions is not { Length: > 0 } ) {
```
cfbao, line 118: this catches both null and zero length `mfaOptions`
(but actually soon we won't need the null check because I'll update code in OktaApi to never return a nullable MFA factor array)

src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
```diff
@@ -8,7 +9,7 @@ internal record AuthenticateResponseInital(
 	string? StateToken,
 	string? SessionToken,
 	[property: JsonPropertyName( "_embedded" )]
-	AuthenticateResponseEmbeddedInitial Embedded
+	AuthenticateResponseEmbeddedInitial? Embedded
```
cfbao, line 12: there's no guarantee that this would exist. similarly here:
https://github.com/Brightspace/bmx/pull/307/files#diff-3da8e534d020fe4f4d0a2b1e6853e57fb85322b1a3a7c39e274c30acac0820b8L22-R23

src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
```diff
@@ -19,9 +20,16 @@ internal record AuthenticateResponseSuccess(
 );
 
 internal struct AuthenticateResponseEmbeddedInitial {
-	public OktaMfaFactor[] Factors { get; set; }
+	public OktaMfaFactor[]? Factors { get; set; }
 }
 
+internal record OktaMfaFactor(
```
cfbao, line 26: moved from https://github.com/Brightspace/bmx/pull/307/files#diff-8e35476e12249edf757d005b7c430c8b983ceecef18e4e6376dc77ffa32ee277

src/D2L.Bmx/Okta/Models/MfaType.cs
cfbao, line 1: this file is bloat...

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -145,4 +137,16 @@ IOktaApi oktaApi
...
+	private static bool IsMfaFactorTypeSupported( string mfaFactorType ) {
+		return mfaFactorType is
+			"call"
+			or "email"
+			or "question"
+			or "sms"
+			or "token:hardware"
+			or "token:hotp"
+			or "token:software:totp"
+			or "token";
```
cfbao, line 149: We currently support two kinds of MFA factors:
1. user enters a code/answer directly
2. we send a message (aka "challenge") to the user and the user enters a code (akak "challenge response") from that message

These all fall into the two kinds.

src/D2L.Bmx/ConsolePrompter.cs
boarnoah: TIl, cool syntax. It does border a little on being too clever. Seems fine for this use case though.

cfbao: From what I've seen on the Internet, pattern matching in C# is freaking divisive...
Many people just say it's "not readable", which I interpret to be "I haven't learned the new pattern matching syntax".
I'm obviously in the pro pattern matching group though 😛

boarnoah: Yeah I am guilty of that, re: rejecting concise syntax because its unfamiliar rather than because its actually too cheeky. 😆

PR #309, opened 2023-05-23
update build configuration

cfbao:
The biggest the motivation for the change is to enable [invariant globalization mode][1].
The default Alpine image doesn't have an ICU library, so the previous bmx Alpine binaries can't run on a clean Alpine container. Enabling invariant globalization removes the ICU dependency, which makes bmx run well on Alpine, and reduces the binary size a bit, too.

[1]: https://github.com/dotnet/runtime/blob/main/docs/design/features/globalization-invariant-mode.md

### DIFF
--- a/Dockerfile.ubuntu
+++ b/Dockerfile.ubuntu
@@ -1,16 +1,19 @@
 FROM ubuntu:18.04 AS build
 
-RUN apt-get update && apt-get install -y --no-install-recommends wget ca-certificates
-RUN wget https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb \
-    --no-check-certificate -O packages-microsoft-prod.deb
-RUN dpkg -i packages-microsoft-prod.deb
-RUN rm packages-microsoft-prod.deb
-
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends \
-       dotnet-sdk-7.0 clang zlib1g-dev
+    && apt-get install -y --no-install-recommends wget ca-certificates clang zlib1g-dev
+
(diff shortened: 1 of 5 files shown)


src/D2L.Bmx/D2L.Bmx.csproj
```diff
@@ -5,20 +5,19 @@
     <PublishAot>true</PublishAot>
     <OutputType>Exe</OutputType>
     <TargetFramework>net7.0</TargetFramework>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
-    <WarningsAsErrors>nullable</WarningsAsErrors>
+    <InvariantGlobalization>true</InvariantGlobalization>
+    <RootNamespace>D2L.Bmx</RootNamespace>
```
cfbao, line 9: some IDE/build tool features work more nicely with this specified when the assembly name is different.

scowing (approved):

Dockerfile.ubuntu
```diff
@@ -1,17 +1,20 @@
...
+
+RUN wget https://dot.net/v1/dotnet-install.sh \
+    && chmod +x ./dotnet-install.sh \
+    && ./dotnet-install.sh --channel 7.0 --install-dir /usr/share/dotnet
 
 WORKDIR /source
 
+# we have InvariantGlobalization=true in csproj file, so this shouldn't be necessary,
+# but the build still blows up without this env var ¯\_(ツ)_/¯
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=true
```
boarnoah, line 16: :(

boarnoah (approved):

PR #310, opened 2023-05-24
Set user read/write permissions when creating files

DannyDaoBoYang:
Update both session and config's permission to 600 on creation

### DIFF
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -69,9 +69,10 @@ internal class BmxConfigProvider : IBmxConfigProvider {
 			//maybe we cannot delete it somehow
 		}
 
-		var op = new FileStreamOptions();
-		op.Mode = FileMode.Create; //Append false, overrite old
-		op.Access = FileAccess.ReadWrite;
+		var op = new FileStreamOptions {
+			Mode = FileMode.Create,
+			Access = FileAccess.Write,
+		};
 		if( !RuntimeInformation.IsOSPlatform( OSPlatform.Windows ) ) {
 			op.UnixCreateMode = UnixFileMode.SetUser | UnixFileMode.UserRead | UnixFileMode.UserWrite;
 		}
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -60,21 +60,12 @@ internal class BmxConfigProvider : IBmxConfigProvider {
 		if( !Directory.Exists( BmxPaths.BMX_DIR ) ) {
 			Directory.CreateDirectory( BmxPaths.BMX_DIR );
 		}
-		//File permission is only applied on Create, so must delete old file
-		try {
-			if( File.Exists( BmxPaths.CONFIG_FILE_NAME ) ) {
-				File.Delete( BmxPaths.CONFIG_FILE_NAME );
-			}
-		} catch( IOException IOException ) {
-			//maybe we cannot delete it somehow
-		}
-
 		var op = new FileStreamOptions {
 			Mode = FileMode.Create,
 			Access = FileAccess.Write,
 		};
 		if( !RuntimeInformation.IsOSPlatform( OSPlatform.Windows ) ) {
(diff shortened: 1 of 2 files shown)


src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -55,12 +57,25 @@ internal class BmxConfigProvider : IBmxConfigProvider {
...
+			//maybe we cannot delete it somehow
+		}
 
-		using var writer = new StreamWriter( BmxPaths.CONFIG_FILE_NAME, append: false );
+		var op = new FileStreamOptions();
+		op.Mode = FileMode.Create; //Append false, overrite old
+		op.Access = FileAccess.ReadWrite;
+		if( !RuntimeInformation.IsOSPlatform( OSPlatform.Windows ) ) {
+			op.UnixCreateMode = UnixFileMode.SetUser | UnixFileMode.UserRead | UnixFileMode.UserWrite;
+		}
```
cfbao, line 77: We don't need `SetUser` right? That's SUID https://www.redhat.com/sysadmin/suid-sgid-sticky-bit

I'm also thinking of using [conditional compilation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation) for this, i.e.
```csharp
		var op = new FileStreamOptions {
			Mode = FileMode.Create,
			Access = FileAccess.Write,
#if !WINDOWS
			UnixCreateMode = UnixFileMode.UserRead | UnixFileMode.UserWrite,
#endif
		};
```
This moves OS detection from runtime to compile time, which is kinda nice. I don't feel strongly though.

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -9,12 +10,31 @@ internal interface IOktaSessionStorage {
 }
 
 internal class OktaSessionStorage : IOktaSessionStorage {
-	void IOktaSessionStorage.SaveSessions( List<OktaSessionCache> sessions ) {
 
+	//Update permission to 600 when a new entry is added
+	private void WriteFile600( string path, string Content ) {
```
cfbao, line 15: It's customary to put private methods after public methods.
[This SO answer](https://stackoverflow.com/a/310967/5916310) gives a good overview, and we more or less (but not always) follow that order.

The content of this method may also need to change pending discussion re: my earlier comment.
I'd also suggest a simpler name... `600` in the name seems confusing to me. Like `WriteTextToFile` maybe? I just don't want to emphasize the fact that we're setting file permissions in our code. As I said, I think of it as mostly a "feel good" measure...

cfbao:
Suggestion for PR title:
Set user read/write permissions when creating files

boarnoah:
We've mentioned how its mostly a feels good measure but I don't think we articulated it properly anywhere.

The reason its feels good is, in this particular caseif things have gone wrong to the point of malware being able to access these files out of the filesystem (presumably running with permissions of the user ex: something bad in an NPM CI script) then there is plenty of bad things they can do already.

ex:
- Firefox stores all cookies in a sqlite db in appdata
- Chrome stores all cookies in a sqlite db in appdata, the value is encrypted with the Windows Data Protection API. Which really only protects against accidental sharing, you can freely decrypt the values while you are operating in the context of the current user.

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -9,12 +10,14 @@ internal interface IOktaSessionStorage {
 }
 
 internal class OktaSessionStorage : IOktaSessionStorage {
-	void IOktaSessionStorage.SaveSessions( List<OktaSessionCache> sessions ) {
 
+	//Update permission to 600 when a new entry is added
+
```
cfbao, line 15:
```suggestion
```

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -41,4 +44,15 @@ internal class OktaSessionStorage : IOktaSessionStorage {
 		}
 		return new();
 	}
+	private void WriteTextToFile( string path, string Content ) {
```
cfbao, line 47: nit: we usually separate methods with empty lines
```suggestion

	private void WriteTextToFile( string path, string Content ) {
```

cfbao (approved):

cfbao (approved):

PR #312, opened 2023-05-24
Enable Inputing either org name or full domain 

DannyDaoBoYang:
### Why
To support Bmx with Okta preview.

### DIFF
--- a/src/D2L.Bmx/Okta/OktaApi.cs
+++ b/src/D2L.Bmx/Okta/OktaApi.cs
@@ -38,7 +38,11 @@ internal class OktaApi : IOktaApi {
 	}
 
 	void IOktaApi.SetOrganization( string organization ) {
-		_httpClient.BaseAddress = new Uri( $"https://{organization}.okta.com/api/v1/" );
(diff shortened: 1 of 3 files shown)


src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -38,7 +38,11 @@ internal class OktaApi : IOktaApi {
 	}
 
 	void IOktaApi.SetOrganization( string organization ) {
-		_httpClient.BaseAddress = new Uri( $"https://{organization}.okta.com/api/v1/" );
+		if( !organization.Contains( '.' ) ) {
```
cfbao, line 41: consider trimming trailing slashes, leading "http://" or leading "https://" to be more tolerant of various user inputs
e.g. "https://d2l.okta.com/"

cfbao (approved):

cfbao:
Please complete the PR description. specifically we'd like to have this to support using bmx with Okta preview.
It's important to write down context for future reference.

PR #313, opened 2023-05-24
refactor/clean up Okta authn code

cfbao:
* Remove bloat
  We have both `OktaAuthenticateState` and `OktaAuthenticatedState`... wut

* Simplify method signature and use better method/type names around Okta authentication API
  There're many `FooOptions` types that can be confusing when there's also the [options pattern][1] in .NET. Calling these types `FooRequest` is more in line with our convention.

* Better handle polymorphic responses from authentication API calls to give more accurate type (including nullability) info and helps clarify caller logic

[1]: https://learn.microsoft.com/en-us/dotnet/core/extensions/options

### DIFF
--- a/src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
+++ b/src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
@@ -2,26 +2,24 @@ using System.Text.Json.Serialization;
 
 namespace D2L.Bmx.Okta.Models;
 
-internal record AuthenticateResponseInital(
+internal abstract record AuthenticateResponse {
+	public record MfaRequired( string StateToken, OktaMfaFactor[] Factors ) : AuthenticateResponse;
+	public record Success( string SessionToken ) : AuthenticateResponse;
+}
+
+internal record AuthenticateResponseRaw(
 	DateTimeOffset ExpiresAt,
 	[property: JsonConverter( typeof( JsonStringEnumConverter ) )]
 	AuthenticationStatus Status,
 	string? StateToken,
 	string? SessionToken,
 	[property: JsonPropertyName( "_embedded" )]
-	AuthenticateResponseEmbeddedInitial? Embedded
+	AuthenticateResponseEmbedded? Embedded
 );
 
-internal record AuthenticateResponseSuccess(
-	DateTimeOffset ExpiresAt,
-	[property: JsonConverter( typeof( JsonStringEnumConverter ) )]
-	AuthenticationStatus Status,
-	string? SessionToken
+internal record AuthenticateResponseEmbedded(
+	OktaMfaFactor[]? Factors
 );
-
-internal struct AuthenticateResponseEmbeddedInitial {
-	public OktaMfaFactor[]? Factors { get; set; }
-}
 
 internal record OktaMfaFactor(
 	string Id,
--- a/src/D2L.Bmx/Okta/OktaApi.cs
+++ b/src/D2L.Bmx/Okta/OktaApi.cs
@@ -13,16 +13,16 @@ namespace D2L.Bmx.Okta;
 internal interface IOktaApi {
 	void SetOrganization( string organization );
 	void AddSession( string sessionId );
-	Task<OktaAuthenticateState> AuthenticateAsync( AuthenticateOptions authOptions );
-	Task<string> VerifyMfaChallengeResponseAsync(
-		OktaAuthenticateState state,
-		OktaMfaFactor mfaFactor,
+	Task<AuthenticateResponse> AuthenticateAsync( string username, string password );
+	Task IssueMfaChallengeAsync( string stateToken, string factorId );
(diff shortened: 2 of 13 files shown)


src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
```diff
@@ -2,27 +2,25 @@
...
 	[property: JsonPropertyName( "_embedded" )]
-	AuthenticateResponseEmbeddedInitial? Embedded
+	AuthenticateResponseEmbedded? Embedded
 );
 
-internal record AuthenticateResponseSuccess(
-	DateTimeOffset ExpiresAt,
-	[property: JsonConverter( typeof( JsonStringEnumConverter ) )]
-	AuthenticationStatus Status,
-	string? SessionToken
```
cfbao, line 19: Both the primary `authn` call and the MFA `authn/factors/${factorId}/verify` call return the same [authentication transaction object](https://developer.okta.com/docs/reference/api/authn/#authentication-transaction-object). In particular, both of them can return a success status. So it's convenient to define their return types in a polymorphic fashion.

Because STJ's [polymorphic deserialization](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism) can't handle this scenario automatically, we will directly deserialize responses to `AuthenticateResponseRaw` first (with all properties included but most nullable), then manually create "tighter" polymorphic objects of type `AuthenticateResponse` depending on values of the raw object.

Related changes:
https://github.com/Brightspace/bmx/pull/313/files#diff-8b1e253817f43a932b584d4f407e885d9f64a848f09317feed09e86d5c3aa1cdR60-R78
https://github.com/Brightspace/bmx/pull/313/files#diff-8b1e253817f43a932b584d4f407e885d9f64a848f09317feed09e86d5c3aa1cdR107-R111

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -49,38 +48,35 @@ bool nonInteractive
...
 			// TODO: Handle retry
 			if( mfaFactor.FactorType is "sms" or "call" or "email" ) {
-				await _oktaApi.IssueMfaChallengeAsync( authState, mfaFactor );
+				await _oktaApi.IssueMfaChallengeAsync( mfaInfo.StateToken, mfaFactor.Id );
 			}
 			string mfaResponse = _consolePrompter.GetMfaResponse( mfaFactor.FactorType == "question" ? "Answer" : "PassCode" );
-			sessionToken = await _oktaApi.VerifyMfaChallengeResponseAsync( authState, mfaFactor, mfaResponse );
-		} else if( authState.OktaSessionToken is not null ) {
-			sessionToken = authState.OktaSessionToken;
+			authnResponse = await _oktaApi.VerifyMfaChallengeResponseAsync( mfaInfo.StateToken, mfaFactor.Id, mfaResponse );
```
cfbao, line 74: this is only possible because of the polymorphic return types, and enables simpler logic below at [L77-L78](https://github.com/Brightspace/bmx/pull/313/files#diff-640013b3685b8a2cbcee8c5b2e26b1afc631ea22dc293c44a6f402feab21ff73R77-R78)

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -140,14 +156,12 @@ string challengeResponse
...
 
-	async Task<string?> IOktaApi.GetMeResponseAsync( string sessionId ) {
+	async Task<string?> IOktaApi.GetCurrentUserIdAsync( string sessionId ) {
 		try {
 			using var meResponse = await _httpClient.GetAsync( "users/me" );
 			if( !meResponse.IsSuccessStatusCode ) {
 				return null;
 			}
-
-			string meJson = await meResponse.Content.ReadAsStringAsync();
```
cfbao, line 150: this isn't used at all...

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -95,22 +91,19 @@ bool nonInteractive
...
 		if( string.IsNullOrEmpty( sessionId ) ) {
-			return new( SuccessfulAuthentication: false, OktaSessionId: "" );
+			return false;
 		}
 
 		oktaApi.AddSession( sessionId );
-		string? userId = await oktaApi.GetMeResponseAsync( sessionId );
-		if( !string.IsNullOrEmpty( userId ) ) {
-			return new( SuccessfulAuthentication: true, OktaSessionId: userId );
-		}
```
cfbao, line 112: assigning user id to session id doesn't make sense... but `OktaSessionId` isn't used by any caller anyway, so removing it is the best.

boarnoah (approved):
> We have both OktaAuthenticateState and OktaAuthenticatedState... wut

Will say it had slightly more sense in the original plan for a stateless OktaAPI. The names weren't great.

PR #314, opened 2023-05-25
Don't create session file on read

cfbao:
### Why

Currently the .NET bmx will create the ~/.bmx directory and an empty session file when the user doesn't have anything configured, which is a weird behaviour.
Creating directory and file should be in the `SaveSessions` method. BMX should leave no trace if the user doesn't configure anything.

### DIFF
--- a/src/D2L.Bmx/Okta/OktaSessionStorage.cs
+++ b/src/D2L.Bmx/Okta/OktaSessionStorage.cs
@@ -6,43 +6,36 @@ namespace D2L.Bmx.Okta;
 
 internal interface IOktaSessionStorage {
 	void SaveSessions( List<OktaSessionCache> sessions );
-	List<OktaSessionCache> Sessions();
+	List<OktaSessionCache> GetSessions();
 }
 
 internal class OktaSessionStorage : IOktaSessionStorage {
 	void IOktaSessionStorage.SaveSessions( List<OktaSessionCache> sessions ) {
+		if( !Directory.Exists( BmxPaths.BMX_DIR ) ) {
+			Directory.CreateDirectory( BmxPaths.BMX_DIR );
+		}
+
 		string jsonString = JsonSerializer.Serialize(
 			sessions,
 			SourceGenerationContext.Default.ListOktaSessionCache );
 		WriteTextToFile( BmxPaths.SESSIONS_FILE_NAME, jsonString );
 	}
 
-	List<OktaSessionCache> IOktaSessionStorage.Sessions() {
-
-		string bmxDirectory = BmxPaths.BMX_DIR;
-		string sessionsFileName = BmxPaths.SESSIONS_FILE_NAME;
-
-		if( !Directory.Exists( bmxDirectory ) ) {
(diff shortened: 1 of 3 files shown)


src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -6,43 +6,36 @@ namespace D2L.Bmx.Okta;
 
 internal interface IOktaSessionStorage {
 	void SaveSessions( List<OktaSessionCache> sessions );
-	List<OktaSessionCache> Sessions();
+	List<OktaSessionCache> GetSessions();
```
cfbao, line 9: naming nit - to make it clearer that this is a read operation.
I've seen some libraries that have a method/function named `thing()` that actually creates a `thing`.

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -6,43 +6,36 @@ namespace D2L.Bmx.Okta;
...
 }
 
 internal class OktaSessionStorage : IOktaSessionStorage {
 	void IOktaSessionStorage.SaveSessions( List<OktaSessionCache> sessions ) {
+		if( !Directory.Exists( BmxPaths.BMX_DIR ) ) {
+			Directory.CreateDirectory( BmxPaths.BMX_DIR );
+		}
```
cfbao, line 16: It worked fine without this because we always read the session file before writing to it.
But
* We don't create the file on read now
* This method should be safe to call regardless of whether we called the read method before

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -6,43 +6,36 @@ namespace D2L.Bmx.Okta;
...
-
-		if( !Directory.Exists( bmxDirectory ) ) {
-			Directory.CreateDirectory( bmxDirectory );
-		}
-
-		if( !File.Exists( sessionsFileName ) ) {
-			WriteTextToFile( sessionsFileName, "[]" );
+	List<OktaSessionCache> IOktaSessionStorage.GetSessions() {
+		if( !File.Exists( BmxPaths.SESSIONS_FILE_NAME ) ) {
+			return new();
```
cfbao, line 26: no reason to do anything more if the session file doesn't exit

src/D2L.Bmx/Okta/OktaSessionStorage.cs
```diff
@@ -6,43 +6,36 @@ namespace D2L.Bmx.Okta;
...
+			return JsonSerializer.Deserialize( sessionsJson, SourceGenerationContext.Default.ListOktaSessionCache )
+				?? new();
 		} catch( JsonException ) {
 			return new();
 		}
-		return new();
 	}
 
-	private void WriteTextToFile( string path, string Content ) {
+	private static void WriteTextToFile( string path, string content ) {
```
cfbao, line 38: nit
* prefer static when appropriate
* parameter names should be camelCase

boarnoah (approved):

PR #316, opened 2023-05-29
add allow-project-config flag

gord5500:
### Why

We want to update our install script for D2L users so that they don't need to enter a value for this when running `bmx configure`

### DIFF
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -6,28 +6,31 @@ using Microsoft.Extensions.FileProviders.Physical;
 namespace D2L.Bmx;
 
 internal interface IBmxConfigProvider {
-	BmxConfig GetConfiguration();
+	BmxConfig GetConfiguration( bool allowProjectConfigs );
 	void SaveConfiguration( BmxConfig config );
 }
 
 internal class BmxConfigProvider : IBmxConfigProvider {
 
-	public BmxConfig GetConfiguration() {
+	public BmxConfig GetConfiguration( bool allowProjectConfigs ) {
+
 		// Main config is at ~/.bmx/config
 		string configFileName = BmxPaths.CONFIG_FILE_NAME;
 
 		// If set, we recursively look up from CWD (all the way to root) for additional bmx config files (labelled as .bmx)
 		var configBuilder = new ConfigurationBuilder().AddIniFile( configFileName, optional: true );
 
-		bool.TryParse( configBuilder.Build()["allow_project_configs"], out bool shouldReadProjectConfig );
+		if( !allowProjectConfigs ) {
+			bool.TryParse( configBuilder.Build()["allow_project_configs"], out allowProjectConfigs );
+		}
 
 		FileInfo? projectConfigInfo = null;
 
-		if( shouldReadProjectConfig ) {
+		if( allowProjectConfigs ) {
 			projectConfigInfo = GetProjectConfigFileInfo();
 		}
 
-		if( shouldReadProjectConfig && projectConfigInfo is not null ) {
+		if( allowProjectConfigs && projectConfigInfo is not null ) {
 			// Default file provider ignores files prefixed with ".", we need to provide our own as a result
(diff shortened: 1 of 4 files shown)


src/D2L.Bmx/BmxConfigProvider.cs
gord5500, line 1: I was thinking we'd want to use this flag for more than just the configure command. Just like how we can override org and user I thought we could add an override for this.

src/D2L.Bmx/Program.cs
```diff
@@ -27,10 +27,12 @@
 		result.ErrorMessage = "Invalid duration";
 	}
 } );
-var nonInteractiveOption = new Option<bool>(
+var nonInteractiveFlag = new Option<bool>(
 	name: "--non-interactive",
-	getDefaultValue: () => false,
```
gord5500, line 32: Wanted to get rid of this and do the same for the new one. My thought was how can I know whether or not the user did not include `--allow-project-configs` option or if they did `--allow-project-configs false`. And depending on that how do I know whether I should prompt them in the configure command for this value if the value was false

gord5500: So now it's just if the flag is present that implies the value is true

gord5500: I thought to rename it too but was 50/50 on whether we should keep to the format we have for everything else or include `Flag` to make it a bit more obvious

src/D2L.Bmx/BmxConfigProvider.cs
gord5500: How useful it can be I'm not sure but thought the functionality should be there since the other config options can be overriden

cfbao: I'd err on the side of caution and _not_ add new features without a good use case.
`allow_project_config` is already kinda hard to explain. Adding it to the "help" info of both "print" and "write" doesn't help making bmx easy to use & understand.

src/D2L.Bmx/ConfigureHandler.cs
```diff
@@ -28,7 +29,9 @@ internal class ConfigureHandler {
 			defaultDuration = _consolePrompter.PromptDefaultDuration();
 		}
 
-		bool allowProjectConfigs = _consolePrompter.PromptAllowProjectConfig();
+		if( !allowProjectConfigs ) {
+			allowProjectConfigs = _consolePrompter.PromptAllowProjectConfig();
```
cfbao, line 33: What if user wants to explicitly specify `--allow-project-configs false` on the CLI and not be prompted here?
I think we need to make `allowProjectConfigs` a `bool?` to allow three states: `null`, `true`, `false`.

src/D2L.Bmx/Program.cs
cfbao: without this, what happens when user doesn't supply the flag? what does `GetValueForOption` return?

cfbao: I'd like to keep calling of them "option" for consistency. I don't think "flag" makes things more obvious here.

gord5500: It's false if they don't supply the option

cfbao: So `getDefaultValue: () => false` isn't doing anything anyway? Yeah good to get rid of it here.

src/D2L.Bmx/ConfigureHandler.cs
gord5500: Yea I originally tried this but didn't change the type and I think that's where my issue was. So changing to `bool?` will fix that

src/D2L.Bmx/BmxConfigProvider.cs
gord5500: I took it out. My thought process was if we already allow the majority of the config files options to be overwritten why not have that functionality for all of them. 

There's no use for this on something like the jumpbox / a machine with multiple users? If they wanted to run a project with a .bmx file they're just pushed in to opening the file themselves and finding out the specific aws account / role from that?

PR #317, opened 2023-05-29
ugly workaround for dotnet/runtime#22314

cfbao:


src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -49,28 +58,43 @@ internal class ConsolePrompter : IConsolePrompter {
...
+			stty_orig=$(stty -g)
+			stty -echo
+			read -r password
+			stty "$stty_orig"
+			echo -n "$password"
+			""" );
+		startInfo.RedirectStandardOutput = true;
+		using var proc = Process.Start( startInfo ) ?? throw new BmxException( "no" );
+		proc.WaitForExit();
+		return proc.StandardOutput.ReadToEnd();
```
cfbao, line 92: 🤮 but it kinda works

src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -49,28 +61,42 @@ internal class ConsolePrompter : IConsolePrompter {
...
+			return new string( passwordChars.ToArray() );
 		}
 
-		return new string( passwordChars.ToArray() );
+		string? password;
+		try {
+			// disable the terminal from echoing user input
+			using var stty = Process.Start( "stty", "-echo" );
+			stty.WaitForExit();
+			password = _stdinReader.ReadLine();
```
cfbao, line 88: This is similar to calling [tcsetattr](https://pubs.opengroup.org/onlinepubs/009696799/functions/tcsetattr.html), which a more proper implementation would do.
E.g. see Python's implementation of [getpass](https://docs.python.org/3/library/getpass.html):
https://github.com/python/cpython/blob/c39500db5254c9efee9937f951f58d87ee14cf2f/Lib/getpass.py#L68-L80

`stty` also calls `tcsetattr` under the hood, and is itself in the POSIX standard, so this should be fairly reliable, and a lot simpler than doing native library calls ourselves.

boarnoah: Nice find

boarnoah (approved):

src/D2L.Bmx/PrintHandler.cs
```diff
@@ -43,9 +43,9 @@ AwsCredsCreator awsCreds
 
 	private static void PrintBash( AwsCredentials credentials ) {
 		Console.WriteLine( $"""
-			export AWS_SESSION_TOKEN='{credentials.SessionToken}'
-			export AWS_ACCESS_KEY_ID='{credentials.AccessKeyId}'
-			export AWS_SECRET_ACCESS_KEY='{credentials.SecretAccessKey}'
+			export AWS_SESSION_TOKEN={credentials.SessionToken}
+			export AWS_ACCESS_KEY_ID={credentials.AccessKeyId}
+			export AWS_SECRET_ACCESS_KEY={credentials.SecretAccessKey}
```
cfbao, line 48: These single quotes ended up in the variable values themselves, resulting in wrong creds.
The distinction between the shell's preprocessing of commands and the actual values can be subtle...

src/D2L.Bmx/ConsolePrompter.cs
cfbao: Both.
I was looking for general advice on reading password from user input on POSIX systems, and I also knew Python had a getpass implementation that I'd like to check out. Both direction converged on using [termios](https://pubs.opengroup.org/onlinepubs/7908799/xsh/termios.h.html).
`stty` is often mentioned in answers to "how to read password in shell scripts", and checking the source code ([coreutils](https://github.com/wertarbyte/coreutils/blob/master/src/stty.c), [busybox](https://github.com/jgunthorpe/busybox/blob/master/coreutils/stty.c)), I see it simply calls termios functions under the hood, so decided to use this simpler method.

I did also look up calling native functions from .NET in the meantime ([docs](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/)), but good cross-platform support (Mac & different distros of Linux) seems sketchy. There's also some boilerplate interop code (e.g. defining the termios struct) that's maybe too much for bmx.

There are also .NET libraries that help with calling libc (e.g. https://github.com/tmds/Tmds.LibC, https://github.com/KSBilodeau/TermiosLib), but they all use the older [`DllImportAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportattribute) not the newer [`LibraryImportAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.libraryimportattribute), and only the latter works with native AOT.

PR #320, opened 2023-05-31
nits

cfbao:


src/D2L.Bmx/SourceGenerationContext.cs
```diff
@@ -1,7 +1,7 @@
 using System.Text.Json.Serialization;
 using D2L.Bmx.Okta.Models;
 
-namespace D2L.Bmx.Okta;
+namespace D2L.Bmx;
```
cfbao, line 4: we have an editorconfig/analyzer rule to enforce namespace matching folder structure, but for some reason it's ineffective when there's a partial class inside.

src/D2L.Bmx/WriteHandler.cs
```diff
@@ -41,13 +41,10 @@ BmxConfig config
...
-		var credentialsFile = new SharedCredentialsFile();
-		if( !string.IsNullOrEmpty( output ) ) {
-			if( !Path.IsPathRooted( output ) ) {
-				output = "./" + output;
-			}
-			credentialsFile = new SharedCredentialsFile( output );
+		if( !string.IsNullOrEmpty( output ) && !Path.IsPathRooted( output ) ) {
+			output = "./" + output;
 		}
+		var credentialsFile = new SharedCredentialsFile( output );
```
cfbao, line 47: Previously, we create two `SharedCredentialsFile` objects if `output` is not empty, which is wasteful.

The `SharedCredentialsFile(string)` constructor accepts a null parameter.
See https://github.com/aws/aws-sdk-net/blob/bb43d8be152ba92f4aeeb0038b8d69bd2a243b18/sdk/src/Core/Amazon.Runtime/CredentialManagement/SharedCredentialsFile.cs#L217-L252

boarnoah (approved):

PR #321, opened 2023-05-31
add .dockerignore

cfbao:
reduce the size of Docker context so it's faster to send to the Docker engine, and improve cache utilization


boarnoah (approved):
Add ``.git``, another high churn directory that isn't relevant for the container

gord5500:
Neat

PR #322, opened 2023-05-31
revamp AWS account & role selection

cfbao:
To help make BMX's data processing clearer

* remove unnecessary wrapper type `OktaAccountState`
* remove complex HTHML & XML processing from API client types `OktaApi` & `AwsClient`, and move it into a dedicated helper class, and have the orchestrating caller `AwsCredsCreator.CreateAwsCredsAsync` call these processing methods explicitly
* nits & fixes

[hide whitespace changes](https://github.com/Brightspace/bmx/pull/322/files?w=1)


src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -1,14 +1,10 @@
-using System.Net;
-using System.Text;
-using System.Xml;
 using Amazon.SecurityToken;
 using Amazon.SecurityToken.Model;
 
 namespace D2L.Bmx.Aws;
 
 internal interface IAwsClient {
-	AwsRoleState GetRoles( string encodedSaml );
```
cfbao, line 10: Methods on an API client type should generally be about making API calls.
Parsing roles from the SAML response is purely internal processing, so moved it out.

src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -18,69 +14,23 @@ internal class AwsClient : IAwsClient {
...
-		}
-		return new AwsRoleState( roles, encodedSaml );
-	}
-
 	async Task<AwsCredentials> IAwsClient.GetTokensAsync(
-		AwsRoleState state,
-		string selectedRole,
-		int durationInMinutes ) {
-		var role = state.AwsRoles
-			.Find( role => string.Equals( role.RoleName, selectedRole, StringComparison.OrdinalIgnoreCase ) );
```
cfbao, line 50: Similarly this is also internal processing that should be done by the caller not the client method.
There's also no need for the client call to know about all the available roles.

src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -18,69 +14,23 @@ internal class AwsClient : IAwsClient {
...
-		return samlToken;
+		string samlAssertion,
+		AwsRole role,
+		int durationInMinutes
+	) {
+		var authResp = await _stsClient.AssumeRoleWithSAMLAsync( new AssumeRoleWithSAMLRequest {
+			PrincipalArn = role.PrincipalArn,
+			RoleArn = role.RoleArn,
+			SAMLAssertion = samlAssertion,
+			DurationSeconds = durationInMinutes * 60,
```
cfbao, line 26: `durationInMinutes` was not used at all... so the duration flag was not effective in .NET BMX

src/D2L.Bmx/HtmlXmlHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Xml;
+using D2L.Bmx.Aws;
+
+namespace D2L.Bmx;
+
+internal static partial class HtmlXmlHelper {
+	public static string GetSamlResponseFromLoginPage( string html ) {
+		// TODO: use proper HTML parsing instead of regex
+		var inputXml = new XmlDocument();
+		inputXml.LoadXml( SamlResponseInputRegex().Match( html ).Value );
```
cfbao, line 11: I see this comment https://github.com/Brightspace/bmx/pull/322/files#diff-8b1e253817f43a932b584d4f407e885d9f64a848f09317feed09e86d5c3aa1cdL173
but I don't think the HTML Okta returns is malformed. Maybe they cleaned it up at some point?
It's not valid XML, but it is valid HTML that can be parsed more properly using a library like [Html Agility Pack](https://html-agility-pack.net/)

src/D2L.Bmx/HtmlXmlHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+		var samlXml = new XmlDocument();
+		using( var ms = new MemoryStream( samlXmlBytes ) ) {
+			samlXml.Load( ms );
+		}
+
+		var nsManager = new XmlNamespaceManager( samlXml.NameTable );
+		nsManager.AddNamespace( "s2", "urn:oasis:names:tc:SAML:2.0:assertion" );
+		XmlNodeList roleNodes = samlXml.SelectNodes(
+			"""//s2:Attribute[@Name="https://aws.amazon.com/SAML/Attributes/Role"]/s2:AttributeValue""",
+			nsManager
```
cfbao, line 28: added namespace so we can query with the tag name for more accuracy

src/D2L.Bmx/HtmlXmlHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+		) ?? throw new BmxException( "Failed to retrieve roles" );
+
+		return roleNodes
+			.Cast<XmlElement>()
+			.Select( node =>
+				node.InnerText.Split( ',' ) is [string principalArn, string roleArn]
+				? new AwsRole(
+					RoleName: roleArn.Split( "/" )[^1],
+					PrincipalArn: principalArn,
+					RoleArn: roleArn
```
cfbao, line 38: * `is [string principalArn, string roleArn]` checks that the result is actually exactly two elements, and assigns them to the two new variables
* `[^1]` grabs the last element, because a role ARN may contain a "path" component, so `[1]` isn't accurate.

src/D2L.Bmx/HtmlXmlHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+				)
+				: throw new BmxException( "Failed to retrieve roles" ) )
+			.ToArray();
+	}
+
+	[GeneratedRegex( """<input name="SAMLResponse" type="hidden" value=".*?"/>""" )]
+	private static partial Regex SamlResponseInputRegex();
```
cfbao, line 45: fancy source generator stuff that .NET analyzer is nudging me to use

src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -18,69 +14,23 @@ internal class AwsClient : IAwsClient {
...
-	}
-
-	private XmlDocument ParseSamlToken( string encodedSaml ) {
-		string[] samlStatements = encodedSaml.Split( ";" );
-
-		// Process the B64 Encoded SAML string to get valid XML doc
-		var samlString = new StringBuilder();
-		foreach( string inputValueString in samlStatements ) {
-			samlString.Append( WebUtility.HtmlDecode( inputValueString ) );
-		}
```
cfbao, line 78: The processing here isn't necessary.
When the SAML response string is extracted from the value attribute [here](https://github.com/Brightspace/bmx/pull/322/files#diff-8b1e253817f43a932b584d4f407e885d9f64a848f09317feed09e86d5c3aa1cdL181), the XML parser already unescaped all the XML escape sequences.
Also, the semicolon is part of an XML escape sequence, so there was no reason to split the string by semicolon even if we do need to unescape the XML string.

src/D2L.Bmx/HtmlXmlHelper.cs
boarnoah: Don't recall what that comment was about :/ Think theis current XML parsing is fine.

boarnoah (approved):

src/D2L.Bmx/HtmlXmlHelper.cs
boarnoah: Looks wild as with all this source generator stuff for the JSON serialization

PR #324, opened 2023-06-05
control all key strokes when entering password

cfbao:
### Why

If we only turn off input echoing, a user pressing Ctrl+C while entering password in a `$(bmx print)` will stop BMX and have their terminal display "crippled" - input echoing still off.
To resolve this issue, use raw mode in the terminal and receive all key strokes, and handle Ctrl+C behaviour ourselves.
This also largely unifies the password input processing logic across both Windows & *nix platforms.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,8 @@
-# DotNet7
-**/bin
-**/obj
-**/.vs
-.vs/
+build/
+publish/
 
-# Rider
+[Bb]in/
+[Oo]bj/
+.vs/
+.vscode/
 .idea/
--- a/src/D2L.Bmx/ConsolePrompter.cs
+++ b/src/D2L.Bmx/ConsolePrompter.cs
@@ -98,7 +98,10 @@ internal class ConsolePrompter : IConsolePrompter {
 				char key = readKey();
 
 				if( key == CTRL_C ) {
-					throw new BmxException( "" );
+					// Ctrl+C should terminate the program.
+					// Using an empty string as the exception message because this message is displayed to the user,
+					// but the user doesn't need to see anything when they themselves ended the program.
+					throw new BmxException( string.Empty );
 				}
 				if( key == '\n' || key == '\r' ) {
 					// when the terminal is in raw mode, writing \r is needed to start the new line properly
--- a/src/D2L.Bmx/ConsolePrompter.cs
+++ b/src/D2L.Bmx/ConsolePrompter.cs
@@ -1,5 +1,5 @@
 using System.Diagnostics;
-using System.Runtime.InteropServices;
+using System.Text;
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,8 +1,8 @@
-# DotNet7
-**/bin
-**/obj
-**/.vs
-.vs/
+build/
+publish/
 
-# Rider
```
cfbao, line 7: The section names here aren't accurate.
Bin & obj are .NET specific but not .NET 7 specific.
.vs is Visual Studio specific, not .NET.
I opted to remove all section headers here because we have so few of these.

src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -59,38 +64,68 @@ internal class ConsolePrompter : IConsolePrompter {
...
+			if( Console.IsInputRedirected ) {
+				Console.Error.WriteLine( """
+					====== WARNING ======
+					Input to BMX is redirected. Password input may be displayed on screen!
+					If you're using mintty (with Git Bash, Cygwin, MSYS2 etc.), consider switching
+					to Windows Terminal for a better experience.
+					If you must use mintty, prefix your bmx command with 'winpty '.
+					=====================
+					""" );
+				readKey = () => (char)_stdinReader.Read();
```
cfbao, line 79: See
https://d2l.slack.com/archives/G4P099831/p1685968878171859
https://trello.com/c/B9pBkurG/1200-support-mintty

src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -59,38 +64,68 @@ internal class ConsolePrompter : IConsolePrompter {
...
+			var passwordBuilder = new StringBuilder();
 			while( true ) {
-				ConsoleKeyInfo input = Console.ReadKey( intercept: true );
-				if( input.Key == ConsoleKey.Enter ) {
-					Console.Error.Write( '\n' );
-					break;
+				char key = readKey();
+
+				if( key == CTRL_C ) {
+					throw new BmxException( "" );
```
boarnoah, line 101: Seems a little naughty to execute the finally block by throwing a fake exception

boarnoah (approved):

src/D2L.Bmx/ConsolePrompter.cs
cfbao: I wouldn't say it's a fake exception though, it actually ends the program. Putting cleanup inside `finally` is pretty standard.

I put an empty string here because this string is displayed to the user and I don't think the user needs to see anything when they themselves pressed Ctrl+C. Maybe a comment there would make sense.

boarnoah (approved):

PR #325, opened 2023-06-05
don't show exception details to user

cfbao:
Having a full screen of unintelligible text dumped on the user is bad UX.

[Hide whitespace changes](https://github.com/Brightspace/bmx/pull/325/files?w=1)

### DIFF
--- a/src/D2L.Bmx/Program.cs
+++ b/src/D2L.Bmx/Program.cs
@@ -60,21 +60,17 @@ var configureCommand = new Command( "configure", "Create a bmx config file to sa
 	durationOption,
 };
 
-configureCommand.SetHandler( ( InvocationContext context ) => {
+configureCommand.SetHandler( ( InvocationContext context ) => RunWithErrorHandlingAsync( context, () => {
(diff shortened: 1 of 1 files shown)


src/D2L.Bmx/Program.cs
```diff
@@ -147,23 +138,37 @@
...
 rootCommand.Add( writeCommand );
 
 return await rootCommand.InvokeAsync( args );
+
+static async Task RunWithErrorHandlingAsync( InvocationContext context, Func<Task> handle ) {
+	try {
+		await handle();
+	} catch( Exception e ) {
+		Console.ResetColor();
+		Console.ForegroundColor = ConsoleColor.Red;
```
cfbao, line 162: not sure why the color needs to be reset before setting it to what we want, but it's what System.CommandLine does internally
https://github.com/dotnet/command-line-api/blob/2.0.0-beta4.22272.1/src/System.CommandLine/Invocation/ParseErrorResult.cs#L21-L33

scowing (approved):

PR #326, opened 2023-06-06
limit max search depth for project config file

cfbao:
### Why

To ensure that project config file lookup won't crash the app or get it stuck.

Related discussion: https://d2l.slack.com/archives/G4P099831/p1685471844688699

[Hide whitespace changes](https://github.com/Brightspace/bmx/pull/326/files?w=1)

### DIFF
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -11,7 +11,6 @@ internal interface IBmxConfigProvider {
 }
 
 internal class BmxConfigProvider : IBmxConfigProvider {
-
 	public BmxConfig GetConfiguration() {
 		// Main config is at ~/.bmx/config
 		string configFileName = BmxPaths.CONFIG_FILE_NAME;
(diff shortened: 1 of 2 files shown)


src/D2L.Bmx/BmxPaths.cs
```diff
@@ -1,7 +1,6 @@
-
 namespace D2L.Bmx;
-internal class BmxPaths {
 
+internal static class BmxPaths {
```
cfbao, line 3: unrelated nit

gord5500 (approved):
It would crash on linux before this?

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -75,23 +74,30 @@ internal class BmxConfigProvider : IBmxConfigProvider {
...
+			// limit the maximum search depth to avoid infinite loop even if there's any symlink trickery
+			for( int i = 0; i < 2000 && currentDirectory is not null; i++ ) {
+				var bmxConfigFiles = currentDirectory.GetFiles( ".bmx" );
 
-			if( bmxConfigFiles.Length > 1 ) {
-				// TODO: Log this
-				return null;
-			}
+				if( bmxConfigFiles.Length > 1 ) {
+					// TODO: Log this
```
boarnoah, line 85: Don't think we will do this?

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -75,23 +74,30 @@ internal class BmxConfigProvider : IBmxConfigProvider {
...
+				}
+
+				currentDirectory = currentDirectory.Parent;
 			}
 
-			currentDirectory = currentDirectory.Parent;
+			return null;
+		} catch( Exception ) {
+			// if there's any error finding a .bmx file, assume it doesn't exist rather than crash the app
+			// TODO: debug log?
```
boarnoah, line 99: ditto https://github.com/Brightspace/bmx/pull/326/files#r1219690309

boarnoah (approved):

cfbao:
> It would crash on linux before this?

The Go version would. Not the .NET version.
This is just to be extra sure, because we always look up the project config file now. It's still conceivable that some weird symlink setup can get the lookup into an infinite loop.

src/D2L.Bmx/BmxConfigProvider.cs
cfbao: Maybe?
I can conceive a scenario where a user might complain "why isn't .bmx config working for me", and we'll instruct them to enable some debug flag (like [this](https://github.com/Brightspace/bmx/pull/325/files#diff-707f49706d83b60f0cc24120487272507787e1c1144b484d1bdfb1d43eb21205R168-R169)) that'll turn on optional debug logging here and we can help them resolve issues more easily.

PR #330, opened 2023-06-12
use proper HTML parsing to extract SAMLResponse

cfbao:
more robust parsing

### DIFF
--- a/src/D2L.Bmx/HtmlXmlHelper.cs
+++ b/src/D2L.Bmx/HtmlXmlHelper.cs
@@ -1,17 +1,23 @@
-using System.Text.RegularExpressions;
+using System.Net;
 using System.Xml;
 using D2L.Bmx.Aws;
+using HtmlAgilityPack;
 
 namespace D2L.Bmx;
 
 internal static partial class HtmlXmlHelper {
 	public static string GetSamlResponseFromLoginPage( string html ) {
-		// TODO: use proper HTML parsing instead of regex
-		var inputXml = new XmlDocument();
-		inputXml.LoadXml( SamlResponseInputRegex().Match( html ).Value );
+		var htmlDoc = new HtmlDocument();
+		htmlDoc.LoadHtml( html );
 
-		return inputXml.SelectSingleNode( "//@value" )?.InnerText
-			?? throw new BmxException( "Error parsing AWS login page" );
+		if(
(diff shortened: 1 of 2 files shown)


src/D2L.Bmx/HtmlXmlHelper.cs
```diff
@@ -1,17 +1,23 @@
...
-		inputXml.LoadXml( SamlResponseInputRegex().Match( html ).Value );
+		var htmlDoc = new HtmlDocument();
+		htmlDoc.LoadHtml( html );
 
-		return inputXml.SelectSingleNode( "//@value" )?.InnerText
-			?? throw new BmxException( "Error parsing AWS login page" );
+		if(
+			// `SelectSingleNode` can return null even though its signature doesn't say so
+			htmlDoc.DocumentNode.SelectSingleNode( "//input[@name='SAMLResponse']" ) is HtmlNode inputNode
+			&& inputNode.GetAttributeValue( "value", /* default: */ null ) is string samlResponse
```
cfbao, line 16: I would've used named arguments, except that this name is `def`, which is confusing as hell...
In my brain "def" == "definition".

src/D2L.Bmx/HtmlXmlHelper.cs
```diff
@@ -1,17 +1,23 @@
...
-		inputXml.LoadXml( SamlResponseInputRegex().Match( html ).Value );
+		var htmlDoc = new HtmlDocument();
+		htmlDoc.LoadHtml( html );
 
-		return inputXml.SelectSingleNode( "//@value" )?.InnerText
-			?? throw new BmxException( "Error parsing AWS login page" );
+		if(
+			// `SelectSingleNode` can return null even though its signature doesn't say so
+			htmlDoc.DocumentNode.SelectSingleNode( "//input[@name='SAMLResponse']" ) is HtmlNode inputNode
+			&& inputNode.GetAttributeValue( "value", /* default */ null ) is string samlResponse
```
cfbao, line 16: I would've used named arguments, except that this name is `def`, which is confusing as hell...
In my brain "def" == "definition".

boarnoah:
How much of an impact does this have on the binary size?

It doesn't really give us much in terms of value since we only need to do the parsing in one place, with not much to do re: graceful handling if parsing goes wrong because the page's markup changed drastically etc...

boarnoah (approved):

PR #333, opened 2023-06-20
write: use ini parser to format credentials file

gord5500:
### Why

The `SharedCredentialsFile` has a good implementation for working with ini files but the output is that of `x=y`. The GO version of BMX had the output such that every `=` sign for each line was aligned. 

### How

Use a third party ini parser library that will let us update / create profiles. The libraries I looked through don't have this same feature for dotnet. 

I considered writing something custom to give us the same kind of output here. While I got that working, I was worried of edge cases where the time spent wouldn't be worth it. Last thing I want is to break someone's credentials file.

So the new output will be in the format `x = y` instead of `x=y`

### DIFF
--- a/src/D2L.Bmx/Program.cs
+++ b/src/D2L.Bmx/Program.cs
@@ -6,6 +6,7 @@ using Amazon.SecurityToken;
 using D2L.Bmx;
 using D2L.Bmx.Aws;
 using D2L.Bmx.Okta;
+using IniParser;
 
 var rootCommand = new RootCommand( "BMX grants you API access to your AWS accounts!" );
 
@@ -131,6 +132,7 @@ var writeCommand = new Command( "write", "Write to AWS credentials file" ) {
 writeCommand.SetHandler( ( InvocationContext context ) => RunWithErrorHandlingAsync( context, () => {
 	var consolePrompter = new ConsolePrompter();
 	var config = new BmxConfigProvider().GetConfiguration();
+	var parser = new FileIniDataParser();
 	var handler = new WriteHandler(
 		new OktaAuthenticator(
 			new OktaApi(),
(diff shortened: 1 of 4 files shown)


src/D2L.Bmx/WriteHandler.cs
```diff
@@ -18,6 +21,11 @@ BmxConfig config
 		_awsCreds = awsCreds;
 		_consolePrompter = consolePrompter;
 		_config = config;
+
+		var x = new IniParserConfiguration();
+
+
```
gord5500, line 27:
```suggestion
```

gord5500:
Oops forgot to open

src/D2L.Bmx/WriteHandler.cs
```diff
@@ -18,6 +21,7 @@ BmxConfig config
 		_awsCreds = awsCreds;
 		_consolePrompter = consolePrompter;
 		_config = config;
+		_parser = new FileIniDataParser();
```
cfbao, line 24: pleas pass it through the constructor parameter if only for consistency.

gord5500: Is here where you mean? https://github.com/Brightspace/bmx/blob/main/src/D2L.Bmx/Program.cs#L131

cfbao: yeah

cfbao: Also, I think we can reuse this ini parser in `BmxConfigProvider` and get rid of `Microsoft.Extensions.Configuration.Ini`.
There are some nice things we can do with a proper parser+writer.
Another time though.

gord5500: Thanks for clarification. I've moved it

src/D2L.Bmx/Program.cs
```diff
@@ -131,6 +132,7 @@
 writeCommand.SetHandler( ( InvocationContext context ) => RunWithErrorHandlingAsync( context, () => {
 	var consolePrompter = new ConsolePrompter();
 	var config = new BmxConfigProvider().GetConfiguration();
+	var parser = new FileIniDataParser();
```
gord5500, line 135: well maybe I should just put it directly in and not create a var. I was thinking maybe we want to configure it more before passing in but I don't think so now

cfbao (approved):

cfbao (approved):

PR #338, opened 2023-06-21
print: default output based on parent process name

gord5500:
### Why

We can do better about choosing the output than just checking the OS. There's a chance someone could be using bash on windows or powershell on mac

### How

Use Windows API and for linux / mac use some shell code to climb up the process tree and get the proper name


src/D2L.Bmx/PrintHandler.cs
```diff
@@ -25,10 +28,24 @@ AwsCredsCreator awsCredsCreator
...
-				PrintBash( awsCreds );
+			var current = Process.GetCurrentProcess();
+			var parent = ProcessCodeMethods.GetParentProcess( new PSObject( current ) ) as Process;
+			string? parentProcName = parent?.ProcessName;
+			switch( parentProcName ) {
+				case "pwsh":
+					PrintPowershell( awsCreds );
+					break;
+				case "bash":
+				case "zsh":
```
gord5500, line 39: This is terminal on mac. Was debating on leaving a comment

cfbao:
Looks like the binary size jumped to 30+MiB with this change...
I wish the PowerShell package would work for us, but I don't think it's worth the size bloat.

gord5500:
> Looks like the binary size jumped to 30+MiB with this change... I wish the PowerShell package would work for us, but I don't think it's worth the size bloat.

Is there any way to do some trimming like we did with AWS package? 

Is there another way you'd prefer to do this instead? I could see what the build size is calling windows / mac api but I think it's a lot harder to read

src/D2L.Bmx/UnixParentProcess.cs
```diff
@@ -0,0 +1,40 @@
+using System.Diagnostics;
+
+namespace D2L.Bmx;
+
+internal class UnixParentProcess {
+
+	public static string GetParentProcessName() {
+		string parentPid = GetParentProcessPid( GetParentProcessPid() );
```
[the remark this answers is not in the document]
gord5500: On mac at least. Need to try linux

src/D2L.Bmx/PrintHandler.cs
boarnoah: Don't think that is necessary, its a common enough shell even with Linux users which we want to support

src/D2L.Bmx/D2L.Bmx.csproj
```diff
@@ -8,6 +8,7 @@
     <InvariantGlobalization>true</InvariantGlobalization>
     <RootNamespace>D2L.Bmx</RootNamespace>
     <LangVersion>preview</LangVersion>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
gord5500, line 11: `NtQueryInformationProcess` needs this

src/D2L.Bmx/WindowsParentProcess.cs
```diff
@@ -0,0 +1,46 @@
...
+internal partial class WindowsParentProcess {
+
+	[DllImport( "ntdll.dll", EntryPoint = "NtQueryInformationProcess" )]
+	internal static extern int NtQueryInformationProcess(
+		IntPtr processHandle,
+		int processInformationClass,
+		ref PROCESS_BASIC_INFORMATION processInformation,
+		int processInformationLength,
+		out int returnLength
+	);
```
gord5500, line 15: This is part of internal windows api and the windows docs recommends something from user level of the api

src/D2L.Bmx/Program.cs
```diff
@@ -173,6 +173,7 @@
 		if( e is BmxException ) {
 			Console.Error.WriteLine( e.Message );
 		} else {
+			Console.WriteLine( e.ToString() );
```
gord5500, line 176:
```suggestion
```

src/D2L.Bmx/WindowsParentProcess.cs
gord5500: And I think that's only because they can change the structure or output in a future update and break the current setup

src/D2L.Bmx/D2L.Bmx.csproj
cfbao: It's really LibraryImport that needs this, but you aren't using LibraryImport yet. Any reason not to?

src/D2L.Bmx/D2L.Bmx.csproj
```diff
@@ -17,6 +18,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Ini" Version="8.0.0-preview.5.23280.8" />
     <PackageReference Include="Microsoft.Extensions.FileProviders.Physical" Version="8.0.0-preview.5.23280.8" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
+    <PackageReference Include="System.Management.Automation" Version="7.4.0-preview.3" />
```
cfbao, line 21: should remove this now

src/D2L.Bmx/D2L.Bmx.csproj
gord5500: Oh right. I was getting some errors when trying to build and it was from using library import. Let me copy them into a slack thread because I may need some help

src/D2L.Bmx/UnixParentProcess.cs
```diff
@@ -0,0 +1,47 @@
+using System.Diagnostics;
+
+namespace D2L.Bmx;
+
+internal class UnixParentProcess {
+
+	public static string GetParentProcessName() {
+		int parentPid = GetParentProcessId( GetParentProcessId( Process.GetCurrentProcess().Id ) );
```
cfbao, line 8: do you see an analyzer suggestion on this?
```suggestion
		int parentPid = GetParentProcessId( GetParentProcessId( Environment.ProcessId ) );
```
also - why two `GetParentProcessId` calls?

src/D2L.Bmx/UnixParentProcess.cs
```diff
@@ -0,0 +1,47 @@
...
+		var proccessStartInfo = new ProcessStartInfo {
+			FileName = "/bin/ps",
+			ArgumentList = { "-p", $"{parentPid}", "-o", "comm=" },
```
cfbao, line 11: given the id, you can get the name (and other info) of a process by using standard .NET APIs
see https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.getprocessbyid

src/D2L.Bmx/UnixParentProcess.cs
```diff
@@ -0,0 +1,47 @@
...
+				proc.WaitForExit();
+				return output;
+			}
+			return "bash";
+		}
+	}
+
+	private static int GetParentProcessId( int pid ) {
+		var proccessStartInfo = new ProcessStartInfo {
+			FileName = "/bin/ps",
```
cfbao, line 29: let the OS figure out that path here
```suggestion
			FileName = "ps",
```

src/D2L.Bmx/UnixParentProcess.cs
```diff
@@ -0,0 +1,47 @@
...
+			ArgumentList = { "-p", $"{pid}", "-o", "ppid=" },
+			RedirectStandardOutput = true,
+			UseShellExecute = false,
+			CreateNoWindow = true,
```
cfbao, line 33: `UseShellExecute` defaults to false
`CreateNoWindow` is ignored on non-Windows platforms
```suggestion
```

src/D2L.Bmx/UnixParentProcess.cs
```diff
@@ -0,0 +1,47 @@
...
+		};
+
+		using( var proc = Process.Start( proccessStartInfo ) ) {
+			if( proc is not null ) {
```
cfbao, line 37: nit:
* use [using declaration](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/using) to reduce nesting
* use early exit to reduce nesting (i.e. if null then return first)

src/D2L.Bmx/D2L.Bmx.csproj
gord5500: Sorry no I did manage to fix it but did not swap it back from DllImport to LibraryImport

src/D2L.Bmx/UnixParentProcess.cs
gord5500: vscode did not suggest that but I can change to what you suggested

gord5500: I think there could be a big tree of potential processes before getting to the shell if something else calls bmx so I'm not sure how rigid this will be unless we want to keep going up trying to find something. For example maybe aws credentials-process would break this and instead of bash being the 2nd parent its some aws process

cfbao: > The first parent is bmx and then we have to go up one more parent to get the shell

This doesn't sound right. I tested this branch in a PowerShell session launched by Bash, and it defaulted to Bash not PowerShell. I don't think the two get parent calls are necessary.

> I think there could be a big tree of potential processes before getting to the shell if something else calls bmx so I'm not sure how rigid this will be unless we want to keep going up trying to find something. For example maybe aws credentials-process would break this and instead of bash being the 2nd parent its some aws process

We only care if the user is launching bmx from within their shell directly. If there's an intermediate process, then the shell is irrelevant.

cfbao: Similarly, when I launched bash from PowerShell then launch bmx, it defaulted to PowerShell.

cfbao: To confirm - you're running the published native AOT build, not `dotnet run` or launching via IDE?

gord5500: Yep I'm using published build. My process is 
`dotnet publish osx-x64 -o out`
`$(./out/bmx print)`

src/D2L.Bmx/WindowsParentProcess.cs
```diff
@@ -0,0 +1,46 @@
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace D2L.Bmx;
+
+internal partial class WindowsParentProcess {
```
cfbao, line 6: would be nice to add a comment noting the PowerShell implementation that's referenced
https://github.com/PowerShell/PowerShell/blob/26f621952910e33840efb0c539fbef1e2a467a0d/src/System.Management.Automation/engine/ProcessCodeMethods.cs

src/D2L.Bmx/UnixParentProcess.cs
```diff
@@ -0,0 +1,31 @@
+using System.Diagnostics;
+
+namespace D2L.Bmx;
+
+internal class UnixParentProcess {
```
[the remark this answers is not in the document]
gord5500: Sure. I've done that and also referenced the powershell library

cfbao (approved):

cfbao (approved):

PR #346, opened 2023-06-22
check binary sizes in CI

cfbao:
easier to know when a change introduces significant bloat to the binary


boarnoah (approved):
Good idea, not super visible, I guess we can remember to check the build size on PRs that add new dependencies etc...

Blocking action with an error once its past a certain size is likely overkill?

gord5500:
> Good idea, not super visible, I guess we can remember to check the build size on PRs that add new dependencies etc...
> 
> Blocking action with an error once its past a certain size is likely overkill?

Could it just make a comment saying the size? Seems useful to know maybe not as useful as the bot that posts terraform plans

PR #358, opened 2023-06-26
Noninteractive configure

jharoutuniand2l:
### Why

Smooth out the install process a bit. 
https://trello.com/c/kKIKoUrM/1271-add-non-interactive-flag-to-bmx-configure

### DIFF
--- a/src/D2L.Bmx/ConfigureHandler.cs
+++ b/src/D2L.Bmx/ConfigureHandler.cs
@@ -16,8 +16,8 @@ internal class ConfigureHandler( IBmxConfigProvider configProvider, IConsoleProm
 			user = consolePrompter.PromptUser();
 		}
 
-		if( duration is null ) {
-			duration = nonInteractive ? 60 : consolePrompter.PromptDuration();
+		if( duration is null && !nonInteractive ) {
+			duration = consolePrompter.PromptDuration();
 		}
 
 		BmxConfig config = new(
--- a/src/D2L.Bmx/ConfigureHandler.cs
+++ b/src/D2L.Bmx/ConfigureHandler.cs
@@ -4,19 +4,26 @@ internal class ConfigureHandler( IBmxConfigProvider configProvider, IConsoleProm
 	public void Handle(
 		string? org,
 		string? user,
-		int? duration
+		int? duration,
+		bool nonInteractive
 	) {
 
 		if( string.IsNullOrEmpty( org ) ) {
+			if( nonInteractive ) {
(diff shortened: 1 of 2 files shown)


src/D2L.Bmx/ConfigureHandler.cs
```diff
@@ -4,19 +4,26 @@ internal class ConfigureHandler( IBmxConfigProvider configProvider, IConsoleProm
 	public void Handle(
 		string? org,
 		string? user,
-		int? duration
+		int? duration,
+		bool nonInteractive
 	) {
 
 		if( string.IsNullOrEmpty( org ) ) {
+			if( nonInteractive ) {
```
jharoutuniand2l, line 12: Note: in other similar areas we use `if( !nonInteractive ) {` conditions to control flow but:
* I don't like double negatives in code when they can be avoided
* I think this is clearer anyways

Happy to switch it back (or update the other spots)

cfbao: I like this style better in general.

That said, I had a different design in mind when I made that card:
If `--non-interactive` is specified and a config property isn't specified on the command line, it will simply be omitted from the config file.

So I'd want this instead:
```csharp
if( string.IsNullOrEmpty( org ) && !nonInteractive ) {
	org = consolePrompter.PromptOrg();
}
```
No error.
`BmxConfigProvider.SaveConfiguration` will skip writing any properties that have a null/default/empty value.

Does this sound reasonable?

jharoutuniand2l: Yup - makes sense and sounds good.
I guess the other commands literally don't work without org/user, whereas this one can still do part of the work.

src/D2L.Bmx/ConfigureHandler.cs
```diff
@@ -4,19 +4,20 @@ internal class ConfigureHandler( IBmxConfigProvider configProvider, IConsoleProm
...
 		}
 
-		if( string.IsNullOrEmpty( user ) ) {
+		if( string.IsNullOrEmpty( user ) && !nonInteractive ) {
 			user = consolePrompter.PromptUser();
 		}
 
 		if( duration is null ) {
-			duration = consolePrompter.PromptDuration();
+			duration = nonInteractive ? 60 : consolePrompter.PromptDuration();
```
cfbao, line 20: Similarly here
```csharp
if( duration is null && !nonInteractive ) {
	duration = consolePrompter.PromptDuration();
}
```

cfbao (approved):

PR #359, opened 2023-06-27
Add bmx login

gord5500:
### Why

We want functionality to create and cache an Okta session without a user having to use print or write. 

### How

We simply call existing code that goes through the login process


src/D2L.Bmx/LoginHandler.cs
```diff
@@ -0,0 +1,13 @@
...
+
+internal class LoginHandler(
+	OktaAuthenticator oktaAuth
+) {
+	public async Task HandleAsync(
+		string? org,
+		string? user
+	) {
+		await oktaAuth.AuthenticateAsync( org, user, nonInteractive: false );
+		Console.WriteLine( "Successfully logged in." );
```
gord5500, line 11: If there's an error in okta auth it will show the error message instead of ever reaching this line

gord5500:
Think my local main branch was outdated? That's a lot of commits

src/D2L.Bmx/LoginHandler.cs
```diff
@@ -0,0 +1,13 @@
+namespace D2L.Bmx;
+
+internal class LoginHandler(
+	OktaAuthenticator oktaAuth
+) {
+	public async Task HandleAsync(
+		string? org,
+		string? user
+	) {
+		await oktaAuth.AuthenticateAsync( org, user, nonInteractive: false );
```
cfbao, line 10: We should check if the config file exists first. If not, Okta session won't be saved, and there's no point in running login. We should exit with error and warn the user about this.

gord5500: Good idea

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -97,6 +98,8 @@ IOktaApi oktaApi
 	private void CacheOktaSession( string userId, string org, string sessionId, DateTimeOffset expiresAt ) {
 		var session = new OktaSessionCache( userId, org, sessionId, expiresAt );
 		var existingSessions = ReadOktaSessionCacheFile();
+		existingSessions = existingSessions.Where( session => session.UserId != userId && session.Org != org )
```
cfbao, line 101: naming nit (it's not "existing sessions" anymore):
```suggestion
		var sessionsToCache = existingSessions.Where( session => session.UserId != userId && session.Org != org )
```
also at [L105](https://github.com/Brightspace/bmx/pull/359/commits/67b27fb22b31d3104ac9d7bbdfa59bb4d3aec04e#diff-640013b3685b8a2cbcee8c5b2e26b1afc631ea22dc293c44a6f402feab21ff73R105)

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -12,7 +12,8 @@ BmxConfig config
 	public async Task<IOktaApi> AuthenticateAsync(
 		string? org,
 		string? user,
-		bool nonInteractive
+		bool nonInteractive,
+		bool ignoreCache = false
```
cfbao, line 16: there's a default here, but you also explicitly specified the default value in print & write handler.
Either remove this default value, or actually utilize it in the other two handlers.

cfbao:
> Think my local main branch was outdated? That's a lot of commits

Not exactly "outdated". Your branch is based on your parent process lookup branch, not the main branch.
You wouldn't get this many commits in the PR if you based it on the trunk of the main branch, even if somewhat outdated.

src/D2L.Bmx/OktaAuthenticator.cs
gord5500: Ah yea. I've removed the default value since none of the other parameters use it

PR #360, opened 2023-06-29
configure: prevent data stomping

gord5500:
### Why

Currently configure command will overwrite all existing settings in the config file

### How

We can use an ini parser the same way we do for `bmx write` and not do a complete rewrite of the file. This requires us to change some of the console prompter code as it currently will throw an exception if no input is given

Related trello card [here](https://trello.com/c/VFHzMaY4/1295-configure-command-should-keep-existing-entries-in-the-config-file)

### DIFF
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -54,13 +54,15 @@ internal class BmxConfigProvider : IBmxConfigProvider {
 			Directory.CreateDirectory( BmxPaths.BMX_DIR );
 		}
 		var op = new FileStreamOptions {
-			Mode = FileMode.Create,
-			Access = FileAccess.Write,
+			Mode = FileMode.OpenOrCreate,
+			Access = FileAccess.ReadWrite,
 		};
 		if( !RuntimeInformation.IsOSPlatform( OSPlatform.Windows ) ) {
 			op.UnixCreateMode = UnixFileMode.UserRead | UnixFileMode.UserWrite;
 		}
-		var data = parser.ReadFile( BmxPaths.CONFIG_FILE_NAME );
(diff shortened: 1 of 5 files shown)


src/D2L.Bmx/Program.cs
```diff
@@ -199,6 +200,7 @@
 			Console.Error.WriteLine( e.Message );
 		} else {
 			Console.Error.WriteLine( "BMX exited with unexpected internal error" );
+			Console.Error.WriteLine( e.ToString() );
```
gord5500, line 203:
```suggestion
```

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -48,28 +49,36 @@ internal class BmxConfigProvider : IBmxConfigProvider {
 		);
 	}
 
-	public void SaveConfiguration( BmxConfig config ) {
+	public void SaveConfiguration( BmxConfig config, FileIniDataParser parser ) {
```
cfbao, line 52: injecting `FileIniDataParser` via the constructor rather than a method parameter would be more consistent with our general style.

cfbao (approved):

src/D2L.Bmx/BmxConfigProvider.cs
gord5500: I was debating on this and what we'd prefer but thanks I'll swap it it to be from constructor

cfbao: it'd be especially handy once we replace `Microsoft.Extensions.Configuration.Ini` with the ini-parser entirely, in the `GetConfiguration` method too.
https://trello.com/c/bwS24Mo4/1238-replace-microsoftextensionsconfigurationini-with-ini-parser-netstandard

cfbao (approved):

PR #368, opened 2023-07-20
CI: include CPU arch in release assets names

cfbao:
### Why

In the future we'll likely want to add ARM builds, and it'd be good to not break the naming pattern (and existing automated processes) when we do.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -112,12 +112,12 @@ jobs:
         working-directory: src/D2L.Bmx/build
         run: |
           if ($env:PLATFORM -eq 'win') {
-            Compress-Archive -Path "bmx.exe" -DestinationPath "bmx-win.zip"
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -112,12 +112,12 @@ jobs:
         working-directory: src/D2L.Bmx/build
         run: |
           if ($env:PLATFORM -eq 'win') {
-            Compress-Archive -Path "bmx.exe" -DestinationPath "bmx-win.zip"
-            gh release upload "$env:RELEASE_TAG" "bmx-win.zip"
+            Compress-Archive -Path "bmx.exe" -DestinationPath "bmx-win-x64.zip"
```
cfbao, line 115: decided against x86-64 because the hyphen can be confusing.
also .NET uses x64 in RIDs.

gord5500 (approved):

jharoutuniand2l (approved):

PR #369, opened 2023-07-20
ci: docker builds set their build version

gord5500:
### Why

We forgot to do this for the docker builds

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -105,7 +105,7 @@ jobs:
         if: github.event_name == 'workflow_dispatch'
         shell: pwsh
         env:
-          RELEASE_TAG: ${{ needs.create_release.outputs.release_id }}
+          RELEASE_ID: ${{ needs.create_release.outputs.release_id }}
           PLATFORM: ${{ matrix.platform }}
           FILE_NAME: ${{ matrix.file_name }}
           GH_TOKEN: ${{ github.token }}
@@ -113,11 +113,11 @@ jobs:
         run: |
           if ($env:PLATFORM -eq 'win') {
             Compress-Archive -Path "bmx.exe" -DestinationPath "bmx-win.zip"
-            gh release upload "$env:RELEASE_TAG" "bmx-win.zip"
+            gh release upload "$env:RELEASE_ID" "bmx-win.zip"
           } elseif ($env:PLATFORM -eq 'osx') {
             chmod +x "bmx"
             tar -czvf "bmx-osx.tar.gz" "bmx"
-            gh release upload "$env:RELEASE_TAG" "bmx-osx.tar.gz"
+            gh release upload "$env:RELEASE_ID" "bmx-osx.tar.gz"
           }
 
   build_docker:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -105,19 +105,19 @@ jobs:
...
+          RELEASE_ID: ${{ needs.create_release.outputs.release_id }}
           PLATFORM: ${{ matrix.platform }}
           FILE_NAME: ${{ matrix.file_name }}
           GH_TOKEN: ${{ github.token }}
         working-directory: src/D2L.Bmx/build
         run: |
           if ($env:PLATFORM -eq 'win') {
             Compress-Archive -Path "bmx.exe" -DestinationPath "bmx-win-x64.zip"
-            gh release upload "$env:RELEASE_TAG" "bmx-win-x64.zip"
+            gh release upload "$env:RELEASE_ID" "bmx-win.zip"
```
cfbao, line 116: please sort out all the `-x64` changes

Dockerfile.alpine
```diff
@@ -1,5 +1,8 @@
 FROM amd64/alpine:3.15 AS build
 
+ARG version
+ARG information_version
```
cfbao, line 4: Thinking of local builds - what happens when we don't provide these arguments on the command line when invoking docker build? Should we set default values for them?

Dockerfile.alpine
```diff
@@ -25,7 +28,13 @@ COPY src/D2L.Bmx/D2L.Bmx.csproj src/D2L.Bmx/D2L.Bmx.csproj
 RUN dotnet restore src/D2L.Bmx -r linux-musl-x64
 
 COPY . .
-RUN dotnet publish src/D2L.Bmx --no-restore -r linux-musl-x64 -o app
+RUN dotnet publish src/D2L.Bmx \
+    -p:Version=$version \
+    -p:InformationVersion=$information_version \
```
cfbao, line 33: variable references should be quoted in shell scripts as a best practice
```suggestion
    -p:Version="$version" \
    -p:InformationVersion="$information_version" \
```

Dockerfile.alpine
gord5500: Yea I can set them to default to 3.0.0 and v3.0.0

.github/workflows/publish.yml
```diff
@@ -152,23 +162,23 @@ jobs:
...
-          RELEASE_TAG: ${{ needs.create_release.outputs.release_id }}
+          RELEASE_ID: ${{ needs.create_release.outputs.release_id }}
           PLATFORM: ${{ matrix.platform }}
           GH_TOKEN: ${{ github.token }}
         working-directory: build
         run: |
           chmod +x "bmx"
-          tar -czvf "bmx-$PLATFORM-x64.tar.gz" "bmx"
-          gh release upload "$RELEASE_TAG" "bmx-$PLATFORM-x64.tar.gz"
+          tar -czvf "bmx-$PLATFORM.tar.gz" "bmx"
```
cfbao, line 171:
```suggestion
          tar -czvf "bmx-$PLATFORM-x64.tar.gz" "bmx"
```

gord5500: A sorry 😢

cfbao (approved):

PR #373, opened 2023-07-28
write: create output file path / file if doesn't exist

gord5500:
### Why

We get an error if the .aws folder or credentials file doesn't exist

### DIFF
--- a/src/D2L.Bmx/WriteHandler.cs
+++ b/src/D2L.Bmx/WriteHandler.cs
@@ -35,10 +35,10 @@ internal class WriteHandler(
 		if( !string.IsNullOrEmpty( output ) && !Path.IsPathRooted( output ) ) {
 			output = "./" + output;
 		}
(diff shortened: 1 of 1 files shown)


src/D2L.Bmx/WriteHandler.cs
```diff
@@ -36,15 +36,23 @@ FileIniDataParser parser
 			output = "./" + output;
 		}
 		var credentialsFile = new SharedCredentialsFile( output );
-		var data = parser.ReadFile( credentialsFile.FilePath );
+		string credentialsFilePath = credentialsFile.FilePath;
```
cfbao, line 39: nit: nice to get rid of the `credentialsFile` variable if it's only used once to get the path
```csharp
string credentialsFilePath = new SharedCredentialsFile( output ).FilePath;
```

cfbao (approved):

PR #377, opened 2023-08-01
Use ini-parser-netstandard for all ini file parsing

DannyDaoBoYang:
### Why
IniParser and Microsoft.Extensions.Configuration.Ini have roughly the same functionalities. Using Iniparser only to make code more consistent

### DIFF
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -25,7 +25,7 @@ internal class BmxConfigProvider( FileIniDataParser parser ) : IBmxConfigProvide
 		FileInfo? projectConfigInfo = GetProjectConfigFileInfo();
 
 		if( projectConfigInfo is not null ) {
-			var configFilePath = Path.Combine( projectConfigInfo.DirectoryName ?? "", projectConfigInfo.Name );
+			string configFilePath = Path.Combine( projectConfigInfo.DirectoryName ?? "", projectConfigInfo.Name );
 			try {
 				var tempdata = parser.ReadFile( configFilePath );
 				data.Merge( tempdata );
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -12,7 +12,7 @@ internal class BmxConfigProvider( FileIniDataParser parser ) : IBmxConfigProvide
 	public BmxConfig GetConfiguration() {
 		// Main config is at ~/.bmx/config
 		string configFileName = BmxPaths.CONFIG_FILE_NAME;
-		IniData data = new IniData();
+		var data = new IniData();
 		if( File.Exists( configFileName ) ) {
 			var tempdata = parser.ReadFile( configFileName );
 			data.Merge( tempdata );
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -1,8 +1,3 @@
-using System.Runtime.InteropServices;
-using IniParser;
-using Microsoft.Extensions.FileProviders;
-using Microsoft.Extensions.FileProviders.Physical;
-
 namespace D2L.Bmx;
 
 internal interface IBmxConfigProvider {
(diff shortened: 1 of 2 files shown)


src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -15,36 +12,42 @@ internal class BmxConfigProvider( FileIniDataParser parser ) : IBmxConfigProvide
...
 		string configFileName = BmxPaths.CONFIG_FILE_NAME;
-
+		var data = new IniData();
+		if( File.Exists( configFileName ) ) {
+			try {
+				var tempdata = parser.ReadFile( configFileName );
+				data.Merge( tempdata );
+			} catch( IniParser.Exceptions.ParsingException ) {
+				//No read permission?
+			}
```
cfbao, line 22: I'm curious what permission error you saw.
Also wondering if it makes sense to just catch all errors here and print a warning to the user

cfbao:
package references to Microsoft.Extensions.Configuration.Ini and Microsoft.Extensions.FileProviders.Physical should both be removed

cfbao:
on PR title - the BMX repo doesn't really have different "components", and what you're changing is not an "API", so the prefix doesn't make sense here.

I'd suggest:
> Use ini-parser-netstandard for all ini file parsing

src/D2L.Bmx/BmxConfigProvider.cs
DannyDaoBoYang: Yes, catch all errors here make more sense.

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -15,36 +12,42 @@ internal class BmxConfigProvider( FileIniDataParser parser ) : IBmxConfigProvide
...
 		// Main config is at ~/.bmx/config
 		string configFileName = BmxPaths.CONFIG_FILE_NAME;
-
+		var data = new IniData();
+		if( File.Exists( configFileName ) ) {
+			try {
+				var tempdata = parser.ReadFile( configFileName );
+				data.Merge( tempdata );
+			} catch( Exception ) {
+				Console.Write( "Error occur while parsing config file." );
```
cfbao, line 21: * We rarely want to write to stdout, because that's reserved for `bmx print` to write the env var setting commands.
* nit: grammar and wording - I'm trying to make it clearer to the user that their config file is entirely ignored in this run. I'm a little worried that "error occurred" could be interpreted as "something wasn't quite right but it's mostly fine". Including the config file path is nice for troubleshooting
```suggestion
				Console.Error.Write( $"WARNING: Failed to load the global config file {configFileName}." );
```

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -15,36 +12,42 @@ internal class BmxConfigProvider( FileIniDataParser parser ) : IBmxConfigProvide
...
 
 		if( projectConfigInfo is not null ) {
-			// Default file provider ignores files prefixed with ".", we need to provide our own as a result
-			var fileProvider =
-				new PhysicalFileProvider( projectConfigInfo.DirectoryName!, ExclusionFilters.None );
-			configBuilder.AddIniFile( fileProvider, projectConfigInfo.Name, optional: false,
-				reloadOnChange: false );
+			string configFilePath = Path.Combine( projectConfigInfo.DirectoryName ?? "", projectConfigInfo.Name );
+			try {
+				var tempdata = parser.ReadFile( configFilePath );
```
cfbao, line 30: we don't need to reconstruct the file path
```suggestion
			try {
				var tempdata = parser.ReadFile( projectConfigInfo.FullName );
```

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -15,36 +12,42 @@ internal class BmxConfigProvider( FileIniDataParser parser ) : IBmxConfigProvide
...
+				data.Merge( tempdata );
+			} catch( Exception ) {
+				Console.Write( "Error occur while parsing additional config file." );
```
cfbao, line 33:
```suggestion
				Console.Error.Write( $"WARNING: Failed to load the local config file {projectConfigInfo.FullName}." );
```

cfbao (approved):

PR #384, opened 2023-08-24
AWS Credentials: Add cache

DannyDaoBoYang:
### Why
Cache the aws auth token, so we don't have to send request every single time


cfbao:
I'll skip reviewing the code itself, but a few comments on the design:

* This functionality should be gated behind a flag, at least on the read side. Users (at least me) probably expect they'd get fresh credentials each time now. Making them not fresh by default is a big breaking change and bad UX IMO. We could potentially always write to the cache regardless of any flag, but I don't feel strongly either way on this.

* The storage format should be JSON, not ini. Ini files are human friendly, so they're good as config files, but there isn't a strict standard for ini files and the parser we use is kinda sketchy too. JSON is much stricter and the serializer is great.

* Definitely don't refresh until the last minute. I wonder if this should be a flag & config too. Or maybe just a single flag like `--no-refresh-until 15min` and this both enables reading from the cache and sets when we want a refresh. (don't actually use this flag name, it's terrible)

boarnoah:
> Definitely don't refresh until the last minute.

Should we concern ourselves with refreshing from cache at all?
The primary use case for the cache is for with use with the external credential provider feature.

The [documentation](https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html#feature-process-credentials-output) for which states:

> The Expiration key is an RFC3339 formatted timestamp. [...] are considered temporary credentials, and they are automatically refreshed by rerunning the credential_process command before the credentials expire.

How I understand this working is the SDK side which consumes credentials from the profile will handle invoking the process again based on the timeout you set within your SDK use.

Adding our own flag to use in the profile for handling invalidating cache ourselves would be redundant for this particular use case.

Maybe useful for when using the AWS CLI or with code we don't want to customize refresh interval with etc..., which seems kind of niche.

cfbao:
I'm confused... on one hand the doc says

> they are automatically refreshed by rerunning the credential_process command before the credentials expire.

but also

> The SDK does not cache external process credentials the way it does assume-role credentials. If caching is required, you must implement it in the external process.

if the SDK doesn't cache the creds, then why does it even care about the expiration time? shouldn't it just call the process every time?

src/D2L.Bmx/Program.cs
```diff
@@ -102,6 +102,9 @@
 		result.ErrorMessage = $"Unsupported value for --output. Must be one of:\n{string.Join( '\n', PrintFormat.All )}";
 	}
 } );
+var cacheOption = new Option<int?>(
```
gord5500, line 105: Why int?

src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -1,30 +1,52 @@
...
 			PrincipalArn = role.PrincipalArn,
 			RoleArn = role.RoleArn,
 			SAMLAssertion = samlAssertion,
 			DurationSeconds = durationInMinutes * 60,
 		} );
-
-		return new AwsCredentials(
+		// What about duration?
+		//cache the result
+		var credentails = new AwsCredentials(
```
gord5500, line 43:
```suggestion
		var credentials = new AwsCredentials(
```

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,64 @@
...
+internal class AwsCredsCache() {
+	private static readonly JsonSerializerOptions _options =
+		new() { DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull };
+	public void SaveToFile( string Org, string User, AwsRole role, AwsCredentials credentials ) {
+		if( !Directory.Exists( BmxPaths.BMX_DIR ) ) {
+			return;
+		}
+		if( Org == "" || User == "" ) { //It's not set in config file
+			return;
+		}
```
gord5500, line 17: What's this for? To not save if we don't have a config file?

DannyDaoBoYang: I want both Org and User already set in the config file. So I can label each Aws session token as which user who it belongs to.

src/D2L.Bmx/Program.cs
DannyDaoBoYang: We want to specify how much time is still left in the token. We don't want to give a token that expires the next second.

gord5500: Okay I see where it's used. From the name I would've thought it would be boolean and not expect to have to give a number here

src/D2L.Bmx/Program.cs
```diff
@@ -102,6 +102,9 @@
...
+	name: "--use-cache",
+	description: ParameterDescriptions.NonInteractive );
```
gord5500, line 107: Should make a description for this

src/D2L.Bmx/ParameterDescriptions.cs
```diff
@@ -11,4 +11,7 @@ internal static class ParameterDescriptions {
 	public const string Output = "Custom path to the AWS credentials file";
 	public const string Format = "Output format of AWS credentials";
 	public const string NonInteractive = "Run non-interactively without showing any prompts";
+
+	public const string UseCache = "Enables Cache and specify time until the expiration " +
```
gord5500, line 15: Maybe mention this is for AWS

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,65 @@
...
+
+internal class AwsCredsCache() {
+	private static readonly JsonSerializerOptions _options =
+		new() { DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull };
+
+	public void SaveToFile( string Org, string User, AwsRole role, AwsCredentials credentials ) {
+		if( !Directory.Exists( BmxPaths.BMX_DIR ) ) {
+			return;
+		}
+		if( Org == "" || User == "" ) { //It's not set in config file
```
gord5500, line 14: You can use string.IsNullOrEmpty instead of `== ""` and we do that in other places already 

https://github.com/Brightspace/bmx/pull/384/files#diff-75585c62d6f9197bcc885d332bb0e7a1d207d5e2b9389a4fa04107c92d9289a6R23

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,65 @@
...
+			string sessionsJson = File.ReadAllText( BmxPaths.AWS_CREDS_CACHE_FILE_NAME );
+			return JsonSerializer.Deserialize( sessionsJson, SourceGenerationContext.Default.ListAwsCacheModel )
+				?? new();
+		} catch( JsonException ) {
+			return new();
+		}
+	}
+
+	public AwsCredentials? GetCachedSession( string Org, string User, AwsRole role, int Cache ) {
+		// Main config is at ~/.bmx/config
```
gord5500, line 56: What's this here for?

DannyDaoBoYang: Get a specified session where Org, User, and role all matches, and expire time is greater than now + Cache

src/D2L.Bmx/ParameterDescriptions.cs
gord5500: The name use-cache doesn't immediately tell me that I need to provide a number with that so would be good to make sure we say what the number is doing

src/D2L.Bmx/Aws/AwsCredsCache.cs
gord5500: The comment I mean sorry 
> // Main config is at ~/.bmx/config

gord5500: Ah okay yea can we remove it then if we're not using that config file so no need to reference this
```suggestion
```

src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -62,6 +64,12 @@ bool nonInteractive
...
 		}
 
-		return await awsClient.GetTokensAsync( samlResponse, selectedRoleData, duration.Value );
+		return await awsClient.GetTokensAsync(
+			samlResponse,
+			selectedRoleData,
+			duration.Value,
+			useCache,
+			config.Org ?? "",
+			config.User ?? "" );
```
gord5500, line 73: If you use string.IsNullOrEmpty in the above suggestion then you can just pass in config.Org config.User

gord5500 (approved):

boarnoah:
From discussion  w/ @cfbao @gord5500 @DannyDaoBoYang 

BMX will determine when to invalidate cache (not user configurable)

| Requested Duration | Invalidate cache when TTL is less than |
|--------------------|----------------------------------------|
| 15 mins            | 10 mins                                |
| 16 mins            | 11 mins                                |
| 20 mins            | 15 mins                                |
| > 20 mins          | 15 mins                                |

Within 15 mins (min assume role credential duration) and 20, we will use a sliding scale to determine when to invalidate credentials

This is because otherwise if you say, request credentials valid for 15 mins you will always blow through cache
\===
Cache key should include username & org, in addition to role name
Never use cached credentials if they aren't for your user
\===
Error when attempting to use cache without a global config file
We use the absence of the global config file in order to control whether BMX stores any credential data on the machine or not (i.e Okta session, or now the AWS Credentials Cache)

So in a shared machine (without the global config file) we shouldn't be caching someone else's credentials to disk)

src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -62,6 +70,32 @@ bool nonInteractive
 			}
 		}
 
-		return await awsClient.GetTokensAsync( samlResponse, selectedRoleData, duration.Value );
+		if( cache &&
+			awsCredentialCache.GetCredentials(
```
boarnoah, line 74: Moved the AWS credential caching up ``AwsCredsCreator`` as an injectable service. 
1. The caching we are adding here is an implementation detail of BMX, doesn't have anything to do with AWS API access
1. Okta credential caching similarly lives a layer up from Okta API code

src/D2L.Bmx/Program.cs
```diff
@@ -185,7 +193,8 @@
 		duration: context.ParseResult.GetValueForOption( durationOption ),
 		nonInteractive: context.ParseResult.GetValueForOption( nonInteractiveOption ),
 		output: context.ParseResult.GetValueForOption( outputOption ),
-		profile: context.ParseResult.GetValueForOption( profileOption )
+		profile: context.ParseResult.GetValueForOption( profileOption ),
+		cacheAwsCredentials: context.ParseResult.GetValueForOption( cacheAwsCredentialsOption )
```
boarnoah, line 197: Renamed this flag outside the AWS namespace to make it clear exactly what we are caching when this is set. The CLI option still remains ``--cache``

src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -1,22 +1,30 @@
...
 	) {
-		OktaApp[] awsApps = await oktaApi.GetAwsAccountAppsAsync();
+		if( cache && !File.Exists( BmxPaths.CONFIG_FILE_NAME ) ) {
+			throw new BmxException(
+				"BMX global config file not found. Will not cache credentials on shared use machines."
+			);
+		}
+
+		//cache this one?
+		OktaApp[] awsApps = await oktaApi.Api.GetAwsAccountAppsAsync();
```
boarnoah, line 27: Good question, I would think we should be doing this, out of scope of this PR.

Realistically I feel like we are more at risk of hitting Okta limits (not necessarily rate limits but rather any DoS/Brute forcing prevention heuristics) than AWS STS limits.

In fact that's the only limit I recall someone (Mark Tse) hitting while using BMX in a script

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,83 @@
...
+			.Where( o =>
+				o.Credentials.Expiration >= ttlLongerThan
+				&& o.User == user
+				&& o.Org == org
+			)
+			// Prune older (closer to expiry) credentials for the same role
+			.GroupBy( o => o.RoleArn, ( roleArn, entries ) => new {
+				Key = roleArn,
+				Value = entries.OrderBy( o => o.Credentials.Expiration )
+			} )
```
boarnoah, line 59: ex: 
```
bmx print --account <account> --role <role> --cache --duration 15
print --account <account>  --role <role> --cache --duration 20
```
Credentials are fungible since they aren't assumed with a specific session name (don't think its possible when AssumeRoleWithSaml), so might as well keep only the longest living one.

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,83 @@
...
+		allEntries.Add( new AwsCacheModel(
+			Org: org,
+			User: user,
+			RoleArn: role.RoleArn,
+			Credentials: credentials
+		) );
+
+		var ttlLongerThan = DateTime.UtcNow.AddMinutes( 10 );
+
+		var prunedEntries = allEntries
```
boarnoah, line 48: Thought about pruning this cache on any BMX use. Its not the greatest look to leave around credentials (granted expired ones) around on someone's home folder.

However without some guards re: "does BMX have write access to write to a .bmx/" etc... it doesn't seem like a good idea to just do that? Especially since unlike the update check process we can't fail this silently.

cfbao: If they're using bmx, they probably also have `~/.aws/credentials` with some expired creds. Sooo... not a big deal I think.

gord5500 (approved):

src/D2L.Bmx/AwsCredsCreator.cs
cfbao: Pretty sure we don't want to hit Okta when we have cached creds.
Don't think we need to cache this though. Can we look up the AWS creds cache first and return early?

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,83 @@
...
+			// Remove expired/expiring & entries not related to current user
+			.Where( o =>
+				o.Credentials.Expiration >= ttlLongerThan
```
cfbao, line 51: nit: personally would prefer wording like `minimumTtl` or `minimumExpiryTime`. I had trouble understanding what `ttlLongerThan` meant what I read the definition above.

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,83 @@
...
+				&& o.User == user
+				&& o.Org == org
+			)
+			// Prune older (closer to expiry) credentials for the same role
+			.GroupBy( o => o.RoleArn, ( roleArn, entries ) => new {
+				Key = roleArn,
+				Value = entries.OrderBy( o => o.Credentials.Expiration )
```
cfbao, line 58: do we need this anonymously typed object here? do we need this `Key` at all?

src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -1,22 +1,30 @@
...
 		string? role,
 		int? duration,
-		bool nonInteractive
+		bool nonInteractive,
+		bool cache
 	) {
-		OktaApp[] awsApps = await oktaApi.GetAwsAccountAppsAsync();
+		if( cache && !File.Exists( BmxPaths.CONFIG_FILE_NAME ) ) {
+			throw new BmxException(
+				"BMX global config file not found. Will not cache credentials on shared use machines."
```
cfbao, line 22: would be nice to also suggest what to do. e.g.

> If you own this machine, run bmx configure.

src/D2L.Bmx/AwsCredsCreator.cs
boarnoah: Hmm, good point. I think so? We can move the return early w/ cache hit here and proceed to the querying the Okta API + prompt user to select a role (if they aren't using ``--non-interactive``)  & there is no match for the given ``--role`` etc..

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,103 @@
...
+
+namespace D2L.Bmx.Aws;
+
+internal interface IAwsCredentialCache {
+	public AwsCredentials? GetCredentials( string org, string user, string roleName, int duration );
+	public void SetCredentials( string org, string user, AwsRole role, AwsCredentials credentials );
+}
+
+internal class AwsCredsCache : IAwsCredentialCache {
+	private static void WriteTextToFile( string path, string content ) {
```
cfbao, line 13: nit: please put private methods after public ones

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,103 @@
...
+		);
+
+		EnsureUniqueRoleNames( matchedEntries );
+
+		return matchedEntries.FirstOrDefault()?.Credentials;
+	}
+
+	// Since we retrieve credentials without talking to APIs we are using RoleName as a cache key,
+	// Since we do not have account ID we are at risk of potentially retrieving credentials for the
+	// the wrong role (since role names aren't globally unique), guard against this
```
cfbao, line 97: should/can we cache by (normalized) account name + role name, and relax this restriction?

("normalized" because we don't require case sensitivity on account name)

boarnoah: I was writing an explanation for this and realised, if we put accountName in cache key, then we can relax this requirement again. Since at least in theory the alias has to be unique?

Otherwise selecting account via BMX would not work correctly (since we match account alias to Okta App label).

cfbao: it's actually the "app name" in Okta and has nothing to do with account alias. but anyway, yes it should be a unique combination within an org if bmx works at all.

boarnoah: I opted to call the additional cache key "accountName" at least throughout the rest of BMX we call it by account name when in practice it relates to name in Okta which in theory is unrelated to AWS Account alias (would make for a very confusing Okta App chiclet at least as far as D2Ler's are used to).

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,97 @@
...
+		string user,
+		string accountName,
+		string roleName,
+		int duration
+	) {
+		List<AwsCacheModel> allEntries = GetAllCache();
+
+		// Too stale, within a window (of 10 mins) if requested duration was <= 20 mins, 15 mins otherwise
+		// Avoids over-eagerly invalidating cache when requesting AWS credentials for short durations
+		var ttlLongerThan = DateTime.UtcNow.AddMinutes( duration > 20 ? 15 : duration - 5 );
```
cfbao, line 86: similar naming change?

src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -0,0 +1,97 @@
...
+		var matchedEntry = allEntries.Find(
+			o => o.Org == org
+			&& o.User == user
+			&& o.AccountName == accountName
```
cfbao, line 90: case insensitive compare?

src/D2L.Bmx/Aws/AwsCacheModel.cs
```diff
@@ -0,0 +1,8 @@
+namespace D2L.Bmx.Aws;
+internal record AwsCacheModel(
+	string Org,
+	string User,
+	string AccountName,
+	string RoleArn,
```
cfbao, line 6: actually I wonder if it makes more sense to just save the role name here. the rest of the role arn isn't really used right?
no strong feeling though

boarnoah: Sorry missed this one, we shouldn't need it because we have a account name in key, might as well just need to re-work the ``groupby`` expression a little

gord5500 (approved):

PR #397, opened 2023-08-30
check for updates on run

gord5500:
### Why

Helps users know we have a new version out. This is a middle step in terms of having an update feature. It'd be good to keep all our users on the latest version.

We will ping github once every 24 hours and save the version tag it returned. So currently if there is a new version the user will see it every time they run BMX. I could change that to once every 24 hours though but this also maybe annoys them enough to want to upgrade lol

If the org is d2l we prompt them to visit out site with the install script. Users outside of d2l will get prompted to visit the github repo
image
image

related trello card: https://trello.com/b/r6wdvBry/vulcan-project-board
slack discussion: https://d2l.slack.com/archives/G4NV94H2A/p1693323866881979

### DIFF
--- a/src/D2L.Bmx/Program.cs
+++ b/src/D2L.Bmx/Program.cs
@@ -7,7 +7,6 @@ using Amazon.SecurityToken;
 using D2L.Bmx;
 using D2L.Bmx.Aws;
 using D2L.Bmx.Okta;
-using D2L.Bmx.Okta.Models;
 using IniParser;
 
 
--- a/src/D2L.Bmx/Program.cs
+++ b/src/D2L.Bmx/Program.cs
@@ -201,10 +201,15 @@ var rootCommand = new RootCommand( "BMX grants you API access to your AWS accoun
 // start bmx
 return await new CommandLineBuilder( rootCommand )
 	.UseDefaults()
-	.AddMiddleware( async ( context, next ) => {
+	.AddMiddleware( async (
+		context,
+		next
+	) => {
 		await UpdateChecker.CheckForUpdatesAsync( configProvider.GetConfiguration() );
 		await next( context );
-	} )
+	},
+		System.CommandLine.Invocation.MiddlewareOrder.ExceptionHandler + 1
+	)
 	.UseExceptionHandler( ( exception, context ) => {
 		Console.ResetColor();
 		Console.ForegroundColor = ConsoleColor.Red;
--- a/src/D2L.Bmx/Program.cs
+++ b/src/D2L.Bmx/Program.cs
@@ -7,7 +7,11 @@ using Amazon.SecurityToken;
 using D2L.Bmx;
 using D2L.Bmx.Aws;
 using D2L.Bmx.Okta;
+using D2L.Bmx.Okta.Models;
 using IniParser;
+
+
+var configProvider = new BmxConfigProvider( new FileIniDataParser() );
 
 // common options
 var orgOption = new Option<string>(
@@ -27,7 +31,7 @@ loginCommand.SetHandler( ( InvocationContext context ) => {
 		new OktaApi(),
 		new OktaSessionStorage(),
 		new ConsolePrompter(),
-		new BmxConfigProvider( new FileIniDataParser() ).GetConfiguration()
+		configProvider.GetConfiguration()
 	) );
 	return handler.HandleAsync(
 		org: context.ParseResult.GetValueForOption( orgOption ),
@@ -64,7 +68,7 @@ var configureCommand = new Command( "configure", "Create or update the global BM
 
 configureCommand.SetHandler( ( InvocationContext context ) => {
 	var handler = new ConfigureHandler(
-		new BmxConfigProvider( new FileIniDataParser() ),
+		configProvider,
 		new ConsolePrompter() );
 	handler.Handle(
 		org: context.ParseResult.GetValueForOption( orgOption ),
@@ -115,7 +119,7 @@ var printCommand = new Command( "print", "Print AWS credentials" ) {
 
 printCommand.SetHandler( ( InvocationContext context ) => {
 	var consolePrompter = new ConsolePrompter();
-	var config = new BmxConfigProvider( new FileIniDataParser() ).GetConfiguration();
+	var config = configProvider.GetConfiguration();
 	var handler = new PrintHandler(
 		new OktaAuthenticator(
 			new OktaApi(),
@@ -159,7 +163,7 @@ var writeCommand = new Command( "write", "Write AWS credentials to the credentia
 
 writeCommand.SetHandler( ( InvocationContext context ) => {
 	var consolePrompter = new ConsolePrompter();
-	var config = new BmxConfigProvider( new FileIniDataParser() ).GetConfiguration();
+	var config = configProvider.GetConfiguration();
 	var handler = new WriteHandler(
 		new OktaAuthenticator(
 			new OktaApi(),
@@ -198,6 +202,10 @@ var rootCommand = new RootCommand( "BMX grants you API access to your AWS accoun
 // start bmx
 return await new CommandLineBuilder( rootCommand )
 	.UseDefaults()
+	.AddMiddleware( async ( context, next ) => {
+		await UpdateChecker.CheckForUpdatesAsync( configProvider.GetConfiguration() );
+		await next( context );
+	} )
 	.UseExceptionHandler( ( exception, context ) => {
 		Console.ResetColor();
 		Console.ForegroundColor = ConsoleColor.Red;
@@ -205,6 +213,7 @@ return await new CommandLineBuilder( rootCommand )
 			Console.Error.WriteLine( exception.Message );
 		} else {
 			Console.Error.WriteLine( "BMX exited with unexpected internal error" );
+			Console.Error.WriteLine( exception.ToString() );
 		}
 		if( Environment.GetEnvironmentVariable( "BMX_DEBUG" ) == "1" ) {
 			Console.Error.WriteLine( exception );
--- a/src/D2L.Bmx/Program.cs
+++ b/src/D2L.Bmx/Program.cs
@@ -213,7 +213,6 @@ return await new CommandLineBuilder( rootCommand )
 			Console.Error.WriteLine( exception.Message );
 		} else {
 			Console.Error.WriteLine( "BMX exited with unexpected internal error" );
-			Console.Error.WriteLine( exception.ToString() );
 		}
 		if( Environment.GetEnvironmentVariable( "BMX_DEBUG" ) == "1" ) {
 			Console.Error.WriteLine( exception );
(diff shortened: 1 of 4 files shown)


src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -0,0 +1,120 @@
...
+			if( latestVersion > localVersion ) {
+				DisplayUpdateMessage( $"A new BMX release is available: v{latestVersion} (current: v{localVersion})\n" +
+					$"Upgrade now at {updateLocation}" );
+			}
+		} catch( Exception ex ) {
+			// Give up and don't bother telling the user we couldn't check for updates
+		}
+	}
+
+	private static void DisplayUpdateMessage( string message ) {
```
gord5500, line 31: A bit of effort to make this message pop out like the PowerShell message. I couldn't get the gray background to not take up the entire width easily like they have though

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -0,0 +1,120 @@
...
+			string updateLocation = ( string.Equals( config.Org, "d2l", StringComparison.OrdinalIgnoreCase ) )
+				? "https://bmx.d2l.dev"
+				: "https://github.com/Brightspace/bmx/releases/latest";
+
+			if( latestVersion > localVersion ) {
+				DisplayUpdateMessage( $"A new BMX release is available: v{latestVersion} (current: v{localVersion})\n" +
+					$"Upgrade now at {updateLocation}" );
+			}
+		} catch( Exception ex ) {
+			// Give up and don't bother telling the user we couldn't check for updates
```
gord5500, line 27: Whatever happens during checking the update we absolutely don't want to break BMX so we catch any errors and just do nothing and move on

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -0,0 +1,119 @@
...
+		httpClient.DefaultRequestHeaders.Add( "User-Agent", "BMX" );
+		var response = await httpClient.GetAsync( "repos/Brightspace/bmx/releases/latest" );
+		response.EnsureSuccessStatusCode();
+
+		using var responseStream = await response.Content.ReadAsStreamAsync();
+		var releaseData = await JsonSerializer.DeserializeAsync(
+			responseStream,
+			SourceGenerationContext.Default.GithubRelease
+		);
+		string version = releaseData?.TagName?.Replace( "v", "" ) ?? string.Empty;
```
boarnoah, line 66: Does this correctly account for ignoring preview releases?

gord5500: We have this pre-release currently and it returns v3.0.0 atm so I would assume so https://github.com/Brightspace/bmx/releases/tag/v3.0.1-preview-20230802

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -0,0 +1,120 @@
...
+		using var fs = new FileStream( BmxPaths.UPDATE_CHECK_FILE_NAME, op );
+		using var writer = new StreamWriter( fs );
+		writer.WriteLine( version );
+	}
+
+	private static DateTimeOffset? GetTimeLastChecked() {
+		if( !File.Exists( BmxPaths.UPDATE_CHECK_FILE_NAME ) ) {
+			return null;
+		}
+		return File.GetLastWriteTimeUtc( BmxPaths.UPDATE_CHECK_FILE_NAME );
```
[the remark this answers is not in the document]
cfbao: I don't see how using file last write time makes this scenario better? It's equally fakable as a string we write to a file.
My suggestion was to discard any written "last checked time" that's in the future.

If people are willing to fake this once a day, then they win, and that's fine.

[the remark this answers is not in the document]
gord5500: Yea it's no better. I mostly just thought it was easier to only store the version in the file and not worry about any parsing heh

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -0,0 +1,119 @@
...
+
+			string updateLocation = ( string.Equals( config.Org, "d2l", StringComparison.OrdinalIgnoreCase ) )
+				? "https://bmx.d2l.dev"
+				: "https://github.com/Brightspace/bmx/releases/latest";
+
+			if( latestVersion > localVersion ) {
+				DisplayUpdateMessage( $"A new BMX release is available: v{latestVersion} (current: v{localVersion})\n" +
+					$"Upgrade now at {updateLocation}" );
+			}
+		} catch( Exception ex ) {
```
boarnoah, line 26:
```suggestion
		} catch( Exception ) {
```
Convention for making it clear you are catching an exception but don't care about the exception itself.

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -0,0 +1,119 @@
...
+	private static async Task<string> GetLatestReleaseVersionAsync() {
+		using var httpClient = new HttpClient {
+			BaseAddress = new Uri( "https://api.github.com" ),
+			Timeout = TimeSpan.FromSeconds( 5 )
+		};
+		httpClient.DefaultRequestHeaders.Add( "User-Agent", "BMX" );
+		var response = await httpClient.GetAsync( "repos/Brightspace/bmx/releases/latest" );
+		response.EnsureSuccessStatusCode();
+
+		using var responseStream = await response.Content.ReadAsStreamAsync();
```
boarnoah, line 61:
```suggestion
		await using var responseStream = await response.Content.ReadAsStreamAsync();
```
This works at present because Stream implements both ``IDisposable`` & ``IAsyncDisposable``, but IIUC the correct thing is to write it as ``await using`` if you are accessing something disposable asynchronously

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -0,0 +1,119 @@
...
+	public static async Task CheckForUpdatesAsync( BmxConfig config ) {
+		try {
+			string? savedLatestVersion = GetSavedLatestVersion();
+			var localVersion = Assembly.GetExecutingAssembly().GetName().Version;
+			var latestVersion = new Version( savedLatestVersion ?? "0.0.0" );
+			if( ShouldFetchLatestVersion( savedLatestVersion ) ) {
+				latestVersion = new Version( await GetLatestReleaseVersionAsync() );
+			}
+
+			string updateLocation = ( string.Equals( config.Org, "d2l", StringComparison.OrdinalIgnoreCase ) )
```
boarnoah, line 18:
```suggestion
			string updateLocation = string.Equals( config.Org, "d2l", StringComparison.OrdinalIgnoreCase )
```
nit: redundant parenthesis

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -0,0 +1,119 @@
...
+		Console.ForegroundColor = originalForegroundColor;
+		Console.ResetColor();
+		Console.Error.WriteLine();
+	}
+
+	private static async Task<string> GetLatestReleaseVersionAsync() {
+		using var httpClient = new HttpClient {
+			BaseAddress = new Uri( "https://api.github.com" ),
+			Timeout = TimeSpan.FromSeconds( 5 )
+		};
```
boarnoah, line 56:
```suggestion
		using var httpClient = new HttpClient();
		httpClient.BaseAddress = new Uri( "https://api.github.com" );
		httpClient.Timeout = TimeSpan.FromSeconds( 5 );
```

When you are initializing properties this way it won't properly dispose if there is an error (again unlikely, but lets say you had an exception constructing the Uri object here).

https://stackoverflow.com/a/57204099/4433646

In this particular case (with HttpClient) I don't think it would have mattered anyway (since just constructing an HttpClient instance doesn't grab hold of anything that needs disposal (?)

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -0,0 +1,119 @@
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace D2L.Bmx;
+internal class UpdateChecker {
```
boarnoah, line 7:
```suggestion
internal static class UpdateChecker {
```

This class is a collection of helper statics, not to be instantiated

src/D2L.Bmx/UpdateChecker.cs
gord5500: Ooh thanks

src/D2L.Bmx/UpdateChecker.cs
gord5500: Ah yea was doing something jank at first 😅

src/D2L.Bmx/Program.cs
```diff
@@ -198,6 +201,15 @@
...
 return await new CommandLineBuilder( rootCommand )
 	.UseDefaults()
+	.AddMiddleware( async (
+		context,
+		next
+	) => {
+		await UpdateChecker.CheckForUpdatesAsync( configProvider.GetConfiguration() );
+		await next( context );
+	},
+		System.CommandLine.Invocation.MiddlewareOrder.ExceptionHandler + 1
```
boarnoah, line 211: I suppose this doesn't hurt, but this particular middle ware (the update checker) won't ever throw an exception so it seems kind of pointless to set this.

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -0,0 +1,127 @@
...
+namespace D2L.Bmx;
+internal static class UpdateChecker {
+
+	public static async Task CheckForUpdatesAsync( BmxConfig config ) {
+		try {
+			var cachedVersion = GetUpdateCheckCache();
+			var localVersion = Assembly.GetExecutingAssembly().GetName().Version;
+			var latestVersion = new Version( cachedVersion?.VersionName ?? "0.0.0" );
+			if( ShouldFetchLatestVersion( cachedVersion ) ) {
+				Console.Error.WriteLine( "Getting new version from github" );
```
boarnoah, line 15: This isn't in the screenshot so presumably you added this during dev to test. We should probably remove it right?

A little confusing to have it, makes it sound like we are potentially fetching & installing the new version.

boarnoah (approved):

src/D2L.Bmx/Program.cs
gord5500: Sorry this is about the placement in the middleware pipeline where this update checker is going to run. At the default order level, this wouldn't get run if a user were to run --version or --help

Or do you mean that the + 1 to have it run after the exception handling middleware is what could go?

src/D2L.Bmx/UpdateChecker.cs
gord5500: It was for debug sorry. Will take this out

boarnoah (approved):

PR #407, opened 2023-09-22
update checker: reduce timetout to 2 seconds and trim tag name

gord5500:
### Why

Shorter timeout so people don't have to wait as long

### DIFF
--- a/src/D2L.Bmx/UpdateChecker.cs
+++ b/src/D2L.Bmx/UpdateChecker.cs
@@ -62,7 +62,7 @@ internal static class UpdateChecker {
 			responseStream,
 			SourceGenerationContext.Default.GithubRelease
 		);
-		string version = releaseData?.TagName?.Trim( 'v' ) ?? string.Empty;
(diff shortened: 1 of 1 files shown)


src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -62,7 +62,7 @@ internal static class UpdateChecker {
 			responseStream,
 			SourceGenerationContext.Default.GithubRelease
 		);
-		string version = releaseData?.TagName?.Replace( "v", "" ) ?? string.Empty;
+		string version = releaseData?.TagName?.Trim( 'v' ) ?? string.Empty;
```
jharoutuniand2l, line 65: I think we specifically want `TrimStart()` here IIUC

gord5500: Oh ye I see docs says it trims beginning and end. I switched

jharoutuniand2l (approved):

PR #408, opened 2023-09-26
Okta and AWS broader error handling

alex-fang0:
### Why
Adding more error handling to network calls to avoid the general "BMX exited with unexpected internal error" error.

### DIFF
--- a/src/D2L.Bmx/Aws/AwsClient.cs
+++ b/src/D2L.Bmx/Aws/AwsClient.cs
@@ -28,11 +28,11 @@ internal class AwsClient( IAmazonSecurityTokenService stsClient ) : IAwsClient {
 				SecretAccessKey: authResp.Credentials.SecretAccessKey,
 				Expiration: authResp.Credentials.Expiration.ToUniversalTime()
 			);
-		} catch( Exception ex ) {
-			if( ex.Message == "The requested DurationSeconds exceeds the MaxSessionDuration set for this role." ) {
-				throw new BmxException( "Duration exceeds the MaxSessionDuration for this role. Lower it in config/parameter", ex );
-			}
-			throw new BmxException( "Error AWS failed to grab credentials", ex );
+		} catch( AmazonSecurityTokenServiceException ex ) when
+			( ex.Message == "The requested DurationSeconds exceeds the MaxSessionDuration set for this role." ) {
+			throw new BmxException( "Duration exceeds the MaxSessionDuration for this role. Lower it in config/parameter", ex );
+		} catch (Exception ex ) {
+			throw new BmxException( "Error getting credentials from AWS", ex );
 		}
 	}
 }
--- a/src/D2L.Bmx/Aws/AwsClient.cs
+++ b/src/D2L.Bmx/Aws/AwsClient.cs
@@ -13,18 +13,26 @@ internal class AwsClient( IAmazonSecurityTokenService stsClient ) : IAwsClient {
 		AwsRole role,
 		int durationInMinutes
 	) {
-		var authResp = await stsClient.AssumeRoleWithSAMLAsync( new AssumeRoleWithSAMLRequest {
-			PrincipalArn = role.PrincipalArn,
-			RoleArn = role.RoleArn,
-			SAMLAssertion = samlAssertion,
-			DurationSeconds = durationInMinutes * 60,
-		} );
 
-		return new AwsCredentials(
-			SessionToken: authResp.Credentials.SessionToken,
-			AccessKeyId: authResp.Credentials.AccessKeyId,
-			SecretAccessKey: authResp.Credentials.SecretAccessKey,
-			Expiration: authResp.Credentials.Expiration.ToUniversalTime()
-		);
+		try {
+			var authResp = await stsClient.AssumeRoleWithSAMLAsync( new AssumeRoleWithSAMLRequest {
+				PrincipalArn = role.PrincipalArn,
+				RoleArn = role.RoleArn,
+				SAMLAssertion = samlAssertion,
+				DurationSeconds = durationInMinutes * 60,
+			} );
+
+			return new AwsCredentials(
+				SessionToken: authResp.Credentials.SessionToken,
+				AccessKeyId: authResp.Credentials.AccessKeyId,
+				SecretAccessKey: authResp.Credentials.SecretAccessKey,
+				Expiration: authResp.Credentials.Expiration.ToUniversalTime()
+			);
+		} catch ( Exception ex ) {
+			if (ex.Message == "The requested DurationSeconds exceeds the MaxSessionDuration set for this role.") {
+				throw new BmxException( "Duration exceeds the MaxSessionDuration for this role. Lower it in config/parameter", ex );
+			}
+			throw new BmxException( "Error AWS failed to grab credentials", ex );
+		}
 	}
 }
--- a/src/D2L.Bmx/Aws/AwsClient.cs
+++ b/src/D2L.Bmx/Aws/AwsClient.cs
@@ -31,7 +31,7 @@ internal class AwsClient( IAmazonSecurityTokenService stsClient ) : IAwsClient {
 		} catch( AmazonSecurityTokenServiceException ex ) when
 			( ex.Message == "The requested DurationSeconds exceeds the MaxSessionDuration set for this role." ) {
 			throw new BmxException( "Duration exceeds the MaxSessionDuration for this role. Lower it in config/parameter", ex );
-		} catch (Exception ex ) {
+		} catch( Exception ex ) {
 			throw new BmxException( "Error getting credentials from AWS", ex );
 		}
 	}
--- a/src/D2L.Bmx/Okta/OktaApi.cs
+++ b/src/D2L.Bmx/Okta/OktaApi.cs
@@ -77,7 +77,7 @@ internal class OktaApi : IOktaApi {
 			throw new BmxException( "Okta API Authentication Response was not Success or MFA Required" );
 		} catch( Exception ex ) {
 			if ( ex is BmxException ) {
-				throw new BmxException( "Okta authentication failed. Check if ./bmx/config org and user is correct", ex );
+				throw new BmxException( "Okta authentication failed. Check if org, user and password is correct", ex );
 			}
 			throw new BmxException( "Error authenticating to Okta" , ex);
 		}
--- a/src/D2L.Bmx/Okta/OktaApi.cs
+++ b/src/D2L.Bmx/Okta/OktaApi.cs
@@ -49,38 +49,36 @@ internal class OktaApi : IOktaApi {
 	}
 
 	async Task<AuthenticateResponse> IOktaApi.AuthenticateAsync( string username, string password ) {
+		HttpResponseMessage resp;
 		try {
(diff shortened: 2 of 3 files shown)


src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -13,18 +13,26 @@ internal class AwsClient( IAmazonSecurityTokenService stsClient ) : IAwsClient {
...
+
+			return new AwsCredentials(
+				SessionToken: authResp.Credentials.SessionToken,
+				AccessKeyId: authResp.Credentials.AccessKeyId,
+				SecretAccessKey: authResp.Credentials.SecretAccessKey,
+				Expiration: authResp.Credentials.Expiration.ToUniversalTime()
+			);
+		} catch ( Exception ex ) {
+			if (ex.Message == "The requested DurationSeconds exceeds the MaxSessionDuration set for this role.") {
+				throw new BmxException( "Duration exceeds the MaxSessionDuration for this role. Lower it in config/parameter", ex );
```
gord5500, line 33: Do you know what the other errors are and are they just not useful or easy to understand?

alex-fang0: Yea I looked over the [rest of the errors](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithSAML.html#API_AssumeRoleWithSAML_Errors) and they seem pretty unlikely to happen and not too user friendly if I we're to add a case

alex-fang0: Some more general [common errors](https://docs.aws.amazon.com/STS/latest/APIReference/CommonErrors.html) here but most of them seem to be handled by how bmx is used. The only one that could be added I would think is ThrottlingException

gord5500: Sounds good 🤝

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -92,44 +101,55 @@ internal class OktaApi : IOktaApi {
...
+		try {
+			var resp = await _httpClient.PostAsJsonAsync(
+				"sessions",
+				new CreateSessionRequest( sessionToken ),
+				SourceGenerationContext.Default.CreateSessionRequest );
+
+			var session = await JsonSerializer.DeserializeAsync(
+				await resp.Content.ReadAsStreamAsync(),
+				SourceGenerationContext.Default.OktaSession );
+			return session ?? throw new BmxException( "Creating Octa Session returned null" );
```
gord5500, line 136:
```suggestion
			return session ?? throw new BmxException( "Creating Okta Session returned null" );
```

Also your title has it as Octa

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -49,41 +49,50 @@ internal class OktaApi : IOktaApi {
...
+				StateToken: not null,
+				Embedded.Factors: not null,
+				Status: AuthenticationStatus.MFA_REQUIRED
+			} ) {
+				return new AuthenticateResponse.MfaRequired(
+					authnResponse.StateToken,
+					authnResponse.Embedded.Factors
+				);
+			}
+			throw new BmxException( "Okta API Authentication Response was not Success or MFA Required" );
```
gord5500, line 77: Would this text ever get shown to the user since it's caught a few lines down and then we throw a new exception with different message

alex-fang0: No this would not show for the user. This is more for when Debug mode is on and it shows in the inner exception to make it a bit easier to figure out what went wrong. Not sure how useful it would be tho

src/D2L.Bmx/Aws/AwsClient.cs
alex-fang0: Looking online I can't even seem to find anyone running into or an example of the ThrottlingException error for AssumeRole. I think it should be fine to leave out

src/D2L.Bmx/Okta/OktaApi.cs
gord5500: Ah Oki. Just wanted to make sure it was planned to not show them

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -92,44 +101,55 @@ internal class OktaApi : IOktaApi {
...
-			?? throw new BmxException( "Error retrieving AWS accounts from Okta" );
+		try {
+			var apps = await _httpClient.GetFromJsonAsync(
+				"users/me/appLinks",
+				SourceGenerationContext.Default.OktaAppArray );
+
+			return apps?.Where( app => app.AppName == "amazon_aws" ).ToArray()
+				?? throw new BmxException( "Retrieving AWS accounts from Okta returned null" );
+		} catch( Exception ex ) {
+			throw new BmxException( "Error retrieving AWS accounts from Okta", ex );
```
gord5500, line 151: I think these extra catches could be misleading. If I change the route [here](https://github.com/Brightspace/bmx/blob/3e365d3e5d48abc8b3f9ec80b3bba868d460d3eb/src/D2L.Bmx/Okta/OktaApi.cs/#L129) from `sessions` to `sessionss` I'd expect to see the exception message mentioning a problem to create an okta session. Instead though I get this message here `Error retrieving AWS accounts from Okta`

gord5500: Maybe we need to also pass up the caught exception message if it's a bmx exception?

src/D2L.Bmx/Okta/OktaApi.cs
cfbao: I'd really like to avoid nesting BmxExceptions. It's intended as a mechanism to show user facing messages, and I'd like to keep it that way. Using it in multiple different ways makes our code more confusing.

Also, I'd consider throwing exception then immediately catch it below not the best form in general. There's usually a better pattern that can express the desired logic flow.

boarnoah: Its hard to articulate exactly what "Okta API Authentication Response was not Success or MFA Required" exception is attempting to capture. 

There is a few situations we need to deal with here (in general):
1. Bad initial credentials
1. Failed MFA challenge
1. Bad connectivity, any other un-handled issues

Wrapping everything under one exception handler only makes sense for that third case (where its an un handled situation that could happen for any  of the **other** number of reasons).

src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -13,18 +13,26 @@ internal class AwsClient( IAmazonSecurityTokenService stsClient ) : IAwsClient {
...
+			} );
+
+			return new AwsCredentials(
+				SessionToken: authResp.Credentials.SessionToken,
+				AccessKeyId: authResp.Credentials.AccessKeyId,
+				SecretAccessKey: authResp.Credentials.SecretAccessKey,
+				Expiration: authResp.Credentials.Expiration.ToUniversalTime()
+			);
+		} catch( Exception ex ) {
+			if( ex.Message == "The requested DurationSeconds exceeds the MaxSessionDuration set for this role." ) {
```
cfbao, line 32: this case has a more specific exception type, right?
you can / should use a more specific `catch` clause in such scenarios, e.g.
```csharp
} catch ( AmazonSecurityTokenServiceException ex ) when ( ex.Message == "foo" ) {
	throw new BmxException( "don't do foo", ex );
} catch ( Exception ex ) {
	throw new BmxException( "don't do... idk... this", ex );
}
```

src/D2L.Bmx/Aws/AwsClient.cs
```diff
@@ -13,18 +13,26 @@ internal class AwsClient( IAmazonSecurityTokenService stsClient ) : IAwsClient {
...
+				throw new BmxException( "Duration exceeds the MaxSessionDuration for this role. Lower it in config/parameter", ex );
+			}
+			throw new BmxException( "Error AWS failed to grab credentials", ex );
```
cfbao, line 35: we can rephrase it better? thinking:

"Failed to get credentials from AWS"
or
"Error getting credentials from AWS"

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -49,10 +49,15 @@ internal class OktaApi : IOktaApi {
...
-			new AuthenticateRequest( username, password ),
-			SourceGenerationContext.Default.AuthenticateRequest );
+		HttpResponseMessage resp;
+		try {
+			resp = await _httpClient.PostAsJsonAsync(
+				"authn",
+				new AuthenticateRequest( username, password ),
+				SourceGenerationContext.Default.AuthenticateRequest );
+		} catch( Exception ex ) {
+			throw new BmxException( "Okta authentication request failed. Try checking for a stable connection.", ex );
```
cfbao, line 59: should probably reword this to be broader. bad network connection isn't the only reason this can fail. It can also be Okta's fault (e.g. site down. it happened a few times)

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -73,17 +78,19 @@ internal class OktaApi : IOktaApi {
...
+			var response = await _httpClient.PostAsJsonAsync(
 			$"authn/factors/{factorId}/verify",
 			new IssueMfaChallengeRequest( stateToken ),
 			SourceGenerationContext.Default.IssueMfaChallengeRequest );
 
-		if( !response.IsSuccessStatusCode ) {
-			throw new BmxException( "Error sending MFA challenge" );
+			response.EnsureSuccessStatusCode();
+		} catch( Exception ex ) {
+			throw new BmxException( "Error sending MFA challenge", ex );
```
cfbao, line 93: I know you only kept the original message here, but I'm thinking we can reword this to be more user friendly. e.g.

> Error initiating MFA with Okta

maybe? is this an accurate description?

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -95,41 +102,55 @@ string challengeResponse
...
-			SourceGenerationContext.Default.VerifyMfaChallengeResponseRequest );
-
+		HttpResponseMessage resp;
+		try {
+			resp = await _httpClient.PostAsJsonAsync(
+				$"authn/factors/{factorId}/verify",
+				request,
+				SourceGenerationContext.Default.VerifyMfaChallengeResponseRequest );
+		} catch( Exception ex ) {
+			throw new BmxException( "Error verifying Okta MFA challenge response. Try checking for a stable connection.", ex );
```
cfbao, line 112: similarly here (and below). the message should be reworded to cover broader possible causes

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -95,41 +102,55 @@ string challengeResponse
...
+		}
 		var authnResponse = await JsonSerializer.DeserializeAsync(
```
cfbao, line 114: a try catch around json deserialization would be good too. possible for Okta to return invalid responses.
similarly in a few other places

cfbao (approved):
Looks good!

(btw I also updated the PR title "Octa" -> "Okta")

PR #413, opened 2023-10-27
okta: removing /me check

alex-fang0:
### Why
Trusting session cache and removing check with api

### DIFF
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -43,7 +43,7 @@ internal class OktaAuthenticator(
 
 		oktaApi.SetOrganization( org );
 
-		if( !ignoreCache && await TryAuthenticateFromCacheAsync( org, user, oktaApi ) ) {
+		if( !ignoreCache && TryAuthenticateFromCache( org, user, oktaApi ) ) {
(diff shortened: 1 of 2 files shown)


src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -100,8 +100,7 @@ IOktaApi oktaApi
 		}
 
 		oktaApi.AddSession( sessionId );
-		string? userId = await oktaApi.GetCurrentUserIdAsync( sessionId );
```
alex-fang0, line 103: Should I remove the oktaApi.GetCurrentUserIdAsync command since its now unused?

alex-fang0:
In the case SessionId cached is invalid and is used, it throws [here](https://github.com/Brightspace/bmx/blob/api/okta-users/me-disable/src/D2L.Bmx/Okta/OktaApi.cs#L167) with a 403 (Forbidden). We can try to ask the user to try bmx login again but 403 (Forbidden) can be caused by other cases too.

src/D2L.Bmx/OktaAuthenticator.cs
scowing: Yeah if a function has no current use and no planned future use then just remove it.

scowing (approved):

PR #414, opened 2023-10-30
Changing cache location and deleting old files

alex-fang0:
### Why
User looking to modify the config file manually may get confused by all the different files in .bmx

### DIFF
--- a/src/D2L.Bmx/Aws/AwsCredsCache.cs
+++ b/src/D2L.Bmx/Aws/AwsCredsCache.cs
@@ -69,13 +69,6 @@ internal class AwsCredsCache : IAwsCredentialCache {
 		string jsonString = JsonSerializer.Serialize( prunedEntries.ToList(),
 			SourceGenerationContext.Default.ListAwsCacheModel );
 
-		if( !Directory.Exists( BmxPaths.CACHE_DIR ) ) {
-			Directory.CreateDirectory( BmxPaths.CACHE_DIR );
-			File.Delete( Path.Join( BmxPaths.BMX_DIR, "awsCredsCache" ) );
-			File.Delete( Path.Join( BmxPaths.BMX_DIR, "sessions" ) );
-			File.Delete( Path.Join( BmxPaths.BMX_DIR, "update_check" ) );
-		}
-
 		WriteTextToFile( BmxPaths.AWS_CREDS_CACHE_FILE_NAME, jsonString );
 	}
 
--- a/src/D2L.Bmx/Aws/AwsCredsCache.cs
+++ b/src/D2L.Bmx/Aws/AwsCredsCache.cs
@@ -69,6 +69,13 @@ internal class AwsCredsCache : IAwsCredentialCache {
 		string jsonString = JsonSerializer.Serialize( prunedEntries.ToList(),
 			SourceGenerationContext.Default.ListAwsCacheModel );
 
+		if( !Directory.Exists( BmxPaths.CACHE_DIR ) ) {
+			Directory.CreateDirectory( BmxPaths.CACHE_DIR );
+			File.Delete( Path.Join( BmxPaths.BMX_DIR, "awsCredsCache" ) );
+			File.Delete( Path.Join( BmxPaths.BMX_DIR, "sessions" ) );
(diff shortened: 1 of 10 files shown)


src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -69,6 +69,13 @@ AwsCredentials credentials
 		string jsonString = JsonSerializer.Serialize( prunedEntries.ToList(),
 			SourceGenerationContext.Default.ListAwsCacheModel );
 
+		if( !Directory.Exists( BmxPaths.CACHE_DIR ) ) {
+			Directory.CreateDirectory( BmxPaths.CACHE_DIR );
+			File.Delete( Path.Join( BmxPaths.BMX_DIR, "awsCredsCache" ) );
+			File.Delete( Path.Join( BmxPaths.BMX_DIR, "sessions" ) );
+			File.Delete( Path.Join( BmxPaths.BMX_DIR, "update_check" ) );
+		}
```
boarnoah, line 77: Consider moving this up closer to the application entry point (the print/write/login handlers?)

Part of the direction we want to move here is to have less duplicated code that has to concern itself about the status of the  files in ``.bmx``

gord5500 (approved):

src/D2L.Bmx/Program.cs
```diff
@@ -215,6 +215,13 @@
 		next
 	) => {
 		await UpdateChecker.CheckForUpdatesAsync( configProvider.GetConfiguration() );
```
alex-fang0, line 217: Note: This check creates .bmx folder if it doesn't exist already so we can remove the other folder checks

src/D2L.Bmx/Program.cs
```diff
@@ -215,6 +215,13 @@
...
+		//Creating new Cache directory and removing old files.
+		if( !Directory.Exists( BmxPaths.CACHE_DIR ) ) {
+			Directory.CreateDirectory( BmxPaths.CACHE_DIR );
+			File.Delete( Path.Join( BmxPaths.BMX_DIR, "awsCredsCache" ) );
+			File.Delete( Path.Join( BmxPaths.BMX_DIR, "sessions" ) );
+			File.Delete( Path.Join( BmxPaths.BMX_DIR, "update_check" ) );
+		}
```
boarnoah, line 224: Makes sense to handle exceptions gracefully here for the especially for the cleanup of the old files.

alex-fang0: I don't think the Directory or File functions error out even if the files don't exist

boarnoah: Mhm, I was more concerned about the other exceptions. Especially for the cleanup part of this (i.e everything aside from CreateDirectory).
https://learn.microsoft.com/en-us/dotnet/api/system.io.file.delete?view=net-7.0#exceptions

Ideally we would like BMX to continue working if for w/e reason it can't actually delete those files.

alex-fang0: Gotcha! makes sense

src/D2L.Bmx/Program.cs
```diff
@@ -217,10 +217,14 @@
...
-			File.Delete( Path.Join( BmxPaths.BMX_DIR, "awsCredsCache" ) );
-			File.Delete( Path.Join( BmxPaths.BMX_DIR, "sessions" ) );
-			File.Delete( Path.Join( BmxPaths.BMX_DIR, "update_check" ) );
+			try {
+				Directory.CreateDirectory( BmxPaths.CACHE_DIR );
+				File.Delete( Path.Join( BmxPaths.BMX_DIR, "awsCredsCache" ) );
+				File.Delete( Path.Join( BmxPaths.BMX_DIR, "sessions" ) );
+				File.Delete( Path.Join( BmxPaths.BMX_DIR, "update_check" ) );
+			} catch ( Exception ex ) {
+				Console.WriteLine( "Error cleaning up old files continuing..." );
```
boarnoah, line 226: ``Console.Error.Write`` here, we pipe the output from stdout when using bmx like: ``bmx print | Invoke-Expression``
So any text that needs to be shown but not passed into piped commands needs to go to stderr.

boarnoah (approved):

PR #418, opened 2023-12-07
Adding update command

gord5500:
### Why

Need a command to update bmx with a new version if one is available

### DIFF
--- a/src/D2L.Bmx/BmxPaths.cs
+++ b/src/D2L.Bmx/BmxPaths.cs
@@ -11,5 +11,5 @@ internal static class BmxPaths {
 	public static readonly string SESSIONS_FILE_LEGACY_NAME = Path.Join( BMX_DIR, "sessions" );
 	public static readonly string UPDATE_CHECK_FILE_NAME = Path.Join( CACHE_DIR, "updateCheck" );
 	public static readonly string AWS_CREDS_CACHE_FILE_NAME = Path.Join( CACHE_DIR, "awsCreds" );
-	public static readonly string OLD_BMX_VERSION_FILE_NAME = Path.Join( BMX_DIR, "bmx" );
+	public static readonly string OLD_BMX_VERSIONS_PATH = Path.Join( BMX_DIR, "temp" );
 }
--- a/src/D2L.Bmx/BmxPaths.cs
+++ b/src/D2L.Bmx/BmxPaths.cs
@@ -11,4 +11,5 @@ internal static class BmxPaths {
 	public static readonly string SESSIONS_FILE_LEGACY_NAME = Path.Join( BMX_DIR, "sessions" );
 	public static readonly string UPDATE_CHECK_FILE_NAME = Path.Join( CACHE_DIR, "updateCheck" );
 	public static readonly string AWS_CREDS_CACHE_FILE_NAME = Path.Join( CACHE_DIR, "awsCreds" );
+	public static readonly string OLD_BMX_VERSION_FILE_NAME = Path.Join( BMX_DIR, "bmx" );
 }
--- a/src/D2L.Bmx/GithubRelease.cs
+++ b/src/D2L.Bmx/GithubRelease.cs
@@ -10,9 +10,6 @@ internal record GithubRelease {
 }
 
 internal record Assets {
-	[JsonPropertyName( "url" )]
-	public string? Url { get; set; }
-
 	[JsonPropertyName( "name" )]
 	public string? Name { get; set; }
 
--- a/src/D2L.Bmx/GithubRelease.cs
+++ b/src/D2L.Bmx/GithubRelease.cs
@@ -10,7 +10,7 @@ internal record GithubRelease {
 	public List<GithubAsset>? Assets { get; init; }
 
 	public Version? GetReleaseVersion() {
-		string? version = TagName?.TrimStart( 'v' ) ?? null;
+		string? version = TagName?.TrimStart( 'v' );
 		if( version is null ) {
 			return null;
 		}
--- a/src/D2L.Bmx/GithubRelease.cs
+++ b/src/D2L.Bmx/GithubRelease.cs
@@ -1,18 +1,42 @@
+using System.Text.Json;
 using System.Text.Json.Serialization;
 
 namespace D2L.Bmx;
 internal record GithubRelease {
 	[JsonPropertyName( "tag_name" )]
-	public string? TagName { get; set; }
+	public string? TagName { get; init; }
 
 	[JsonPropertyName( "assets" )]
-	public List<Assets>? Assets { get; set; }
+	public List<GithubAsset>? Assets { get; init; }
 }
 
-internal record Assets {
+internal record GithubAsset {
 	[JsonPropertyName( "name" )]
-	public string? Name { get; set; }
+	public string? Name { get; init; }
 
 	[JsonPropertyName( "browser_download_url" )]
-	public string? BrowserDownloadUrl { get; set; }
+	public string? BrowserDownloadUrl { get; init; }
+}
+
+internal static class GithubUtilities {
+	public static string GetReleaseVersion( GithubRelease? releaseData ) {
+		string version = releaseData?.TagName?.TrimStart( 'v' ) ?? string.Empty;
+		return version;
+	}
+
+	public static async Task<GithubRelease?> GetLatestReleaseDataAsync() {
+		using var httpClient = new HttpClient();
+		httpClient.BaseAddress = new Uri( "https://api.github.com" );
+		httpClient.Timeout = TimeSpan.FromSeconds( 2 );
+		httpClient.DefaultRequestHeaders.Add( "User-Agent", "BMX" );
+		var response = await httpClient.GetAsync( "repos/Brightspace/bmx/releases/latest" );
+		response.EnsureSuccessStatusCode();
+
+		await using var responseStream = await response.Content.ReadAsStreamAsync();
+		var releaseData = await JsonSerializer.DeserializeAsync(
+			responseStream,
+			SourceGenerationContext.Default.GithubRelease
+		);
+		return releaseData;
+	}
 }
--- a/src/D2L.Bmx/GithubRelease.cs
+++ b/src/D2L.Bmx/GithubRelease.cs
@@ -1,0 +1,21 @@
+using System.Text.Json.Serialization;
+
+namespace D2L.Bmx;
+internal record GithubRelease {
+	[JsonPropertyName( "tag_name" )]
+	public string? TagName { get; set; }
+
+	[JsonPropertyName( "assets" )]
+	public List<Assets>? Assets { get; set; }
+}
+
+internal record Assets {
+	[JsonPropertyName( "url" )]
+	public string? Url { get; set; }
+
+	[JsonPropertyName( "name" )]
+	public string? Name { get; set; }
+
+	[JsonPropertyName( "browser_download_url" )]
+	public string? BrowserDownloadUrl { get; set; }
+}
--- a/src/D2L.Bmx/GithubRelease.cs
+++ b/src/D2L.Bmx/GithubRelease.cs
@@ -1,13 +1,31 @@
-using System.Text.Json;
+using System.Net.Http.Json;
 using System.Text.Json.Serialization;
 
 namespace D2L.Bmx;
 internal record GithubRelease {
 	[JsonPropertyName( "tag_name" )]
-	public string? TagName { get; init; }
+	public string? TagName { get; init;}
 
 	[JsonPropertyName( "assets" )]
 	public List<GithubAsset>? Assets { get; init; }
+
+	public Version? GetReleaseVersion() {
+		string? version = TagName?.TrimStart( 'v' ) ?? null;
+		if( version is null ) {
+			return null;
+		}
+		return new Version( version );
+	}
+
+	public static async Task<GithubRelease?> GetLatestReleaseDataAsync() {
+		using var httpClient = new HttpClient();
+		httpClient.BaseAddress = new Uri( "https://api.github.com" );
+		httpClient.Timeout = TimeSpan.FromSeconds( 2 );
+		httpClient.DefaultRequestHeaders.Add( "User-Agent", "BMX" );
+		return await httpClient.GetFromJsonAsync(
+			"repos/Brightspace/bmx/releases/latest",
+			SourceGenerationContext.Default.GithubRelease );
+	}
 }
 
 internal record GithubAsset {
@@ -16,27 +34,4 @@ internal record GithubAsset {
 
 	[JsonPropertyName( "browser_download_url" )]
 	public string? BrowserDownloadUrl { get; init; }
-}
-
-internal static class GithubUtilities {
-	public static string GetReleaseVersion( GithubRelease? releaseData ) {
-		string version = releaseData?.TagName?.TrimStart( 'v' ) ?? string.Empty;
-		return version;
-	}
-
-	public static async Task<GithubRelease?> GetLatestReleaseDataAsync() {
-		using var httpClient = new HttpClient();
-		httpClient.BaseAddress = new Uri( "https://api.github.com" );
-		httpClient.Timeout = TimeSpan.FromSeconds( 2 );
-		httpClient.DefaultRequestHeaders.Add( "User-Agent", "BMX" );
-		var response = await httpClient.GetAsync( "repos/Brightspace/bmx/releases/latest" );
-		response.EnsureSuccessStatusCode();
-
-		await using var responseStream = await response.Content.ReadAsStreamAsync();
-		var releaseData = await JsonSerializer.DeserializeAsync(
-			responseStream,
-			SourceGenerationContext.Default.GithubRelease
-		);
-		return releaseData;
-	}
 }
--- a/src/D2L.Bmx/GithubRelease.cs
+++ b/src/D2L.Bmx/GithubRelease.cs
@@ -4,7 +4,7 @@ using System.Text.Json.Serialization;
 namespace D2L.Bmx;
 internal record GithubRelease {
 	[JsonPropertyName( "tag_name" )]
-	public string? TagName { get; init;}
+	public string? TagName { get; init; }
 
 	[JsonPropertyName( "assets" )]
 	public List<GithubAsset>? Assets { get; init; }
--- a/src/D2L.Bmx/UpdateChecker.cs
+++ b/src/D2L.Bmx/UpdateChecker.cs
@@ -11,7 +11,9 @@ internal static class UpdateChecker {
 			var localVersion = Assembly.GetExecutingAssembly().GetName().Version;
 			var latestVersion = new Version( cachedVersion?.VersionName ?? "0.0.0" );
 			if( ShouldFetchLatestVersion( cachedVersion ) ) {
-				latestVersion = new Version( GetLatestReleaseVersion( await GetLatestReleaseDataAsync() ) );
+				var latestVersionString = GithubUtilities.GetReleaseVersion( await GithubUtilities.GetLatestReleaseDataAsync() );
+				latestVersion = new Version ( latestVersionString );
+				SaveVersion( latestVersionString );
 			}
 
 			string updateLocation = string.Equals( config.Org, "d2l", StringComparison.OrdinalIgnoreCase )
@@ -48,32 +50,7 @@ internal static class UpdateChecker {
 		Console.Error.WriteLine();
 	}
 
-	internal static string GetLatestReleaseVersion( GithubRelease? releaseData ) {
-		string version = releaseData?.TagName?.TrimStart( 'v' ) ?? string.Empty;
-		SaveLatestVersion( version );
-		return version;
-	}
-
-	internal static async Task<GithubRelease?> GetLatestReleaseDataAsync() {
-		using var httpClient = new HttpClient();
-		httpClient.BaseAddress = new Uri( "https://api.github.com" );
-		httpClient.Timeout = TimeSpan.FromSeconds( 2 );
-		httpClient.DefaultRequestHeaders.Add( "User-Agent", "BMX" );
-		var response = await httpClient.GetAsync( "repos/Brightspace/bmx/releases/latest" );
-		response.EnsureSuccessStatusCode();
-
-		await using var responseStream = await response.Content.ReadAsStreamAsync();
-		var releaseData = await JsonSerializer.DeserializeAsync(
-			responseStream,
-			SourceGenerationContext.Default.GithubRelease
-		);
-		return releaseData;
-	}
-
-	private static void SaveLatestVersion( string version ) {
-		if( string.IsNullOrWhiteSpace( version ) ) {
-			return;
-		}
+	private static void SaveVersion( string version ) {
 		var cache = new UpdateCheckCache {
 			VersionName = version,
 			TimeLastChecked = DateTimeOffset.UtcNow
--- a/src/D2L.Bmx/UpdateChecker.cs
+++ b/src/D2L.Bmx/UpdateChecker.cs
@@ -111,7 +111,7 @@ internal static class UpdateChecker {
 		}
 		return false;
 	}
-	
+
 	public static void UpdateSelf( byte[] buffer ) {
 		string self = System.Reflection.Assembly.GetExecutingAssembly().Location;
 		self = self.Replace( ".dll", ".exe" );
@@ -148,12 +148,26 @@ internal static class UpdateChecker {
 			Environment.Exit( 0 );
 		}
 	}
-	
+
 }
 
 internal record GithubRelease {
 	[JsonPropertyName( "tag_name" )]
 	public string? TagName { get; set; }
+
+	[JsonPropertyName( "assets" )]
+	public List<Assets>? Assets { get; set; }
+}
+
+internal record Assets {
+	[JsonPropertyName( "url" )]
+	public string? Url { get; set; }
+
+	[JsonPropertyName( "name" )]
+	public string? Name { get; set; }
+
+	[JsonPropertyName( "browser_download_url" )]
+	public string? BrowserDownloadUrl { get; set; }
 }
 
 internal record UpdateCheckCache {
--- a/src/D2L.Bmx/UpdateChecker.cs
+++ b/src/D2L.Bmx/UpdateChecker.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text.Json;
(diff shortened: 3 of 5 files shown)


src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,136 @@
...
+			}
+		} catch	( Exception ex ) {
+			throw new BmxException( "Failed to download the update", ex );
+		}
+
+
+		string? currentProcessPath = Environment.ProcessPath;
+		string? currentProcessFileName = Path.GetFileNameWithoutExtension( currentProcessPath );
+		string? currentProcessExtention = Path.GetExtension( currentProcessPath );
+		string backupPath = $"{currentProcessFileName}v{localVersion}{currentProcessExtention}";
```
gord5500, line 43: I don't think the cleanup function will find these now

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,136 @@
...
+
+			string newExecutablePath = Path.Combine(
+				extractPath!,
+				Path.GetFileName( currentProcessPath )!
+			);
+			File.Move( newExecutablePath, currentProcessPath, overwrite: true );
+		} catch( Exception ex ) {
+			if (currentProcessPath != "C:/bin") {
+				File.Move( backupPath, currentProcessPath );
+			}
```
gord5500, line 72:
```suggestion
```

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,136 @@
...
+		var latestVersion = new Version( UpdateChecker.GetLatestReleaseVersion( releaseData ) );
+		if ( latestVersion <= localVersion ) {
+			Console.WriteLine( $"Already own the latest version {latestVersion}" );
+			return;
+		}
+		string archiveName = GetOSFileName();
+		string downloadUrl = releaseData?.Assets?.FirstOrDefault( a => a.Name == archiveName )?.BrowserDownloadUrl
+			?? string.Empty;
+		if( string.IsNullOrWhiteSpace( downloadUrl ) ) {
+			return;
```
gord5500, line 23: returning means we're cancelling the update and the user won't know what happened if we just kill the run

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,136 @@
...
+	public static void Cleanup() {
+		string processDirectory = Path.GetDirectoryName( Environment.ProcessPath ) ?? string.Empty;
+		if( string.IsNullOrEmpty( processDirectory ) ) {
+			return;
+		}
+
+		Console.WriteLine( $"Cleaning up old binaries in {processDirectory}" );
+		foreach( string file in Directory.GetFiles( processDirectory, "*.old.bak" ) ) {
+			try {
+				Console.WriteLine( $"Cleaning up {file}" );
```
gord5500, line 99:
```suggestion
```

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}
+
+	public static void Cleanup() {
+		string processDirectory = Path.GetDirectoryName( Environment.ProcessPath ) ?? string.Empty;
+		if( string.IsNullOrEmpty( processDirectory ) ) {
+			return;
+		}
+
+		Console.WriteLine( $"Cleaning up old binaries in {processDirectory}" );
```
gord5500, line 96:
```suggestion
```

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,136 @@
...
+		try {
+			string extension = Path.GetExtension( downloadUrl ).ToLowerInvariant();
+			string? extractPath = Path.GetDirectoryName( currentProcessPath );
+
+			if( extension.Equals( ".zip" ) ) {
+				DecompressZipFile( downloadPath, extractPath! );
+			} else if( extension.Equals( ".gz" ) ) {
+				DecompressTarGzipFile( downloadPath, extractPath! );
+			} else {
+				Console.WriteLine( extension );
```
gord5500, line 60:
```suggestion
```

src/D2L.Bmx/GithubRelease.cs
```diff
@@ -0,0 +1,21 @@
...
+	public string? TagName { get; set; }
+
+	[JsonPropertyName( "assets" )]
+	public List<Assets>? Assets { get; set; }
+}
+
+internal record Assets {
+	[JsonPropertyName( "url" )]
+	public string? Url { get; set; }
+
```
gord5500, line 15:
```suggestion
```

Don't think this is getting used

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		try {
+			TarFile.ExtractToDirectory( tarPath, decompressedFilePath, true );
+		} finally {
+			File.Delete( tarPath );
+		}
+	}
+
+	private static void DecompressZipFile( string compressedFilePath, string decompressedFilePath ) {
```
gord5500, line 120:
```suggestion
	private static void ExtractZipFile( string compressedFilePath, string decompressedFilePath ) {
```

Wonder if `Extract` makes more sense here

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+		string archiveName = GetOSFileName();
+		string downloadUrl = releaseData?.Assets?.FirstOrDefault( a => a.Name == archiveName )?.BrowserDownloadUrl
+			?? string.Empty;
+		if( string.IsNullOrWhiteSpace( downloadUrl ) ) {
+			throw new BmxException( "Failed to get update download url" );
+		}
+
+		string? downloadPath = Path.GetTempFileName();
+
+		string? currentProcessPath = Environment.ProcessPath;
```
gord5500, line 28:
```suggestion
		string? currentProcessPath = Path.GetDirectoryName(Process.GetCurrentProcess().MainModule.FileName);
```

Looks like `Environment.ProcessPath` isn't what we want and it is placing the old outdated file in the current directory your shell has

gord5500: Which means the old versions are going to be placed in probably the wrong spot and won't get cleaned up

gord5500: `string baseDirectory = AppContext.BaseDirectory;` can get you the folder where bmx is. `Process.GetCurrentProcess().MainModule!.FileName` can get you the executable name

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+		string? currentProcessFileName = Path.GetFileNameWithoutExtension( currentProcessPath );
+		string backupPath = $"{currentProcessFileName}v{localVersion}.old.bak";
+
+		if( !string.IsNullOrEmpty( currentProcessPath ) ) {
+			File.Move( currentProcessPath, backupPath );
```
gord5500, line 33: Am getting an IO exception with permission denied message here on Mac. We should catch that and if so prompt the user to elevate permission

Also instead of creating the backup file in the existing directory we should maybe put it somewhere under .bmx. After the update happens and the user runs bmx again, it would try to clear out the backup file in /bin/ again which would then fail as it requires sudo to delete

src/D2L.Bmx/UpdateHandler.cs
gord5500: I made a branch [here](https://github.com/Brightspace/bmx/commit/86fdf104ecb4513e95079db5df6bb516754c3eee) trying it out

alex-fang0: Strange? I thought Environment.ProcessPath is fine to grab the current file but we can place the old outdated version in another folder and modify the cleanup for that aswell

gord5500: Here's what you can do to reproduce:
- Build bmx and replace the executable C:/bin
- Open powershell and pick any folder
- type `bmx update` and you will find that the bmx.*.old.bak is in that folder you chose above

gord5500: But yea moving to .bmx on the rename should fix it

alex-fang0: Ah gotcha thanks for the help

src/D2L.Bmx/UpdateHandler.cs
alex-fang0: Just made some changes to prompt on IOexceptions! Should catch it but I don't have a mac to test

src/D2L.Bmx/GithubRelease.cs
```diff
@@ -0,0 +1,42 @@
...
+		httpClient.DefaultRequestHeaders.Add( "User-Agent", "BMX" );
+		var response = await httpClient.GetAsync( "repos/Brightspace/bmx/releases/latest" );
+		response.EnsureSuccessStatusCode();
+
+		await using var responseStream = await response.Content.ReadAsStreamAsync();
+		var releaseData = await JsonSerializer.DeserializeAsync(
+			responseStream,
+			SourceGenerationContext.Default.GithubRelease
+		);
+		return releaseData;
```
cfbao, line 40: use [GetFromJsonAsync](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.httpclientjsonextensions.getfromjsonasync?view=net-8.0) to simplify this

```suggestion
		return await httpClient.GetFromJsonAsync(
			"repos/Brightspace/bmx/releases/latest",
			SourceGenerationContext.Default.GithubRelease );
```

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -49,28 +50,7 @@ internal static class UpdateChecker {
...
-		);
-		string version = releaseData?.TagName?.TrimStart( 'v' ) ?? string.Empty;
-		SaveLatestVersion( version );
-		return version;
-	}
-
-	private static void SaveLatestVersion( string version ) {
-		if( string.IsNullOrWhiteSpace( version ) ) {
-			return;
-		}
```
cfbao, line 73: why do we not need to check empty string now?
I mean... I'd be happy to get rid of it if we really don't fall back to empty strings.

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+using System.IO.Compression;
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace D2L.Bmx;
+
+internal class UpdateHandler {
+
+	public async Task HandleAsync() {
+		using var httpClient = new HttpClient();
```
cfbao, line 11: We probably have enough GitHub interaction code to warrant a `GitHubClient` class. Then, ideally, I'd inject that into dependent classes via constructor parameters, and avoid instantiating new clients directly.
DI is generally nice (for testing and dependency/behaviour tracking), and consistent with our general coding practice.
I'm willing to let this PR through without DI for now, but I would like us to refactor this soon-ish.

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+		var releaseData = await GithubUtilities.GetLatestReleaseDataAsync();
+		var localVersion = Assembly.GetExecutingAssembly().GetName().Version;
+		var latestVersion = new Version( GithubUtilities.GetReleaseVersion( releaseData ) );
+		if( latestVersion <= localVersion ) {
+			Console.WriteLine( $"Already own the latest version {latestVersion}" );
+			return;
+		}
+		string archiveName = GetOSFileName();
+		string downloadUrl = releaseData?.Assets?.FirstOrDefault( a => a.Name == archiveName )?.BrowserDownloadUrl
+			?? string.Empty;
```
cfbao, line 21: we check null or whitespace below, so there's no reason at all to fall back to empty string here
```suggestion
		string? downloadUrl = releaseData?.Assets?.FirstOrDefault( a => a.Name == archiveName )?.BrowserDownloadUrl;
```

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+			}
+		} catch( Exception ex ) {
+			File.Move( backupPath, currentFilePath );
+			throw new BmxException( "Failed to download the update", ex );
+		}
+
+		try {
+			string extension = Path.GetExtension( downloadUrl ).ToLowerInvariant();
+
+			if( extension.Equals( ".zip" ) ) {
```
cfbao, line 58: instead of calling `ToLowerInvariant`, use [Equals(String, StringComparison)](https://learn.microsoft.com/en-us/dotnet/api/system.string.equals?view=net-8.0#system-string-equals(system-string-system-stringcomparison)) with `StringComparison.OrdinalIgnoreCase`

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+	private static string GetOSFileName() {
+
+		if( RuntimeInformation.IsOSPlatform( OSPlatform.Windows ) ) {
+			return "bmx-win-x64.zip";
+		} else if( RuntimeInformation.IsOSPlatform( OSPlatform.OSX ) ) {
+			return "bmx-osx-x64.tar.gz";
+		} else if( RuntimeInformation.IsOSPlatform( OSPlatform.Linux ) ) {
+			return "bmx-linux-x64.tar.gz";
+		} else {
+			throw new Exception( "Unknown OS" );
```
cfbao, line 82: please use BmxException with a better error message.

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+
+	public static void Cleanup() {
+		string processDirectory = Path.GetDirectoryName( BmxPaths.OLD_BMX_VERSION_FILE_NAME ) ?? string.Empty;
+		if( string.IsNullOrEmpty( processDirectory ) ) {
+			return;
+		}
+		foreach( string file in Directory.GetFiles( processDirectory, "*.old.bak" ) ) {
+			try {
+				File.Delete( file );
+			} catch( Exception ex ) {
```
cfbao, line 94: as per build warning
```suggestion
			} catch( Exception ) {
```

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+			throw new BmxException( "BMX could not update" );
+		}
+
+		try {
+			var archiveRes = await httpClient.GetAsync( downloadUrl, HttpCompletionOption.ResponseHeadersRead );
+			using( var fs = new FileStream( downloadPath, FileMode.Create, FileAccess.Write, FileShare.None ) ) {
+				await archiveRes.Content.CopyToAsync( fs );
+				await fs.FlushAsync();
+				fs.Dispose();
+			}
```
cfbao, line 49: * don't call dispose when you're already in scope of a using
* prefer [using declaration](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/using#using-declaration) to reduce indentation wherever possible. same here: https://github.com/Brightspace/bmx/pull/418/files#diff-48acde0deb16dbeda7f36eb7087c301cf5a486a4fe2044fc77522fdd67191353R121

```suggestion
			using var fs = new FileStream( downloadPath, FileMode.Create, FileAccess.Write, FileShare.None );
			await archiveRes.Content.CopyToAsync( fs );
			await fs.FlushAsync();
```

src/D2L.Bmx/BmxPaths.cs
```diff
@@ -11,4 +11,5 @@ internal static class BmxPaths {
 	public static readonly string SESSIONS_FILE_LEGACY_NAME = Path.Join( BMX_DIR, "sessions" );
 	public static readonly string UPDATE_CHECK_FILE_NAME = Path.Join( CACHE_DIR, "updateCheck" );
 	public static readonly string AWS_CREDS_CACHE_FILE_NAME = Path.Join( CACHE_DIR, "awsCreds" );
+	public static readonly string OLD_BMX_VERSION_FILE_NAME = Path.Join( BMX_DIR, "bmx" );
```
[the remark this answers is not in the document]
alex-fang0: Yea that makes sense! Added a temp folder for the old versions and millisecond timestamp for the file.

src/D2L.Bmx/GithubRelease.cs
```diff
@@ -0,0 +1,42 @@
...
+internal record GithubAsset {
+	[JsonPropertyName( "name" )]
+	public string? Name { get; init; }
+
+	[JsonPropertyName( "browser_download_url" )]
+	public string? BrowserDownloadUrl { get; init; }
+}
+
+internal static class GithubUtilities {
+	public static string GetReleaseVersion( GithubRelease? releaseData ) {
```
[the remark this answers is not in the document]
alex-fang0: Makes sense that we have consistency and clarity on returning nulls. Made some changes to put the methods into the record itself

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,129 @@
...
+		using var httpClient = new HttpClient();
+		GithubRelease? releaseData = await GithubRelease.GetLatestReleaseDataAsync();
+		Version? latestVersion = releaseData?.GetReleaseVersion();
+		if( latestVersion is null ) {
+			throw new BmxException( "Failed to grab the latest version" );
+		}
+
+		var localVersion = Assembly.GetExecutingAssembly().GetName().Version;
+		if( latestVersion <= localVersion ) {
+			Console.WriteLine( $"Already own the latest version {latestVersion}" );
```
cfbao, line 20: nit
```suggestion
			Console.WriteLine( $"You already have the latest version {latestVersion}" );
```

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,129 @@
...
+			return;
+		}
+
+		string archiveName = GetOSFileName();
+		string? downloadUrl = releaseData?.Assets?.FirstOrDefault( a => a.Name == archiveName )?.BrowserDownloadUrl;
+		if( string.IsNullOrWhiteSpace( downloadUrl ) ) {
+			throw new BmxException( "Failed to get update download url" );
```
cfbao, line 27:
```suggestion
			throw new BmxException( "Failed to find the download URL of the latest BMX" );
```

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,129 @@
...
+			throw new BmxException( "BMX could not update" );
+		}
+
+		string currentDirectory = Path.GetDirectoryName( currentFilePath )!;
+		long backupPathTimeStamp = DateTime.Now.Millisecond;
+		string backupPath = Path.Join( BmxPaths.OLD_BMX_VERSIONS_PATH, $"bmx-v{localVersion}-{backupPathTimeStamp}-old.bak" );
+		try {
+			File.Move( currentFilePath, backupPath );
+		} catch( IOException ex ) {
+			throw new BmxException( "Could move the old version, try with elevated permissions", ex );
```
cfbao, line 41: IIUC
```suggestion
			throw new BmxException( "Could not remove the old version. Please try again with elevated permissions.", ex );
```

I know it's technically a move not a delete, but I think saying saying "move" would be more confusing to users not familiar with our implementation details.

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,129 @@
...
+
+
+		string downloadPath = Path.GetTempFileName();
+		try {
+			var archiveRes = await httpClient.GetAsync( downloadUrl, HttpCompletionOption.ResponseHeadersRead );
+			using var fs = new FileStream( downloadPath, FileMode.Create, FileAccess.Write, FileShare.None );
+			await archiveRes.Content.CopyToAsync( fs );
+			await fs.FlushAsync();
+		} catch( Exception ex ) {
+			File.Move( backupPath, currentFilePath );
```
cfbao, line 54: Instead of moving the current file to backup path first then moving it back on error, how about downloading and extracting the new file into temp directory first, then do all the moves afterwards?

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,129 @@
...
+
+		try {
+			string extension = Path.GetExtension( downloadUrl );
+
+			if( extension.Equals( ".zip", StringComparison.OrdinalIgnoreCase ) ) {
+				ExtractZipFile( downloadPath, currentDirectory );
+			} else if( extension.Equals( ".gz", StringComparison.OrdinalIgnoreCase ) ) {
+				ExtractTarGzipFile( downloadPath, currentDirectory );
+			} else {
+				throw new Exception( "Unknown archive type" );
```
cfbao, line 66: use BmxException with better error message

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	public static void Cleanup() {
+		if( !Directory.Exists( BmxPaths.OLD_BMX_VERSIONS_PATH ) ) {
+			try {
+				Directory.CreateDirectory( BmxPaths.OLD_BMX_VERSIONS_PATH );
+			} catch( Exception ex ) {
+				throw new BmxException( "Failed to initialize temporary BMX file directory (~/.bmx/temp)", ex );
+			}
+		}
```
cfbao, line 96: I don't like that we're creating a directory in a function named "Cleanup", and that our update logic relies on this cleanup function running first:
https://github.com/Brightspace/bmx/pull/418/files#diff-48acde0deb16dbeda7f36eb7087c301cf5a486a4fe2044fc77522fdd67191353R37-R39

Also not a fan of always creating the temp directory even when it's usually not needed, but this isn't a deal breaker. If we do want to always ensure the existence of this temp directory, we should do this together with other directory init logic here:
https://github.com/Brightspace/bmx/blob/8c28c21d264da16e15b8af636b58e14dde192cf5/src/D2L.Bmx/Program.cs#L215-L227

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,129 @@
...
+		string processDirectory = BmxPaths.OLD_BMX_VERSIONS_PATH;
+		foreach( string file in Directory.GetFiles( processDirectory, "*old.bak" ) ) {
+			try {
+				File.Delete( file );
```
cfbao, line 100: I think we can just delete everything inside this "temp" directory.

src/D2L.Bmx/UpdateHandler.cs
alex-fang0: Gotcha! How about creating a new directory every update call and then deleting the directory if it exists in the clean up so the directory itself doesn't exist when it is not needed? Would that be an overkill of creation and deletions of directories?

cfbao: I think that makes sense. easier cleanup should be one of the benefits of using a subdirectory

src/D2L.Bmx/UpdateHandler.cs
alex-fang0: I think for the TarFile extract we don't set or return the file name so I think it might be a bit complicated keeping track of which files to move in the directory if there are multiple updates running.

cfbao: Sorry, when I said "temp directory" I was actually thinking of a uniquely named subdirectory in the system temp directory, i.e. what's returned by [`Path.GetTempFileName`](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename).

Even if we use `~/.bmx/temp`, we can also create more subdirectories in it with unique names, e.g.
```
~/.bmx/temp/58ba67a7-10f3-418c-a686-90894c2b6045
~/.bmx/temp/b90477fc-469b-4651-8ace-0ab522c935c8
...
```
we can generate and keep track of the UUID (or other types of unique identifiers e.g. timestamps?) in memory, and not worry about multiple runs interfering with each other.

src/D2L.Bmx/UpdateHandler.cs
alex-fang0: This is already a nested exception _(not sure if we can un-nest it without making the block messy_) so I don't think the user will see it. I think the "Failed to update with new files" message should capture this tho

src/D2L.Bmx/UpdateHandler.cs
alex-fang0: I made a commit trying out the temp directory first to push moving files to the end, seems a bit messy so I'm not sure if we should go for this. We can revert if needed

src/D2L.Bmx/UpdateHandler.cs
cfbao: ah I see. Didn't notice the nesting. I'd still like to avoid it, but it's good for now.

src/D2L.Bmx/UpdateHandler.cs
cfbao: The new change looks good. We can make it cleaner, but even as is it doesn't look too messy to me.
I do vastly prefer the new approach of "not moving files until we have everything ready".

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,156 @@
...
+
+			if( extension.Equals( ".zip", StringComparison.OrdinalIgnoreCase ) ) {
+				ExtractZipFile( downloadPath, extractFolder );
+			} else if( extension.Equals( ".gz", StringComparison.OrdinalIgnoreCase ) ) {
+				ExtractTarGzipFile( downloadPath, extractFolder );
+			} else {
+				throw new Exception( "Unknown archive type" );
+			}
+		} catch( Exception ex ) {
+			Directory.Delete( extractFolder, true );
```
cfbao, line 75: Boolean literals should almost always be named explicitly
```suggestion
			Directory.Delete( extractFolder, recursive: true );
```
similarly for other `Directory.Delete` calls

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -0,0 +1,156 @@
...
+		} finally {
+			File.Delete( tarPath );
+		}
+	}
+
+	private static void ExtractZipFile( string compressedFilePath, string decompressedFilePath ) {
+		using ZipArchive archive = ZipFile.OpenRead( compressedFilePath );
+		foreach( ZipArchiveEntry entry in archive.Entries ) {
+			string destinationPath = Path.GetFullPath( Path.Combine( decompressedFilePath!, entry.FullName ) );
+			if( destinationPath.StartsWith( decompressedFilePath!, StringComparison.Ordinal ) ) {
```
cfbao, line 151: `decompressedFilePath` isn't nullable
```suggestion
			string destinationPath = Path.GetFullPath( Path.Combine( decompressedFilePath, entry.FullName ) );
			if( destinationPath.StartsWith( decompressedFilePath, StringComparison.Ordinal ) ) {
```

cfbao (dismissed):

cfbao (approved):

PR #435, opened 2024-01-03
tweak update check & self-update

cfbao:
Partly code refactor/nits, partly behavioural changes:

* Use `~/.bmx/temp` exclusively as a scratch space and don't use system temp folder. Rationale is that we have cleanup logic for the former, but not the latter.
* Each BMX run uses an exclusive subdirectory under `~/.bmx/temp` to avoid any conflicts, even if cleanup fails
* Respects the filename of the currently running BMX executable during self-update
* The message from auto update check now directs users to run `bmx update` instead of going to GitHub or bmx.d2l.dev
* Print informational messages to console during the update process

### DIFF
--- a/src/D2L.Bmx/Program.cs
+++ b/src/D2L.Bmx/Program.cs
@@ -233,10 +233,16 @@ return await new CommandLineBuilder( rootCommand )
 					throw new BmxException( "Failed to initialize BMX directory (~/.bmx)", ex );
 				}
 			}
+			// clean up temp files from earlier runs
+			if( Directory.Exists( BmxPaths.TEMP_DIR ) ) {
+				try {
+					Directory.Delete( BmxPaths.TEMP_DIR, recursive: true );
+				} catch( Exception ) {
+					// ignore errors as this doesn't impact normal BMX usage
+				}
+			}
 
-			UpdateHandler.Cleanup();
-
-			var updateChecker = new UpdateChecker( github, config );
+			var updateChecker = new UpdateChecker( github );
 			await updateChecker.CheckForUpdatesAsync();
 
 			await next( context );
--- a/src/D2L.Bmx/UpdateChecker.cs
+++ b/src/D2L.Bmx/UpdateChecker.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text.Json;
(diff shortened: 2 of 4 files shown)


src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -1,34 +1,36 @@
...
 using System.Text.Json;
 using D2L.Bmx.GitHub;
 
 namespace D2L.Bmx;
 
-internal class UpdateChecker( IGitHubClient github, BmxConfig config ) {
+internal class UpdateChecker( IGitHubClient github ) {
 	public async Task CheckForUpdatesAsync() {
 		try {
-			var cachedVersion = GetUpdateCheckCache();
```
cfbao, line 11: The name `cachedVersion` suggests a `Version` object or a string, but it's actually an `UpdateCheckCache`...

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -1,34 +1,36 @@
...
-			var localVersion = Assembly.GetExecutingAssembly().GetName().Version;
-			var latestVersion = new Version( cachedVersion?.VersionName ?? "0.0.0" );
```
cfbao, line 13: don't use default or fallback values if not necessary

src/D2L.Bmx/UpdateChecker.cs
```diff
@@ -89,19 +91,18 @@ internal class UpdateChecker( IGitHubClient github, BmxConfig config ) {
...
+		return cache?.VersionName is null
+			|| cache.TimeLastChecked is null
+			|| DateTimeOffset.UtcNow - cache.TimeLastChecked > TimeSpan.FromDays( 1 )
+			|| cache.TimeLastChecked > DateTimeOffset.UtcNow;
 	}
 }
 
 
-internal record UpdateCheckCache {
-	public string? VersionName { get; set; }
```
cfbao, line 105: * These properties should not be settable
* `Version` can be deserialized to from JSON directly.

src/D2L.Bmx/Program.cs
```diff
@@ -233,10 +233,16 @@
...
+			// clean up temp files from earlier runs
+			if( Directory.Exists( BmxPaths.TEMP_DIR ) ) {
+				try {
+					Directory.Delete( BmxPaths.TEMP_DIR, recursive: true );
+				} catch( Exception ) {
+					// ignore errors as this doesn't impact normal BMX usage
+				}
+			}
 
-			UpdateHandler.Cleanup();
```
cfbao, line 237: inlining the logic here, so we can see all directory init logic in one place (in this middleware)

src/D2L.Bmx/UpdateChecker.cs
boarnoah: How did you figure that out re: Version, at least as far as I can tell VSCode, Resharper etc.. don't inform of the fact it has a Serializable annotation.

cfbao: types don't need serializable annotation to be JSON serializable.
also, easy enough just to try with https://github.com/waf/CSharpRepl
image

boarnoah (approved):

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -8,115 +8,107 @@ namespace D2L.Bmx;
...
-			File.Move( backupPath, currentFilePath );
-			throw new BmxException( "BMX could not update with the new version", ex );
-		} finally {
-			Directory.Delete( extractFolder, recursive: true );
+			string errorMessage = "Failed to update with the new version.";
+			try {
+				File.Move( backupPath, currentFilePath );
+			} catch {
+				errorMessage += $"""
+
```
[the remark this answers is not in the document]
cfbao: to add a break line.
otherwise it'd directly continue from the end of the original message without a new line.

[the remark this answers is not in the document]
cfbao: the raw string literal
```csharp
"""
abc
"""
```
is literally "abc" without any newline at the beginning or the end.

PR #436, opened 2024-01-04
disable auto update check when running the update command

cfbao:
It's weird to say "you can update now!" when the user is literally updating.

### DIFF
--- a/src/D2L.Bmx/Program.cs
+++ b/src/D2L.Bmx/Program.cs
@@ -10,11 +10,6 @@ using D2L.Bmx.GitHub;
 using D2L.Bmx.Okta;
 using IniParser;
 
-// pre-create objects used by all commands and/or the middleware
-var configProvider = new BmxConfigProvider( new FileIniDataParser() );
-var config = configProvider.GetConfiguration();
-var github = new GitHubClient();
-
 // common options
 var orgOption = new Option<string>(
 	name: "--org",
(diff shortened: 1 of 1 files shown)


src/D2L.Bmx/Program.cs
```diff
@@ -10,11 +10,6 @@
 using D2L.Bmx.Okta;
 using IniParser;
 
-// pre-create objects used by all commands and/or the middleware
-var configProvider = new BmxConfigProvider( new FileIniDataParser() );
-var config = configProvider.GetConfiguration();
-var github = new GitHubClient();
```
cfbao, line 16: Loosely related - with the recent changes, these objects are no longer needed for all BMX runs, and creating them here doesn't result in better sharing, but instead may result in unnecessary actions that slow down startup (e.g. reading config on a `bmx --version` is unnecessary)

So now I'm inlining them into the respective command handler and middleware instead.

src/D2L.Bmx/Program.cs
```diff
@@ -242,8 +240,10 @@
 				}
 			}
 
-			var updateChecker = new UpdateChecker( github );
-			await updateChecker.CheckForUpdatesAsync();
+			if( context.ParseResult.CommandResult.Command != updateCommand ) {
+				var updateChecker = new UpdateChecker( new GitHubClient() );
+				await updateChecker.CheckForUpdatesAsync();
+			}
```
cfbao, line 246: this is the real relevant change

gord5500 (approved):

boarnoah (approved):

PR #437, opened 2024-01-04
display current version more accurately

cfbao:
This should rarely if ever happen for regular users, but...
if someone was using a preview release (later than "latest") and ran "bmx update", they would've seen something like:

> You already have the latest version 3.0.0.0

Problems:
* "3.0.0.0" is the "latest" release as indicated by GitHub, not the actual version the user is using, which should look like "3.1.0-preview"
* We never display the four part version format anywhere else.

This change makes this message displays the actual version the user currently has, and uses the "informational version" to align with what users would see with "bmx --version".

### DIFF
--- a/src/D2L.Bmx/Program.cs
+++ b/src/D2L.Bmx/Program.cs
@@ -199,7 +199,7 @@ writeCommand.SetHandler( ( InvocationContext context ) => {
 
 var updateCommand = new Command( "update", "Updates BMX to the latest version" );
 updateCommand.SetHandler( ( InvocationContext context ) => {
-	var handler = new UpdateHandler( new GitHubClient() );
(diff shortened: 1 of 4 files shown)


src/D2L.Bmx/Program.cs
```diff
@@ -199,7 +199,7 @@
 
 var updateCommand = new Command( "update", "Updates BMX to the latest version" );
 updateCommand.SetHandler( ( InvocationContext context ) => {
-	var handler = new UpdateHandler( new GitHubClient() );
+	var handler = new UpdateHandler( new GitHubClient(), new VersionProvider() );
```
cfbao, line 202: use DI as this provides "ambient" data that typically would be mocked in unit tests (if we ever add them)

jharoutuniand2l (approved):

PR #438, opened 2024-01-11
use system temporary directory for update

cfbao:
avoid creating arbitrary directories in unexpected places on users' computers
https://d2l.slack.com/archives/G4P099831/p1704992370878349

### DIFF
--- a/src/D2L.Bmx/UpdateHandler.cs
+++ b/src/D2L.Bmx/UpdateHandler.cs
@@ -7,15 +7,6 @@ namespace D2L.Bmx;
 
 internal class UpdateHandler( IGitHubClient github, IVersionProvider versionProvider ) {
 	public async Task HandleAsync() {
-		string workspaceDir = Path.Join( BmxPaths.TEMP_DIR, Path.GetRandomFileName() );
-		try {
-			Directory.CreateDirectory( workspaceDir );
-		} catch( Exception ex ) {
-			throw new BmxException( "Failed to create temporary BMX directory in ~/.bmx/temp", ex );
-		}
-
-		var bmxFileInfo = GetFileInfo();
-
 		Console.WriteLine( "Finding the latest BMX release..." );
 		GitHubRelease latestRelease;
 		try {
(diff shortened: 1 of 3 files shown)


src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -7,15 +7,6 @@ namespace D2L.Bmx;
...
 internal class UpdateHandler( IGitHubClient github, IVersionProvider versionProvider ) {
 	public async Task HandleAsync() {
-		string workspaceDir = Path.Join( BmxPaths.TEMP_DIR, Path.GetRandomFileName() );
-		try {
-			Directory.CreateDirectory( workspaceDir );
-		} catch( Exception ex ) {
-			throw new BmxException( "Failed to create temporary BMX directory in ~/.bmx/temp", ex );
-		}
-
-		var bmxFileInfo = GetFileInfo();
```
cfbao, line 17: moved below to be closer to where they're actually needed

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -61,7 +61,7 @@ internal class UpdateHandler( IGitHubClient github, IVersionProvider versionProv
 		Console.WriteLine( "Replacing the currently running BMX executable..." );
 		string currentFilePath = Environment.ProcessPath
 			?? throw new BmxException( "Failed to locate the current BMX executable." );
-		string backupPath = Path.Join( workspaceDir, $"bmx-v{localVersion}-old.bak" );
+		string backupPath = Path.Join( workspaceDir, $"bmx-{displayVersion}-old.bak" );
```
cfbao, line 64: this is also something that the user sees, so it should use the "display version"

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -87,6 +87,13 @@ internal class UpdateHandler( IGitHubClient github, IVersionProvider versionProv
 		}
 
 		Console.WriteLine( $"BMX updated to v{latestVersion} successfully!" );
+
+		try {
+			Directory.Delete( workspaceDir, recursive: true );
+		} catch {
+			// Ignore any errors from clean-up.
+			// This just isn't always possible (e.g. when on Windows), and it's harmless anyway.
+		}
```
cfbao, line 96: this will delete as much stuff as it can. on Windows it should be everything except the old executable. on Mac & Linux it should be everything.

src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -87,6 +87,13 @@ internal class UpdateHandler( IGitHubClient github, IVersionProvider versionProv
 		}
 
 		Console.WriteLine( $"BMX updated to v{latestVersion} successfully!" );
+
+		try {
+			Directory.Delete( workspaceDir, recursive: true );
```
boarnoah, line 92: Is there any point in deleting stuff moved to ``/tmp`` explicitly like this? Most distros have some sort of cleanup that generally takes care of this.

cfbao: At least Windows doesn't auto delete temp files by default.

gord5500 (approved):

PR #439, opened 2024-01-12
check UnauthorizedAccessException for permission error

cfbao:


src/D2L.Bmx/UpdateHandler.cs
```diff
@@ -64,8 +64,9 @@ internal class UpdateHandler( IGitHubClient github, IVersionProvider versionProv
 		string backupPath = Path.Join( workspaceDir, $"bmx-v{localVersion}-old.bak" );
 		try {
 			File.Move( currentFilePath, backupPath );
-		} catch( IOException ex ) {
-			throw new BmxException( "Could not remove the old version. Please try again with elevated permissions.", ex );
+		} catch( UnauthorizedAccessException ex ) {
+			throw new BmxException(
+				"Permission denied when removing the old version. Please try again with elevated permissions.", ex );
```
cfbao, line 69: @gord5500 I know you said you saw an IOException on Mac, but on WSL I'm seeing an `UnauthorizedAccessException` with `IOException` as an inner exception, and the old code wasn't catching the permission issue. Wonder if you misread the top level exception?
Can check again when you test it on Mac.

cfbao: `UnauthorizedAccessException` also just makes sense...

gord5500: I think you are right about that. I will make sure to check when I test

tmanti (approved):

src/D2L.Bmx/UpdateHandler.cs
jharoutuniand2l: Instead of saying "old version" (or in addition to), could we include the file path in the message? That way the user can just delete it in whichever way they see fit.

scowing (approved):

src/D2L.Bmx/UpdateHandler.cs
cfbao: Not a bad idea to include the file path for general troubleshooting. I'll update in another PR.

PR #442, opened 2024-01-15
allow interactive prompts when using cached creds

cfbao:
In my testing, I find the strict behaviour when using `--cache` to be counter intuitive and unnecessarily harsh.
It also makes extending code more difficult.

More here: https://github.com/Brightspace/bmx/pull/442#discussion_r1452696462

### DIFF
--- a/src/D2L.Bmx/AwsCredsCreator.cs
+++ b/src/D2L.Bmx/AwsCredsCreator.cs
@@ -23,16 +23,12 @@ internal class AwsCredsCreator(
 			);
 		}
 
-		if( string.IsNullOrEmpty( account ) ) {
-			if( !string.IsNullOrEmpty( config.Account ) ) {
-				account = config.Account;
-			}
+		if( string.IsNullOrEmpty( account ) && !string.IsNullOrEmpty( config.Account ) ) {
+			account = config.Account;
 		}
 
-		if( string.IsNullOrEmpty( role ) ) {
-			if( !string.IsNullOrEmpty( config.Role ) ) {
-				role = config.Role;
-			}
+		if( string.IsNullOrEmpty( role ) && !string.IsNullOrEmpty( config.Role ) ) {
+			role = config.Role;
 		}
 
 		if( duration is null or 0 ) {
@@ -43,11 +39,8 @@ internal class AwsCredsCreator(
 			}
 		}
 
-		if( cache ) {
-			if( string.IsNullOrEmpty( account ) || string.IsNullOrEmpty( role ) ) {
-				throw new BmxException( "Account & Role must be provided when using cached AWS credentials" );
-			}
-
+		// if using cache, avoid calling Okta at all if possible
+		if( cache && !string.IsNullOrEmpty( account ) && !string.IsNullOrEmpty( role ) ) {
 			var cachedCredentials = awsCredentialCache.GetCredentials(
 				org: oktaApi.Org,
 				user: oktaApi.User,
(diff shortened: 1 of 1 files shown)


src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -43,11 +39,8 @@ bool cache
 			}
 		}
 
-		if( cache ) {
-			if( string.IsNullOrEmpty( account ) || string.IsNullOrEmpty( role ) ) {
-				throw new BmxException( "Account & Role must be provided when using cached AWS credentials" );
-			}
-
+		// if using cache, avoid calling Okta at all if possible
+		if( cache && !string.IsNullOrEmpty( account ) && !string.IsNullOrEmpty( role ) ) {
```
cfbao, line 43: Real change part 1 of 2:

Don't fail the run if account & role aren't provided via config or CLI.

* In my testing, I find erroring out here to be counter intuitive and unnecessarily harsh. It's perfectly fine if users want to get the cached creds early in an interactive run (which will call Okta), and later utilize the cached creds in a script (where account & role are provided via flags and Okta won't be called).
* Erroring out here kinda mixes the purpose of the `cache` and `nonInteractive` parameters of this method, which makes extending this method to work with writing credential process command ([this propsal](https://d2l.slack.com/archives/G4P099831/p1705333949277619)) more difficult.

src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -83,12 +74,26 @@ bool cache
...
-			} else {
+			if( nonInteractive ) {
 				throw new BmxException( "Role value was not provided" );
 			}
+			string[] roles = rolesData.Select( r => r.RoleName ).ToArray();
+			role = consolePrompter.PromptRole( roles );
+		}
+
+		// try getting from cache again even if calling Okta is inevitable (we still avoid the AWS call)
+		if( cache ) {
```
cfbao, line 85: Real change part 2 of 2.

src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -23,16 +23,12 @@ bool cache
 			);
 		}
 
-		if( string.IsNullOrEmpty( account ) ) {
-			if( !string.IsNullOrEmpty( config.Account ) ) {
-				account = config.Account;
-			}
+		if( string.IsNullOrEmpty( account ) && !string.IsNullOrEmpty( config.Account ) ) {
+			account = config.Account;
```
cfbao, line 27: Unrelated nit: combine `if`s to reduce nesting

src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -61,16 +54,14 @@ bool cache
...
-		//cache this one?
 		OktaApp[] awsApps = await oktaApi.Api.GetAwsAccountAppsAsync();
 
 		if( string.IsNullOrEmpty( account ) ) {
-			if( !nonInteractive ) {
-				string[] accounts = awsApps.Select( app => app.Label ).ToArray();
-				account = consolePrompter.PromptAccount( accounts );
-			} else {
+			if( nonInteractive ) {
 				throw new BmxException( "Account value was not provided" );
```
cfbao, line 61: Unrelated nit: revert condition to early exit & reduce nesting.

src/D2L.Bmx/AwsCredsCreator.cs
boarnoah: Seems fine, I think  the code as originally written would have guarded against valid credentials in BMX that Okta no longer believes are valid (user unassigned from app etc...?)

That doesn't seem like something BMX would be able to care for, it could only be done authoritatively on the AWS side by revoking old sessions altogether.

boarnoah (approved):

src/D2L.Bmx/AwsCredsCreator.cs
cfbao: > I think the code as originally written would have guarded against valid credentials in BMX that Okta no longer believes are valid (user unassigned from app etc...?)

No, there's no material difference here.

If user is unassigned from the app, and
* if they provide everything upfront, we don't call Okta in both worlds, and they can use the creds just fine
* if they don't provide everything upfront
  * in the old world BMX fails directly
  * in the new world they won't see the account & role they want in the selection prompts

PR #443, opened 2024-01-15
add '--use-credential-process' to write command

cfbao:
Make it easier for people to start using BMX in `credential_process`, and reduce user errors.

https://d2l.slack.com/archives/G4P099831/p1705333949277619

### DIFF
--- a/src/D2L.Bmx/ParameterDescriptions.cs
+++ b/src/D2L.Bmx/ParameterDescriptions.cs
@@ -8,8 +8,14 @@ internal static class ParameterDescriptions {
 	public const string Role = "AWS role name";
 	public const string Duration = "AWS session duration in minutes";
 	public const string Profile = "AWS profile name";
-	public const string Output = "Custom path to the AWS credentials file";
+	public const string Output =
+		"Custom path to the AWS credentials file, or the AWS config file if '--use-credential-process' is supplied";
 	public const string Format = "Output format of AWS credentials";
 	public const string NonInteractive = "Run non-interactively without showing any prompts";
-	public const string CacheAwsCredentials = "Enables Cache for AWS tokens";
+	public const string CacheAwsCredentials =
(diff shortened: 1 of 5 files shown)


src/D2L.Bmx/ParameterDescriptions.cs
```diff
@@ -8,8 +8,14 @@ internal static class ParameterDescriptions {
...
 	public const string Duration = "AWS session duration in minutes";
 	public const string Profile = "AWS profile name";
-	public const string Output = "Custom path to the AWS credentials file";
+	public const string Output =
+		"Custom path to the AWS credentials file, or the AWS config file if '--use-credential-process' is supplied";
 	public const string Format = "Output format of AWS credentials";
 	public const string NonInteractive = "Run non-interactively without showing any prompts";
-	public const string CacheAwsCredentials = "Enables Cache for AWS tokens";
+	public const string CacheAwsCredentials =
+		"Enables Cache for AWS tokens. Implied if '--use-credential-process' is supplied";
```
boarnoah, line 16: It does make the bit about --cache a little confusing if the recommend way to start using bmx in credential process is to use ``-use-credential-process``

Should we keep the flag but just not document it, suggest it should always be used indirectly.

boarnoah (approved):

gord5500 (approved):

src/D2L.Bmx/ParameterDescriptions.cs
cfbao: I think we can omit talking about it in the README, if that's what you mean.
But I think it's worth documenting all the flags in the CLI binary's help text for power users.
e.g. we use it here: https://github.com/Brightspace/vulcan-scratch/blob/0c8e6d00229339d9108240c03bb6602fdbba9705/slims/helper_scripts/lib/Invoke-AuthTokenRestMethod.psm1#L38

PR #444, opened 2024-01-17
update README for 3.1 release

cfbao:


README.md
```diff
@@ -6,80 +6,86 @@ BMX provides API access to your AWS accounts using existing Okta credentials. In
 
 Download the appropriate binary from the [releases](https://github.com/Brightspace/bmx/releases) page. For D2Lers, visit [bmx.d2l.dev](https://bmx.d2l.dev) for installation.
 
-## Commands
 
-### configure
+## Usage
```
cfbao, line 10: I'd like to present `bmx write --use-credential-process` as a separate subsection, because I feel it's distinct and important enough. Therefore, a detailed rundown of all the commands & options isn't the most suitable format for this.
So I reorganized the doc here to present various "usage" of BMX, which I think is more what we really want to do.

gord5500 (approved):

PR #451, opened 2024-06-12
display okta user and org in password prompt / failed auth error message

gord5500:
### Why

Potential for the config file to have bad user / org values and it's not immediately obvious to the user that those could be the reason for their okta auth failing. This can happen with our internal install script when running wsl and the user gets set to 'root' in the config file.

image

### Ticket

VUL-330

### DIFF
--- a/src/D2L.Bmx/ConsolePrompter.cs
+++ b/src/D2L.Bmx/ConsolePrompter.cs
@@ -8,7 +8,7 @@ internal interface IConsolePrompter {
 	string PromptOrg( bool allowEmptyInput );
 	string PromptProfile();
 	string PromptUser( bool allowEmptyInput );
-	string PromptPassword();
+	string PromptPassword( string user, string org );
 	int? PromptDuration();
 	string PromptAccount( string[] accounts );
 	string PromptRole( string[] roles );
@@ -63,7 +63,7 @@ internal class ConsolePrompter : IConsolePrompter {
 		return user;
 	}
 
-	string IConsolePrompter.PromptPassword() {
+	string IConsolePrompter.PromptPassword( string user, string org ) {
 		Func<char> readKey;
 		if( IS_WINDOWS ) {
 			// On Windows, Console.ReadKey calls native console API, and will fail without a console attached
@@ -84,7 +84,8 @@ internal class ConsolePrompter : IConsolePrompter {
 			readKey = () => (char)_stdinReader.Read();
 		}
 
-		Console.Error.Write( $"{ParameterDescriptions.Password}: " );
+		string passwordPrompt = string.Format( ParameterDescriptions.Password, user, org );
+		Console.Error.Write( $"{passwordPrompt}: " );
 
 		string? originalTerminalSettings = null;
 		try {
--- a/src/D2L.Bmx/ConsolePrompter.cs
+++ b/src/D2L.Bmx/ConsolePrompter.cs
@@ -84,8 +84,9 @@ internal class ConsolePrompter : IConsolePrompter {
 			readKey = () => (char)_stdinReader.Read();
 		}
 
-		string passwordPrompt = string.Format( ParameterDescriptions.Password, user, org );
-		Console.Error.Write( $"{passwordPrompt}: " );
+		Console.Error.WriteLine( $"{ParameterDescriptions.Org}: {org}" );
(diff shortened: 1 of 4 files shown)


src/D2L.Bmx/ParameterDescriptions.cs
```diff
@@ -3,7 +3,7 @@ namespace D2L.Bmx;
 internal static class ParameterDescriptions {
 	public const string Org = "Okta org short name or domain name";
 	public const string User = "Okta username";
-	public const string Password = "Okta password";
+	public const string Password = "Okta password for username '{0}' in org '{1}'";
```
gord5500, line 6: The wording to include org felt a little weird to come up with

cfbao, line 6: `ParameterDescriptions.Password` should remain unchanged. It's meant to be succinct description of the "password" parameter itself.
It's also used in the `--help` message where it really doesn't make sense to show it like this (including org & username).

src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -84,7 +84,8 @@ internal class ConsolePrompter : IConsolePrompter {
 			readKey = () => (char)_stdinReader.Read();
 		}
 
-		Console.Error.Write( $"{ParameterDescriptions.Password}: " );
+		string passwordPrompt = string.Format( ParameterDescriptions.Password, user, org );
+		Console.Error.Write( $"{passwordPrompt}: " );
```
cfbao, line 88: A single line is getting long. Not super readable at a glance.

Let's do it like this
```csharp
Console.Error.Write( $"{ParameterDescriptions.Org}: {org}" );
Console.Error.Write( $"{ParameterDescriptions.User}: {user}" );
Console.Error.Write( $"{ParameterDescriptions.Password}: " );
```

This also keeps all parameter descriptions unchanged.

cfbao (changes requested):

src/D2L.Bmx/ConsolePrompter.cs
gord5500: > Let's do it like this

Sure

src/D2L.Bmx/ParameterDescriptions.cs
gord5500: Ah okay I didn't think about it showing up in `--help`

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -23,15 +23,18 @@ internal interface IOktaApi {
 internal class OktaApi : IOktaApi {
 	private readonly CookieContainer _cookieContainer;
 	private readonly HttpClient _httpClient;
+	private string organization = string.Empty;
```
cfbao, line 26: Please don't initialize things to invalid non-null values just so we can mark things as non-nullable and skip null checks.
Nullables are good. Forced null checks are good. That's the compiler helping us avoid null errors.
False non-null values are much worse.

also - we name private fields with an underscore
```suggestion
	private string? _organization;
```

src/D2L.Bmx/ParameterDescriptions.cs
cfbao: should check all usages (find references) before such a change

src/D2L.Bmx/Okta/OktaApi.cs
gord5500: Yea that's fair. I saw we couldn't reach the exception without having that value set already but I can see the reasoning

cfbao: related - https://desire2learn.atlassian.net/browse/VUL-104 this should help

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -83,7 +86,11 @@ internal class OktaApi : IOktaApi {
 				authnResponse.Embedded.Factors
 			);
 		}
-		throw new BmxException( "Okta authentication failed. Check if org, user and password is correct" );
+
+		string org = _organization is not null ? _organization : "unknown";
```
cfbao, line 90:
```suggestion
		string org = _organization ?? "unknown";
```

cfbao (approved):

cfbao (approved):

PR #453, opened 2024-06-17
write: update config file section with same name if no cred process flag

gord5500:
### Why

Default credentials file takes priority over the default config which is where we store profiles using `credential_process`. If someone had an existing profile in config file and then runs `bmx write`, creds will be stored into the default credentials file making the existing one useless. We should remove the existing key in the config file so there is no confusion.

`credential_process` is the better way for majority of people but there are some scenarios where it's not supported. So we are leaving some information to the user that it is no longer being used. Just in case they unintentionally ran write without the flag. Not going with a prompt to confirm because that could break existing scripts

### Ticket

VUL-385

### DIFF
--- a/src/D2L.Bmx/WriteHandler.cs
+++ b/src/D2L.Bmx/WriteHandler.cs
@@ -86,23 +86,23 @@ internal class WriteHandler(
 				+ $" --duration {awsCredsInfo.Duration}";
 		} else {
 			if( File.Exists( SharedCredentialsFile.DefaultConfigFilePath ) ) {
-				string profileName = $"profile {profile}";
+				string sectionName = $"profile {profile}";
 				var defaultConfigFile = parser.ReadFile( SharedCredentialsFile.DefaultConfigFilePath );
-				if( defaultConfigFile.Sections.ContainsSection( profileName )
-					&& defaultConfigFile[profileName].ContainsKey( "credential_process" ) ) {
+				if( defaultConfigFile.Sections.ContainsSection( sectionName )
+					&& defaultConfigFile[sectionName].ContainsKey( "credential_process" ) ) {
 
-					if( defaultConfigFile[profileName].Count == 1 ) {
-						defaultConfigFile.Sections.RemoveSection( profileName );
+					if( defaultConfigFile[sectionName].Count == 1 ) {
+						defaultConfigFile.Sections.RemoveSection( sectionName );
 					} else {
-						defaultConfigFile[profileName].RemoveKey( "credential_process" );
+						defaultConfigFile[sectionName].RemoveKey( "credential_process" );
 					}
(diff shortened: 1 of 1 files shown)


src/D2L.Bmx/WriteHandler.cs
```diff
@@ -85,6 +85,27 @@ bool useCredentialProcess
...
+						defaultConfigFile.Sections.RemoveSection( profileName );
+					} else {
+						defaultConfigFile[profileName].RemoveKey( "credential_process" );
+					}
+					parser.WriteFile( SharedCredentialsFile.DefaultConfigFilePath, defaultConfigFile );
+					Console.WriteLine(
+						"""
+						An existing profile with the same name using the `credential_process` key was found in the default config file.
+						The key has been removed, and static credentials will be used for the profile.
+						To continue using non-static credentials, rerun the command with the --use-credential-process flag.
```
cfbao, line 104: "key" is a pretty technical term. I find it a little weird/confusing for user-facing language.
Maybe use "setting", as that's what the AWS doc uses:
https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html

```suggestion
						An existing profile with the same name using the `credential_process` setting was found in the default config file.
						The setting has been removed, and static credentials will be used for the profile.
						To continue using non-static credentials, rerun the command with the --use-credential-process flag.
```

src/D2L.Bmx/WriteHandler.cs
```diff
@@ -85,6 +85,27 @@ bool useCredentialProcess
 				+ $" --role {awsCredsInfo.Role}"
 				+ $" --duration {awsCredsInfo.Duration}";
 		} else {
+			if( File.Exists( SharedCredentialsFile.DefaultConfigFilePath ) ) {
+				string profileName = $"profile {profile}";
```
cfbao, line 89: nit: it's really the section name rather than profile name
```suggestion
				string sectionName = $"profile {profile}";
```

cfbao: I feel like there's an opportunity to refactor and dedupe some logic and declarations in this whole "write" area. It's getting a bit messy now. But we can do it another time.

src/D2L.Bmx/WriteHandler.cs
gord5500: Yea I went back and forth on what to call that but setting does sound better

cfbao (approved):

src/D2L.Bmx/WriteHandler.cs
gord5500: Specifically the section removing part or something else?

cfbao: everything about handling AWS creds and config files

PR #454, opened 2024-06-21
refactor ini file handling logic for write command

cfbao:
follow up to this comment: https://github.com/Brightspace/bmx/pull/453#discussion_r1642971687

### DIFF
--- a/src/D2L.Bmx/WriteHandler.cs
+++ b/src/D2L.Bmx/WriteHandler.cs
@@ -1,6 +1,8 @@
+using System.Diagnostics;
 using System.Text;
 using Amazon.Runtime.CredentialManagement;
 using IniParser;
+using IniParser.Model;
 
 namespace D2L.Bmx;
 
@@ -11,6 +13,9 @@ internal class WriteHandler(
 	BmxConfig config,
 	FileIniDataParser parser
 ) {
+	// this is different from `Encoding.UTF8` which has the undesirable `encoderShouldEmitUTF8Identifier: true`
+	private static readonly UTF8Encoding Utf8 = new();
+
 	public async Task HandleAsync(
 		string? org,
 		string? user,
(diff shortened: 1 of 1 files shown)


src/D2L.Bmx/WriteHandler.cs
```diff
@@ -64,39 +69,50 @@ bool useCredentialProcess
 			using( File.Create( output ) ) { };
 		}
 
-		var data = parser.ReadFile( output );
+		string awsConfigFilePath = useCredentialProcess ? output : SharedCredentialsFile.DefaultConfigFilePath;
+		var awsConfig = GetParsedIniFileOrNull( awsConfigFilePath );
+
+		string awsCredentialsFilePath = useCredentialProcess ? SharedCredentialsFile.DefaultFilePath : output;
+		var awsCredentials = GetParsedIniFileOrNull( awsCredentialsFilePath );
```
cfbao, line 76: This is the main change.

My main pet peeve in this area is that `output` and `data` could have one of two different meanings, depending on the value of `useCredentialProcess`. I find this confusing lower down where the logic splits into two branches.

So, here, I changed the variable assignment so that `output` will no longer be used beyond this point, and we will have distinct names for the two different types of files/data.

As a nice side-effect, this is also the last place where we do any INI file parsing.

src/D2L.Bmx/WriteHandler.cs
```diff
@@ -64,39 +69,50 @@ bool useCredentialProcess
...
+					&& awsConfig[sectionName].ContainsKey( "credential_process" )
+				) {
+					if( awsConfig[sectionName].Count == 1 ) {
+						awsConfig.Sections.RemoveSection( sectionName );
 					} else {
-						defaultConfigFile[sectionName].RemoveKey( "credential_process" );
+						awsConfig[sectionName].RemoveKey( "credential_process" );
 					}
-					parser.WriteFile( SharedCredentialsFile.DefaultConfigFilePath, defaultConfigFile );
+					parser.WriteFile( awsConfigFilePath, awsConfig, Utf8 );
```
cfbao, line 115: nit: without explicitly specifying the encoding, `parser` will write file using UTF8-BOM 🤢

src/D2L.Bmx/WriteHandler.cs
```diff
@@ -64,39 +69,50 @@ bool useCredentialProcess
...
+		var awsConfig = GetParsedIniFileOrNull( awsConfigFilePath );
+
+		string awsCredentialsFilePath = useCredentialProcess ? SharedCredentialsFile.DefaultFilePath : output;
+		var awsCredentials = GetParsedIniFileOrNull( awsCredentialsFilePath );
+
 		if( useCredentialProcess ) {
-			string sectionName = $"profile {profile}";
-			if( !data.Sections.ContainsSection( sectionName ) ) {
-				data.Sections.AddSection( sectionName );
+			Debug.Assert( awsConfig is not null );
```
boarnoah, line 79: What is the thought process behind this being an assertion only in debug builds?

cfbao: The main goal here is to make the compiler's nullable analysis happy in a clear and intentional way.

`awsConfig` should not be null here when `useCredentialProcess` is true (because we'd create the AWS config file above). The compiler cannot infer this, so by default it'd mark `awsConfig` as "may be null", and issue warnings at [L88](https://github.com/Brightspace/bmx/pull/454/files#diff-a6eaa1b3c5a510cd785801e0c37dfeb64b19bdb9263596f8ac5e9ed45bc9eba5R88) and below.

We could use the null-forgiving operator `!`, but it doesn't make our assumptions and intent clear:
* Why can we assume `awsConfig` isn't null in these places?
* When does this assumption hold?

By using `Debug.Assert`, we're effectively making clarifying statements like
* `foo` isn't null inside this if/else/for/while block
* `bar` isn't null once this above statement is executed successfully

We can also avoid sprinkle `!` in many different places.

---

The fact that this is also a runtime assertion in debug builds is more or less a side effect that I don't care that much. I don't think we should make it actually asserts at runtime when `BMX_DEBUG` is set either.

cfbao: We've used this technique to document our assumptions in code before, e.g.

* https://github.com/Brightspace/instance-metadata-store/blob/75ad7c513c12c6898d8e8397292ba67bc56cc9d0/api/src/D2L.InstanceMetadataStore.Api/CacheBuster.cs#L74

* https://github.com/Brightspace/instance-metadata-store/blob/75ad7c513c12c6898d8e8397292ba67bc56cc9d0/worker/src/D2L.InstanceMetadataStore.Worker/ConfigEtl/Extractor/ConfigExtractorProcessor.cs#L80

It's also used by the .NET runtime libraries, e.g.

* https://github.com/dotnet/runtime/blob/e1efa6b633d01c00ec019a8e222b2ebdfa593bcb/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/RedirectHandler.cs#L34

---

These are assumptions that should always hold true. If there's a real non-negligible chance that they could be false, then they should become real runtime assertions like
```csharp
if ( !assumption ) {
	throw new InvalidOperationException( "bad condition" );
}
```

boarnoah (approved):

PR #458, opened 2024-07-03
Improve question MFA prompt

### LINKED ISSUE #446: Authentication improvements for OKTA-question choice
Used recently the authentication process and noticed that:
1. when OKTA-question is selected as an authentication method the question is not displayed
2. when OKTA-question is selected as an authentication method the answer is not obfuscated

boarnoah:
### Why

This MFA factor (which incidentally I thought we didn't allow with D2L's Okta) looked like:

```
Okta org short name or domain name: dev-1234.oktapreview.com
Okta username: awijayathilaka
Okta password: *******************
MFA Required
[1] OKTA-question
[2] GOOGLE-token:software:totp
Select an available MFA option: 1
Answer: <answer in plain text>
Successfully logged in and Okta session has been cached.
```

Now, it will mask the input for the security question (which is a static value) + prompt with the question text rather than "Answer:"
```
Okta org short name or domain name: dev-1234.oktapreview.com
Okta username: awijayathilaka
Okta password: *******************
MFA Required
[1] OKTA-question
[2] GOOGLE-token:software:totp
Select an available MFA option: 1
What is your favorite security question?: ****************
Successfully logged in and Okta session has been cached.
```

### Ticket

https://desire2learn.atlassian.net/browse/VUL-400
 
resolves #446

### DIFF
--- a/src/D2L.Bmx/ConsolePrompter.cs
+++ b/src/D2L.Bmx/ConsolePrompter.cs
@@ -130,12 +130,12 @@ internal class ConsolePrompter : IConsolePrompter {
 		}
 
 		if( mfaOptions.Length == 1 ) {
-			Console.Error.WriteLine( $"MFA method: {mfaOptions[0].Provider}-{mfaOptions[0].FactorType}" );
+			Console.Error.WriteLine( $"MFA method: {mfaOptions[0].Provider}-{mfaOptions[0].FactorName}" );
 			return mfaOptions[0];
 		}
 
 		for( int i = 0; i < mfaOptions.Length; i++ ) {
-			Console.Error.WriteLine( $"[{i + 1}] {mfaOptions[i].Provider}-{mfaOptions[i].FactorType}" );
+			Console.Error.WriteLine( $"[{i + 1}] {mfaOptions[i].Provider}-{mfaOptions[i].FactorName}" );
 		}
 		Console.Error.Write( "Select an available MFA option: " );
 		if( !int.TryParse( _stdinReader.ReadLine(), out int index ) || index > mfaOptions.Length || index < 1 ) {
--- a/src/D2L.Bmx/ConsolePrompter.cs
+++ b/src/D2L.Bmx/ConsolePrompter.cs
@@ -13,7 +13,7 @@ internal interface IConsolePrompter {
 	string PromptAccount( string[] accounts );
 	string PromptRole( string[] roles );
 	OktaMfaFactor SelectMfa( OktaMfaFactor[] mfaOptions );
-	string GetMfaResponse( string mfaInputPrompt );
+	string GetMfaResponse( string mfaInputPrompt, bool maskInput );
 }
 
 internal class ConsolePrompter : IConsolePrompter {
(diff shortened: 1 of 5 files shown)


src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -208,16 +144,90 @@ internal class ConsolePrompter : IConsolePrompter {
...
+			mfaInput = _stdinReader.ReadLine();
+		}
 
 		if( mfaInput is not null ) {
 			return mfaInput;
 		}
 		throw new BmxException( "Invalid MFA Input" );
 	}
 
+	private string GetMaskedInput( string prompt ) {
```
boarnoah, line 163: Refactored this masked input read to be re-usable (for GetMfaResponse)

src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
```diff
@@ -21,11 +21,16 @@ internal record AuthenticateResponseEmbedded(
...
+	string QuestionText
+);
+
 internal record OktaMfaFactor(
 	string Id,
 	string FactorType,
 	string Provider,
-	string VendorName
+	string VendorName,
+	OktaMfaQuestionProfile Profile
```
boarnoah, line 33: This I really do not like. 

It isn't straightforward to handle extracting nested value & do type discrimination based on FactorType for de serialization.

Without using a custom convertor at least AFAICS, so went with this approach.

gord5500 (dismissed):

src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
cfbao: Isn't this what type discriminator is designed for? I don't see the need for a custom convertor.

boarnoah: Couple of reasons,

1. Out of order discriminator isn't in as of .NET 8 ? https://github.com/dotnet/runtime/pull/97474
1. Need a bit more of a code change to differentiate between types without FactorType as a property (to not conflict with it's use as the type discriminator) 
1. It still looks kind of silly,  how even with type discrimination we would still need need to do some additional work to hoist QuestionText up from nested profile, ex:
      ```csharp
      // Binding to a nested property in polymorphic Profile
      [EditorBrowsable( EditorBrowsableState.Never )]
      public required OktaMfaQuestionProfile Profile { get; set; }
      public string QuestionText => Profile.QuestionText;
      ```

 All of which to me suggests, a custom convertor is the way to go, if we want to take it all the way (even if we discard the third point)

src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -208,16 +144,90 @@ internal class ConsolePrompter : IConsolePrompter {
...
+		Func<char> readKey;
+		if( IS_WINDOWS ) {
+			// On Windows, Console.ReadKey calls native console API, and will fail without a console attached
+			if( Console.IsInputRedirected ) {
+				Console.Error.WriteLine( """
+					====== WARNING ======
```
boarnoah, line 169: It is going to pop this twice once for password + security question, in this case. But meh, that is a very niche edge case

src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
cfbao: 1. We can use the STJ preview package.
2. Yes, which should be straightforward.
3. Not following... why?

boarnoah: Moved to a separate PR targeting this ( #459 )since there is a few ways we can go about this.

re: 3,
Agreed it isn't necessary. Shouldn't really manipulate the Okta API types unnecessarily, after all its all still encapsulated within the context of the Okta authentication related code (no additional leak of implementation details involved).

gord5500 (dismissed):

boarnoah:
Changed separator now that we can't use Factor Type directly & need to use a custom pretty name for Factor
```
Okta org short name or domain name: dev-1234.oktapreview.com
Okta username: awijayathilaka
Okta password: *******************
MFA Required
[1] OKTA : Security Question
[2] OKTA : Call
[3] OKTA : SMS
[4] GOOGLE : Software TOTP
[5] OKTA : Software TOTP
Select an available MFA option: 1
What is your favorite security question?: ****************
Successfully logged in and Okta session has been cached.
```

Probably shouldn't align factor list vertically with padding since vendor name length has unknown bounds + it remains closer to the original look.

cfbao (approved):

PR #459, opened 2024-07-08
De-serialize Security Question MFA to different type

boarnoah:
### Why

With .NET 9 we can use properties aside from the first in JSON to do type discrimination (in this case using ``factorType`` in the API).

So we can use it de-serialize the different properties in Factor's Profile section (in this case we only care about QuestionText for the Security Question factor).

### Ticket

N/A

### DIFF
--- a/src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
+++ b/src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
@@ -38,7 +38,8 @@ internal record OktaMfaFactor {
 	public required string Provider { get; set; }
 	public required string VendorName { get; set; }
 
-	public const string FactorType = "unknown";
+	public const string UnsupportedMfaFactor = "unknown";
+
 	[JsonIgnore]
 	public virtual string FactorName => "unknown";
 	[JsonIgnore]
--- a/src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
+++ b/src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
@@ -21,17 +21,68 @@ internal record AuthenticateResponseEmbedded(
 	OktaMfaFactor[]? Factors
 );
 
+[JsonPolymorphic(
+	TypeDiscriminatorPropertyName = "factorType",
+	IgnoreUnrecognizedTypeDiscriminators = true
+)]
+[JsonDerivedType( typeof( OktaMfaQuestionFactor ), "question" )]
+[JsonDerivedType( typeof( OktaMfaHardwareToken ), "token" )]
+[JsonDerivedType( typeof( OktaMfaHardwareTokenFactor ), "token:hardware" )]
+[JsonDerivedType( typeof( OktaMfaSoftwareTotpFactor ), "token:software:totp" )]
+[JsonDerivedType( typeof( OktaMfaHotpFactor ), "token:hotp" )]
+[JsonDerivedType( typeof( OktaMfaSmsFactor ), "sms" )]
+[JsonDerivedType( typeof( OktaMfaCallFactor ), "call" )]
+[JsonDerivedType( typeof( OktaMfaEmailFactor ), "email" )]
+internal record OktaMfaFactor {
+	public required string Id { get; set; }
+	public required string Provider { get; set; }
+	public required string VendorName { get; set; }
+
+	[JsonIgnore]
+	public virtual string FactorName => "Unknown";
+	[JsonIgnore]
+	public virtual bool RequireChallengeIssue => false;
+}
+
+internal record OktaMfaQuestionFactor() : OktaMfaFactor() {
+	public override string FactorName => "Security Question";
+	public required OktaMfaQuestionProfile Profile { get; set; }
+}
+
 internal record OktaMfaQuestionProfile(
 	string QuestionText
 );
 
-internal record OktaMfaFactor(
-	string Id,
-	string FactorType,
-	string Provider,
-	string VendorName,
-	OktaMfaQuestionProfile Profile
-);
+internal record OktaMfaHardwareToken() : OktaMfaFactor {
+	public override string FactorName => "Hardware Token";
+}
+
+internal record OktaMfaHardwareTokenFactor() : OktaMfaFactor {
+	public override string FactorName => "Hardware TOTP";
+}
+
+internal record OktaMfaSoftwareTotpFactor() : OktaMfaFactor {
+	public override string FactorName => "Software TOTP";
+}
+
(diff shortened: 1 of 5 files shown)


src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
```diff
@@ -21,17 +21,68 @@ internal record AuthenticateResponseEmbedded(
 	OktaMfaFactor[]? Factors
 );
 
+[JsonPolymorphic(
+	TypeDiscriminatorPropertyName = "factorType",
```
boarnoah, line 25: Unfortunately, STJ doesn't offer a way to include the property in JSON used for type discrimination in the de-serialized object.

I did find a issue asking for this capability (can't find it now). It makes sense since ideally if you are using type discrimination, then the rest of your code should use the C# type system to handle differences in the various types.

In this case, where we only want to treat one of several types differently (but include different names for each type), this is incredibly cumbersome.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -57,18 +57,18 @@ bool ignoreCache
 		if( authnResponse is AuthenticateResponse.MfaRequired mfaInfo ) {
 			OktaMfaFactor mfaFactor = consolePrompter.SelectMfa( mfaInfo.Factors );
 
-			if( !IsMfaFactorTypeSupported( mfaFactor.FactorType ) ) {
+			if( mfaFactor.FactorName == "Unknown" ) {
```
boarnoah, line 60: Not a fan of this at all

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -57,18 +57,18 @@ bool ignoreCache
...
 				throw new BmxException( "Selected MFA not supported by BMX" );
 			}
 
 			// TODO: Handle retry
-			if( mfaFactor.FactorType is "sms" or "call" or "email" ) {
+			if( mfaFactor.RequireChallengeIssue ) {
```
boarnoah, line 65: Alternatively ``if ( mfaFactor is OktaMfaEmailFactor ... ) {``  etc....?

src/D2L.Bmx/OktaAuthenticator.cs
cfbao: This can be a type check against the (maybe renamed) base type, or at least you can make the string a constant so there's no possibility of a typo.

src/D2L.Bmx/OktaAuthenticator.cs
cfbao: as is seems good

src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
```diff
@@ -21,17 +21,68 @@ internal record AuthenticateResponseEmbedded(
...
+
+	[JsonIgnore]
+	public virtual string FactorName => "Unknown";
+	[JsonIgnore]
+	public virtual bool RequireChallengeIssue => false;
+}
+
+internal record OktaMfaQuestionFactor() : OktaMfaFactor() {
+	public override string FactorName => "Security Question";
+	public required OktaMfaQuestionProfile Profile { get; set; }
```
cfbao, line 49: does this need to have a setter?

src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
```diff
@@ -21,17 +21,68 @@ internal record AuthenticateResponseEmbedded(
...
+	public required string VendorName { get; set; }
+
+	[JsonIgnore]
+	public virtual string FactorName => "Unknown";
+	[JsonIgnore]
+	public virtual bool RequireChallengeIssue => false;
+}
+
+internal record OktaMfaQuestionFactor() : OktaMfaFactor() {
+	public override string FactorName => "Security Question";
```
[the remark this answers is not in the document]
boarnoah: I did play around with that, didn't find a great solution for keeping the discriminator as a static value but also to be able to use derived type's value for it without downcasting.

src/D2L.Bmx/Okta/Models/AuthenticateResponse.cs
boarnoah: Don't think so. This could probably a constructor parameter.

src/D2L.Bmx/OktaAuthenticator.cs
boarnoah: Doesn't look great IMO. https://github.com/Brightspace/bmx/pull/459/commits/19413d471f72c72cba5822489427766baa96cec9

I think we should instead rely on something using ``UnknownDerivedTypeHandling = JsonUnknownDerivedTypeHandling.FailSerialization``
& filtering out failures seems better although definitely more involved code wise (de-serialization call for this API response handling will need rework).

After all we aren't proceeding when unknown factor type is selected, might as well not populate the list with them (maybe a separate message for "x, y factors are not supported by BMX"

cfbao: I was thinking of giving "unknown" a special constant, rather than reusing the "FactoryType" thingy we're using elsewhere. Reusing that would be confusing for sure.

What about doing a runtime type check though? We could call the base type something like `OktaMfaFactorUnspecified`, and check the exact type (rather than using `is` which considers child types a match).

More involved code change is precisely why I didn't suggest failing serialization directly, but that's an option for sure. We can even make the base type abstract.

cfbao: actually, with a better base type name, the constant check wouldn't look so confusing either.

boarnoah: Think we should go with a standalone const + factor name check, 

https://github.com/Brightspace/bmx/pull/459/commits/b57bf153c866c1ed23d077d374a5cd4448073e1e is confusing in it's own right (since we use the base type name in quite a few contexts, it is awkward to name it "unknown"/"unsupported".

boarnoah: Maybe ```BaseOktaMfaFactor```  or something might help, but it still feels wrong.

cfbao: Yeah, I see. Renaming the base type doesn't work. A distinctly named constant is good. The `FactorType` constant isn't in the type hierarchy anyway.

boarnoah: Changed to a better named const for this, should revisit if this ends up happening https://github.com/dotnet/runtime/issues/93471

PR #462, opened 2024-07-19
update to .NET 9

cfbao:
### Why

The trimmer seems to have trouble preserving the correct type when we reference a package that provides types included in the runtime.

Specifically, we're using [a System.Text.Json preview package](https://www.nuget.org/packages/System.Text.Json/9.0.0-preview.5.24306.7) to utilize the new `AllowOutOfOrderMetadataProperties` feature, but the trimmer isn't keeping relevant new members in Json Serialization types (even if explicitly instructed to by a trimmer root descriptor file), resulting in runtime errors.

Updating to .NET 9 solves this problem, as we no longer need the preview package, and the types/members we need are in the runtime.

This is likely a trimmer bug that we should report to https://github.com/dotnet/runtime/issues.

### Ticket

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -17,7 +17,7 @@ jobs:
       - name: set up dotnet
         uses: Brightspace/third-party-actions@actions/setup-dotnet
         with:
-          dotnet-version: 8.0.x
+          dotnet-version: 9.0.x
 
       - name: format
         id: format
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -76,7 +76,7 @@ jobs:
       - name: set up .NET
         uses: Brightspace/third-party-actions@actions/setup-dotnet
         with:
-          dotnet-version: 8.0.x
+          dotnet-version: 9.0.x
 
       - name: publish
         shell: pwsh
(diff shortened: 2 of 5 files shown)


jharoutuniand2l (approved):

boarnoah (approved):
> but the trimmer isn't keeping relevant new members in Json Serialization types 

Not clear to me, how did you notice something like this.

I think my local testing while implementing that PR was incomplete since it didn't include testing with an actual trimmed build?

cfbao:
> > but the trimmer isn't keeping relevant new members in Json Serialization types
> 
> Not clear to me, how did you notice something like this.

There are build time warnings & runtime errors that tell me so.

> I think my local testing while implementing that PR was incomplete since it didn't include testing with an actual trimmed build?

Correct. Should test with a published build.

cfbao:
@boarnoah 

> > but the trimmer isn't keeping relevant new members in Json Serialization types
> 
> Not clear to me, how did you notice something like this.

There are build time warnings & runtime errors that tell me so.

> I think my local testing while implementing that PR was incomplete since it didn't include testing with an actual trimmed build?

Correct. Should test with a published build.

cfbao:
Issue filed in .NET runtime repo: https://github.com/dotnet/runtime/issues/105274

PR #463, opened 2024-07-22
don't print duplicate org & user info

cfbao:
### Why

This is niche, but glaring once you notice it:
If Okta org and/or username aren't specified in BMX config file or provided via command line args, BMX will prompt the user to enter the info, and immediately print the same info again:

Screenshot 2024-07-20 092346

A good solution isn't exactly obvious or straightforward as one might think.

A few options:

1. Keep the behaviour as is:
  Simple, but looks bad in niche cases

1. Print org or user at the time of password prompt, but only if there was no org or user prompt:
   Simple, but the order can be inconsistent in certain cases.
   E.g. if only org is prompted, we'd get this:
   ```
   Okta org name or domain name: d2l
   Okta username: cbao
   Okta password: ***
   ```
   if only user is prompted, we'd get this:
   ```
   Okta username: cbao
   Okta org name or domain name: d2l
   Okta password: ***
   ```
   Also, it'd be nice if people can see the Okta org when they're being prompted for their Okta username

1. Always print both Okta org and user (and potentially also AWS account name & role name) when not entered via a prompt
  Simple and good interactive UX, but BMX will become a lot more chatty than before, which some users might find jarring? Possibly considered a breaking change for some scripts?

1. Print both Okta org and user exactly once and always in the same order if and only if one or more of org, user or password is prompted.
  Most consistent with BMX 3.1.0- behaviour. Very self-consistent too. However, the implementation logic will be complex and unintuitive.
  This PR implements this option to showcase its complexity.

### Ticket

### DIFF
--- a/src/D2L.Bmx/ConsolePrompter.cs
+++ b/src/D2L.Bmx/ConsolePrompter.cs
@@ -8,7 +8,7 @@ internal interface IConsolePrompter {
 	string PromptOrg( bool allowEmptyInput );
 	string PromptProfile();
 	string PromptUser( bool allowEmptyInput );
-	string PromptPassword( string user, string org );
+	string PromptPassword();
 	int? PromptDuration();
 	string PromptAccount( string[] accounts );
 	string PromptRole( string[] roles );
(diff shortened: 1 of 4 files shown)


cfbao:
As per discussion, will proceed with option 3.
Will open a different PR for it.

PR #464, opened 2024-07-22
print parameter value if and only if provided via CLI arg or config

cfbao:
### Why

Follow up to https://github.com/Brightspace/bmx/pull/463
As per Zoom discussion, BMX will always print parameter value & source if and only if it's provided via CLI args or config file.

I also made it colourful for legibility. The line seems too long and crowded without these colours.
image

The colours don't work if stdout is redirected, which is typical when using `bmx print`. I know why this is happening. Will try to fix in a separate PR.

### Ticket

### DIFF
--- a/src/D2L.Bmx/AwsCredsCreator.cs
+++ b/src/D2L.Bmx/AwsCredsCreator.cs
@@ -13,6 +13,7 @@ internal record AwsCredentialsInfo(
 internal class AwsCredsCreator(
 	IAwsClient awsClient,
 	IConsolePrompter consolePrompter,
+	IConsoleWriter consoleWriter,
 	IAwsCredentialCache awsCredentialCache,
 	BmxConfig config
 ) {
@@ -30,12 +31,22 @@ internal class AwsCredsCreator(
 			);
 		}
 
+		var accountSource = ParameterSource.CliArg;
 		if( string.IsNullOrEmpty( account ) && !string.IsNullOrEmpty( config.Account ) ) {
 			account = config.Account;
+			accountSource = ParameterSource.Config;
+		}
+		if( !string.IsNullOrEmpty( account ) && !nonInteractive ) {
+			consoleWriter.WriteParameter( ParameterDescriptions.Account, account, accountSource );
 		}
 
+		var roleSource = ParameterSource.CliArg;
 		if( string.IsNullOrEmpty( role ) && !string.IsNullOrEmpty( config.Role ) ) {
 			role = config.Role;
+			roleSource = ParameterSource.Config;
+		}
+		if( !string.IsNullOrEmpty( role ) && !nonInteractive ) {
+			consoleWriter.WriteParameter( ParameterDescriptions.Role, role, roleSource );
 		}
 
 		if( duration is null or 0 ) {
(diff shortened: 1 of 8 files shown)


src/D2L.Bmx/ParameterSource.cs
```diff
@@ -0,0 +1,14 @@
+namespace D2L.Bmx;
+
+internal record ParameterSource {
```
cfbao, line 3: use a record type rather than an enum so it's:
* easy to customize `ToString()`
* easy to extend in the future to supply config file path

src/D2L.Bmx/ConsoleWriter.cs
```diff
@@ -0,0 +1,17 @@
+namespace D2L.Bmx;
+
+internal interface IConsoleWriter {
```
cfbao, line 3: We're reasonably close to a point where some unit tests around our prompting and printing logic would be worthwhile.
I could really use a few when I was writing up https://github.com/Brightspace/bmx/pull/463
Interfaces are handy then.

src/D2L.Bmx/AwsCredsCreator.cs
```diff
@@ -30,12 +31,22 @@ bool cache
 			);
 		}
 
+		var accountSource = ParameterSource.CliArg;
 		if( string.IsNullOrEmpty( account ) && !string.IsNullOrEmpty( config.Account ) ) {
 			account = config.Account;
+			accountSource = ParameterSource.Config;
+		}
+		if( !string.IsNullOrEmpty( account ) && !nonInteractive ) {
+			consoleWriter.WriteParameter( ParameterDescriptions.Account, account, accountSource );
```
cfbao, line 40: seems reasonable to not print these out when user has explicitly specified `--non-interactive`?

src/D2L.Bmx/ConsoleWriter.cs
```diff
@@ -0,0 +1,17 @@
...
+}
+
+internal class ConsoleWriter : IConsoleWriter {
+	void IConsoleWriter.WriteParameter( string description, string value, ParameterSource source ) {
+		Console.ResetColor();
+		Console.Error.Write( $"{description}: " );
+		Console.ForegroundColor = ConsoleColor.Cyan;
+		Console.Error.Write( value );
+		Console.ForegroundColor = ConsoleColor.DarkGray;
+		Console.Error.WriteLine( $" (from {source})" );
```
cfbao, line 14: alternatively:
```suggestion
		Console.Error.WriteLine( $" (source: {source})" );
```
🤷

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -21,24 +22,32 @@ BmxConfig config
 		bool nonInteractive,
 		bool ignoreCache
 	) {
+		var orgSource = ParameterSource.CliArg;
+		if( string.IsNullOrEmpty( org ) && !string.IsNullOrEmpty( config.Org ) ) {
```
cfbao, line 26: aside: writing all this overwrite/print/prompt logic really has a fizz buzz feel - it feels like you should be able to consolidate some actions and/or logical branches, but actually you can't...

src/D2L.Bmx/ConsoleWriter.cs
gord5500: My vote is `from`

gord5500 (approved):

PR #466, opened 2024-07-23
use ANSI escape codes to control colours of parameters we print

cfbao:
### Why

.NET's `Console.ForegroundColor` only targets stdout.
If we rely on `Console.ForegroundColor` and stdout is redirected (e.g. `bmx print | iex`), we won't get any colour on stderr either.
Related .NET issue: https://github.com/dotnet/runtime/issues/83146

Changing to use ANSI escape codes to control text colours to work around this issue.

### Ticket

---

depends on https://github.com/Brightspace/bmx/pull/464

### DIFF
--- a/src/D2L.Bmx/VirtualTerminal.cs
+++ b/src/D2L.Bmx/VirtualTerminal.cs
@@ -1,0 +1,54 @@
+using System.Runtime.InteropServices;
+
+namespace D2L.Bmx;
+
+// The code here exists because legacy Windows console doesn't support ANSI escape codes by default.
+// We can consider deleting everything here once Windows 10 is out of support and
+// Windows Terminal becomes the default console on all supported Windows versions.
+internal static partial class VirtualTerminal {
+	private const string Kernel32 = "kernel32.dll";
+	private const int STD_ERROR_HANDLE = -12;
+	private const uint ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004;
+
+	private static bool _enablingAttempted = false;
+	private static bool _enabled = false;
+
+	// https://learn.microsoft.com/en-us/windows/console/classic-vs-vt
+	public static bool TryEnableOnStderr() {
+		if( _enablingAttempted ) {
+			return _enabled;
+		}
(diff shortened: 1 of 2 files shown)


scowing (approved):

gord5500 (dismissed):

src/D2L.Bmx/VirtualTerminal.cs
```diff
@@ -0,0 +1,54 @@
...
+		}
+		nint stderrHandle = GetStdHandle( STD_ERROR_HANDLE );
+		if( !GetConsoleMode( stderrHandle, out int mode ) ) {
+			_enabled = false;
+			return false;
+		}
+		if( ( mode & ENABLE_VIRTUAL_TERMINAL_PROCESSING ) == ENABLE_VIRTUAL_TERMINAL_PROCESSING ) {
+			_enabled = true;
+			return true;
+		}
```
boarnoah, line 36: This seems to be for "its already enabled"?

src/D2L.Bmx/VirtualTerminal.cs
```diff
@@ -0,0 +1,54 @@
...
+		_enabled = SetConsoleMode( stderrHandle, (int)( mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING ) );
+		return _enabled;
+	}
+
+	// https://learn.microsoft.com/en-us/windows/console/getstdhandle
+	[LibraryImport( Kernel32 )]
+	private static partial IntPtr GetStdHandle( int nStdHandle );
```
boarnoah, line 43: You sure we don't need to do [``SetLastError = true``](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportattribute.setlasterror)  here?

If we are calling into un-managed code & checking if that was a success (I might be mis-understanding the purpose of that flag).

``On .NET, the error information is cleared (set to 0) before invoking the callee when this field is set to true``
So maybe that only applies for repeated calls with subsequent successes?

boarnoah (dismissed):

gord5500 (approved):

PR #468, opened 2024-07-24
fix up & touch up user facing messages

cfbao:
Noticed some minor issues with user-facing messages.

### DIFF
--- a/src/D2L.Bmx/Aws/AwsCredsCache.cs
+++ b/src/D2L.Bmx/Aws/AwsCredsCache.cs
@@ -61,7 +61,7 @@ internal class AwsCredsCache : IAwsCredentialCache {
 				&& o.Org == org
 			)
 			// Prune older (closer to expiry) credentials for the same role
-			.GroupBy( o => $"{o.AccountName}.${o.RoleName}", ( _, entries ) =>
+			.GroupBy( o => (o.AccountName, o.RoleName), ( _, entries ) =>
 				entries.OrderBy( o => o.Credentials.Expiration )
 			)
 			.Select( o => o.Last() );
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -8,7 +8,11 @@ internal interface IBmxConfigProvider {
 	void SaveConfiguration( BmxConfig config );
 }
 
-internal class BmxConfigProvider( FileIniDataParser parser ) : IBmxConfigProvider {
+internal class BmxConfigProvider(
+	FileIniDataParser parser,
+	IConsoleWriter consoleWriter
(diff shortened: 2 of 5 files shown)


src/D2L.Bmx/Aws/AwsCredsCache.cs
```diff
@@ -61,7 +61,7 @@ AwsCredentials credentials
 				&& o.Org == org
 			)
 			// Prune older (closer to expiry) credentials for the same role
-			.GroupBy( o => $"{o.AccountName}.${o.RoleName}", ( _, entries ) =>
+			.GroupBy( o => (o.AccountName, o.RoleName), ( _, entries ) =>
```
cfbao, line 64: Somewhat unrelated - this isn't a user-facing message, but it's also a string-related issue.

* [Value tuples](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples) are preferred when we want to use multiple values as a single key. Converting them to a single string is unnecessary and has more overhead.
* Even if we do want to use a string here, it should've been
  ```diff
  -$"{o.AccountName}.${o.RoleName}"
  +$"{o.AccountName}.{o.RoleName}"
  ```
  A common mistake when using C# interpolated strings.

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -18,7 +22,7 @@ internal class BmxConfigProvider( FileIniDataParser parser ) : IBmxConfigProvide
 				var tempdata = parser.ReadFile( configFileName );
 				data.Merge( tempdata );
 			} catch( Exception ) {
-				Console.Error.Write( $"WARNING: Failed to load the global config file {configFileName}." );
+				consoleWriter.WriteWarning( $"WARNING: Failed to load the global config file {configFileName}." );
```
cfbao, line 25: It's hard to reproduce such a failure scenario so it's never manifested, but `Console.Error.Write` (without a newline) would've never worked nicely.

src/D2L.Bmx/Okta/OktaApi.cs
```diff
@@ -88,9 +88,10 @@ internal class OktaApi : IOktaApi {
 		}
 
 		string org = _organization ?? "unknown";
-		throw new BmxException(
-			$"Okta authentication for user '{username}' in org '{org}'"
-				+ "failed. Check if org, user, and password is correct" );
+		throw new BmxException( $"""
+			Okta authentication for user '{username}' in org '{org}' failed.
+			Check if org, user, and password is correct.
+			""" );
```
cfbao, line 94: This one was missing a space before "failed", and the usage of period was inconsistent:
image

Easier (and clearer, I think) to use a raw string literal with multiple lines instead.

gord5500 (approved):

PR #469, opened 2024-07-24
fix NO_COLOR branching & support coloured text in mintty

cfbao:
testing in mintty revealed a few problems...


src/D2L.Bmx/ConsoleWriter.cs
```diff
@@ -31,6 +32,8 @@ internal class ConsoleWriter : IConsoleWriter {
 	void IConsoleWriter.WriteUpdateMessage( string text ) {
 		if( _noColor ) {
 			Console.Error.WriteLine( text );
+			Console.Error.WriteLine();
```
cfbao, line 35: extra newline to match with existing behaviour and the coloured text behaviour.

boarnoah (approved):

PR #472, opened 2024-07-26
fix how we find parent process

cfbao:
### Why

Turns out BusyBox only implements a small portion of the [POSIX spec][1] for its [ps][2], so our code that gets the parent process (and hence `bmx print`) never worked on Alpine...
(I guess no one ever used `bmx print` on alpine...)

Switch to use native API call, which is actually easier.

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ps.html
[2]: https://boxmatrix.info/wiki/Property:ps

### Ticket


.github/workflows/publish.yml
```diff
@@ -45,9 +45,9 @@ jobs:
             Throw "Tag name format incorrect"
           }
           if ($env:PRERELEASE -eq 'true') {
-              $releaseUrl = gh release create "$env:RELEASE_TAG" --repo Brightspace/bmx --draft --prerelease --generate-notes --title "Release $env:RELEASE_TAG"
+              $releaseUrl = gh release create "$env:RELEASE_TAG" --repo Brightspace/bmx --draft --prerelease --generate-notes --title "Release $env:RELEASE_TAG" --target $env:GITHUB_SHA
           } else {
-              $releaseUrl = gh release create "$env:RELEASE_TAG" --repo Brightspace/bmx --draft --generate-notes --title "Release $env:RELEASE_TAG"
+              $releaseUrl = gh release create "$env:RELEASE_TAG" --repo Brightspace/bmx --draft --generate-notes --title "Release $env:RELEASE_TAG" --target $env:GITHUB_SHA
```
cfbao, line 50: unrelated fix - our release job only worked on the main branch.

src/D2L.Bmx/ParentProcess.cs
```diff
@@ -2,43 +2,41 @@
...
 /**
 This file may eventually be replaced if the following dotnet runtime api proposals are implemented:
 - https://github.com/dotnet/runtime/issues/21941
 - https://github.com/dotnet/runtime/issues/24423
-
-Uses the same approach of calling NtQueryInformationProcess as in the PowerShell library
-https://github.com/PowerShell/PowerShell/blob/26f621952910e33840efb0c539fbef1e2a467a0d/src/System.Management.Automation/engine/ProcessCodeMethods.cs
 */
-internal partial class WindowsParentProcess {
+internal partial class ParentProcess {
```
cfbao, line 11: Consolidated both the previous Windows and Unix classes under the same one now.
The high level method `GetParentProcessName` has identical logic.

src/D2L.Bmx/ParentProcess.cs
```diff
@@ -2,43 +2,41 @@
...
+	public static string? GetParentProcessName() {
+		int parentPid = -1;
+		if( OperatingSystem.IsWindows() ) {
+			parentPid = GetWindowsParentProcessId();
+		} else if( OperatingSystem.IsLinux() || OperatingSystem.IsMacOS() ) {
+			parentPid = GetPosixParentProcessId();
+		}
+		if( parentPid == -1 ) {
+			return null;
+		}
```
cfbao, line 21: This is new.
`GetWindowsParentProcessId` returns -1 on error. We should check that, instead of passing it to `GetProcessById`, which will throw.

src/D2L.Bmx/ParentProcess.cs
```diff
@@ -2,43 +2,41 @@
...
+		return Process.GetProcessById( parentPid ).ProcessName;
+	}
 
-	[LibraryImport( "ntdll.dll", EntryPoint = "NtQueryInformationProcess" )]
-	internal static partial int NtQueryInformationProcess(
```
cfbao, line 16: Moved below, as per our convention to put public/high-level methods first, and private/low-level/helper methods after.
Also changed it to `private`, as it's not used elsewhere.

gord5500 (approved):

PR #474, opened 2024-07-26
replace RuntimeInformation.IsOSPlatform with OperatingSystem.Is*

cfbao:
`OperatingSystem.Is*` methods are newer and more efficient. They are basically constants.
A very rough check suggests that, with these new methods, the native AOT trimmer/linker can even eliminate OS-specific dead code!

See https://github.com/dotnet/aspnetcore/issues/24653

### DIFF
--- a/src/D2L.Bmx/Aws/AwsCredsCache.cs
+++ b/src/D2L.Bmx/Aws/AwsCredsCache.cs
@@ -1,4 +1,3 @@
-using System.Runtime.InteropServices;
 using System.Text.Json;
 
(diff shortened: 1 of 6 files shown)


gord5500 (approved):

boarnoah:
How did you find this?

Deprecation Warning isn't appropriate for this, but I wonder if something softer is there that .NET could have used to bring awareness.

cfbao:
@boarnoah We were already using both APIs inconsistently within BMX, so I wondered if we should consolidate on a single one, and searched for info specifically.

PR #475, opened 2024-07-26
don't print duration if --non-interactive supplied

cfbao:
Not a big deal and shouldn't affect any practical use cases, but at least for consistency sake, duration should be like other parameters - don't print if the `--non-interactive` flag is supplied on the command line.


boarnoah (approved):

gord5500 (approved):

cfbao:
Note for posterity:

This actually turns out to be important. Some AWS SDKs (unclear which ones) read both stdout & stderr from credential process indiscriminately. Printing the duration could cause auth failure.

PR #476, opened 2024-07-31
remove console IO hack

cfbao:
https://github.com/dotnet/runtime/issues/22314 is fixed in .NET 9, so we don't need these hacks now.

### Ticket

https://desire2learn.atlassian.net/browse/VUL-427

---

[hide whitespace changes](https://github.com/Brightspace/bmx/pull/476/files?w=1)


jharoutuniand2l (approved):
very nice

boarnoah (approved):

PR #477, opened 2024-08-16
make Okta API client stateless

cfbao:
### Why

The original `OktaApi` class is stateful, i.e. it matters in what order you call its methods. This makes it hard to use & error prone.

I've split up its methods and different states into three different classes, so each one is stateless and has a smaller scope.
When & how you can transition from one state (unauthed) to another (authed) is also enforced via the API contract (rather than  relying on knowledge of its underlying behaviour).

### Ticket

https://desire2learn.atlassian.net/browse/VUL-104

### DIFF
--- a/src/D2L.Bmx/Okta/OktaClient.cs
+++ b/src/D2L.Bmx/Okta/OktaClient.cs
@@ -1,62 +1,72 @@
 using System.Net;
-using System.Net.Http.Headers;
 using System.Net.Http.Json;
 using System.Text.Json;
 using D2L.Bmx.Okta.Models;
 
 namespace D2L.Bmx.Okta;
 
-internal interface IOktaApi {
-	void SetOrganization( string organization );
-	void AddSession( string sessionId );
+internal interface IOktaClientFactory {
+	IOktaAnonymousClient CreateAnonymousClient( string org );
+	IOktaAuthenticatedClient CreateAuthenticatedClient( string org, string sessionId );
+}
+
+internal interface IOktaAnonymousClient {
 	Task<AuthenticateResponse> AuthenticateAsync( string username, string password );
 	Task IssueMfaChallengeAsync( string stateToken, string factorId );
 	Task<AuthenticateResponse> VerifyMfaChallengeResponseAsync(
 		string stateToken,
 		string factorId,
-		string challengeResponse );
+		string challengeResponse
+	);
 	Task<OktaSession> CreateSessionAsync( string sessionToken );
-	Task<OktaApp[]> GetAwsAccountAppsAsync();
-	Task<string> GetPageAsync( string samlLoginUrl );
 }
 
-internal class OktaApi : IOktaApi {
-	private readonly CookieContainer _cookieContainer;
-	private readonly HttpClient _httpClient;
-	private string? _organization;
+internal interface IOktaAuthenticatedClient {
+	Task<OktaApp[]> GetAwsAccountAppsAsync();
+	Task<string> GetPageAsync( string url );
+}
 
-	public OktaApi() {
-		_cookieContainer = new CookieContainer();
-		_httpClient = new HttpClient( new HttpClientHandler { CookieContainer = _cookieContainer } ) {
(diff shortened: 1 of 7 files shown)


src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -1,62 +1,72 @@
...
 
 namespace D2L.Bmx.Okta;
 
-internal interface IOktaApi {
-	void SetOrganization( string organization );
-	void AddSession( string sessionId );
+internal interface IOktaClientFactory {
+	IOktaAnonymousClient CreateAnonymousClient( string org );
+	IOktaAuthenticatedClient CreateAuthenticatedClient( string org, string sessionId );
+}
```
cfbao, line 11: The most problematic methods in the original `OktaApi` class were `SetOrganization` & `AddSession`, as they are the ones manipulating internal state and has impact on when you can call which other methods.
Pulling them out into a factory (and hence naturally separating the anonymous & authed states) is the key to making the new classes stateless.

The new factory interface/class is also the key to understand the structure of the new code.
The rest is pretty mundane copy/paste.

src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -1,62 +1,72 @@
...
-		_cookieContainer = new CookieContainer();
-		_httpClient = new HttpClient( new HttpClientHandler { CookieContainer = _cookieContainer } ) {
-			Timeout = TimeSpan.FromSeconds( 30 )
+internal class OktaClientFactory : IOktaClientFactory {
+	IOktaAnonymousClient IOktaClientFactory.CreateAnonymousClient( string org ) {
+		var httpClient = new HttpClient {
+			Timeout = TimeSpan.FromSeconds( 30 ),
+			BaseAddress = GetBaseAddress( org ),
 		};
-		_httpClient.DefaultRequestHeaders.Accept.Add( new MediaTypeWithQualityHeaderValue( "application/json" ) );
```
cfbao, line 33: I didn't actually mean to remove it, but now that I noticed, I'm not sure I'd add it back:
* Okta API works fine without this header.
* This header is a lie when used in the [`GetPageAsync`](https://github.com/Brightspace/bmx/pull/477/files#diff-68472d07d2f2a8650bacbb76d8a2511a58d8d662adf585119f1299282196441cL181-R191) method (we expect HTML there)

src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -87,16 +98,12 @@ await resp.Content.ReadAsStreamAsync(),
 			);
 		}
 
-		string org = _organization ?? "unknown";
-		throw new BmxException( $"""
-			Okta authentication for user '{username}' in org '{org}' failed.
-			Check if org, user, and password is correct.
-			""" );
+		return new AuthenticateResponse.Failure( resp.StatusCode );
```
cfbao, line 101: This is new.

The `_organization` field was saved in the original `OktaApi` class for the sole purpose of constructing this error message. This isn't great, however.
The `OktaApi` / `Okta*Client` classes are meant to be low level and only care about talking to Okta. Constructing informative error messages with rich context info is more of a "business concern", and better left to higher level code, especially when not doing so complicates the low level classes (e.g. saving an otherwise useless extra field).

src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -137,13 +144,13 @@ await resp.Content.ReadAsStreamAsync(),
 		if( authnResponse?.SessionToken is not null ) {
 			return new AuthenticateResponse.Success( authnResponse.SessionToken );
 		}
-		throw new BmxException( "Error verifying MFA with Okta." );
+		return new AuthenticateResponse.Failure( resp.StatusCode );
```
cfbao, line 147: Similar in spirit to the comment above.
This one isn't as bad though, because it didn't make the code more complex. But changing anyway, at least for consistency.

scowing (approved):

PR #478, opened 2024-08-19
trim each line when printing update reminder

cfbao:
### Why

The update reminder's background area is misaligned on Windows:
image

This is because we didn't account for `\r` on Windows.
Still a mystery why this only happens with GitHub Actions' build but not with our local builds. Regardless, this fix works with both.

### Ticket

https://desire2learn.atlassian.net/browse/VUL-447

### DIFF
--- a/src/D2L.Bmx/ConsoleWriter.cs
+++ b/src/D2L.Bmx/ConsoleWriter.cs
@@ -35,7 +35,9 @@ internal class ConsoleWriter : IConsoleWriter {
 			Console.Error.WriteLine();
 			return;
 		}
-		string[] lines = text.Split( '\n' );
+		// Trim entries so we don't have extra `\r` characters on Windows.
+		// Splitting on `Environment.NewLine` isn't as safe, because we might also use `\n` on Windows.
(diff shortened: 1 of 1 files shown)


gord5500 (approved):

cfbao:
OMG... I know why.

Which newline string is used (LF vs CRLF) in raw string literals is determined by which one you use in the source file (well... they _are_ "literals").
E.g. given the source code
```csharp
string s = """
	1
	2
	""";
Console.WriteLine(
	Convert.ToHexString(System.Text.Encoding.UTF8.GetBytes(s))
);
```
the program will
* print `310A32` if the file uses LF as newline characters
* print `310D0A32` if the file uses CRLF as newline characters

On our dev machines, we set `core.autocrlf=false` in Git and have `end_of_line = lf` in .editorconfig, so we always have LF as newline characters.
GitHub Actions Windows runners seems to be using `core.autocrlf=true`, so all source files are cloned with CRLF line endings...(!!!) and hence the executables built has different behaviour, because it doesn't use equivalent source code!

PR #479, opened 2024-08-19
add .gitattribute to prevent line ending conversion on any machine

cfbao:
### Why

Git's line ending conversion has been tampering with our source code on GitHub Actions runners to be inequivalent to ones we use on our local machines.

Which line ending is used (LF vs CRLF) in raw string literals is determined by which one you use in the source file (well... they _are_ "literals").
E.g. given the source code
```csharp
string s = """
	1
	2
	""";
Console.WriteLine(s == "1\n2");
Console.WriteLine(s == "1\r\n2");
```
the program will
* print `True False` if the source file uses LF as newline characters
* print `False True` if the source file uses CRLF as newline characters

On our dev machines, we set `core.autocrlf=false` in Git and have `end_of_line = lf` in .editorconfig, so we always have LF as newline characters.
GitHub Actions Windows runners seems to be using `core.autocrlf=true`, so all source files were checked out with CRLF line endings...(!!!) and hence the executables built has different behaviour, because it doesn't use equivalent source code!

Unsetting the text attribute should disable line ending conversion entirely on any machine.

### Ticket

https://desire2learn.atlassian.net/browse/VUL-447


.gitattributes
```diff
@@ -0,0 +1 @@
+* -text
```
cfbao, line 1: I think this is unfortunate naming on Git's part.
The `text` attribute only controls line ending conversion. It doesn't affect whether Git treats the file as text or binary.

See:
https://git-scm.com/docs/gitattributes#_text
https://stackoverflow.com/a/10017566/5916310

boarnoah (approved):

PR #480, opened 2024-09-17
okta: add passwordless dsso support

gord5500:
### Why

It'll probably make a lot of people happy

This is going to be on by default and if any failure occurs, fall back to the password prompt.

Adding an additional flag `--experimental-bypass-browser-security` only for using bmx with elevated permissions. Chromium sandbox doesn't work without the `--no-sanbox` arg as admin. This can only be passed as a command line arg

File size jumped from 13.3 MB to 18.6 MB. Doesn't seem too bad?

### Ticket

VUL-453

### DIFF
--- a/src/D2L.Bmx/Browser.cs
+++ b/src/D2L.Bmx/Browser.cs
@@ -13,8 +13,8 @@ public class Browser {
 
 	// https://github.com/microsoft/playwright/blob/6763d5ab6bd20f1f0fc879537855a26c7644a496/packages/playwright-core/src/server/registry/index.ts#L457-L459
 	private static readonly string[] WindowsPartialPaths = [
-		"\\Google\\Chrome\\Application\\chrome.exe",
-		"\\Microsoft\\Edge\\Application\\msedge.exe"
+		"\\Microsoft\\Edge\\Application\\msedge.exe",
+		"\\Google\\Chrome\\Application\\chrome.exe"
 	];
 	private static readonly string[] MacPaths = [
 		"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
--- a/src/D2L.Bmx/Browser.cs
+++ b/src/D2L.Bmx/Browser.cs
@@ -32,7 +32,7 @@ public class Browser {
 		}
 
 		var launchOptions = new LaunchOptions {
-			Headless = false,
+			Headless = Environment.GetEnvironmentVariable( "BMX_DEBUG" ) != "1",
 			ExecutablePath = browserPath,
 			Args = noSandbox ? ["--no-sandbox"] : []
 		};
--- a/src/D2L.Bmx/Browser.cs
+++ b/src/D2L.Bmx/Browser.cs
@@ -32,7 +32,7 @@ public class Browser {
 		}
 
 		var launchOptions = new LaunchOptions {
-			Headless = true,
+			Headless = false,
 			ExecutablePath = browserPath,
 			Args = noSandbox ? ["--no-sandbox"] : []
 		};
--- a/src/D2L.Bmx/Browser.cs
+++ b/src/D2L.Bmx/Browser.cs
@@ -1,0 +1,64 @@
+using PuppeteerSharp;
+
+namespace D2L.Bmx;
+
+public class Browser {
+
+	// https://github.com/microsoft/playwright/blob/6763d5ab6bd20f1f0fc879537855a26c7644a496/packages/playwright-core/src/server/registry/index.ts#L630
+	private static readonly string[] WindowsEnvironmentVariables = [
+		"LOCALAPPDATA",
+		"PROGRAMFILES",
+		"PROGRAMFILES(X86)",
+	];
+
+	// https://github.com/microsoft/playwright/blob/6763d5ab6bd20f1f0fc879537855a26c7644a496/packages/playwright-core/src/server/registry/index.ts#L457-L459
+	private static readonly string[] WindowsPartialPaths = [
+		"\\Google\\Chrome\\Application\\chrome.exe",
+		"\\Microsoft\\Edge\\Application\\msedge.exe"
+	];
+	private static readonly string[] MacPaths = [
+		"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
+		"/Applications/Microsoft Edge.app/Contents/MacOS/Microsoft Edge"
+	];
+	private static readonly string[] LinuxPaths = [
+		"/opt/google/chrome/chrome",
+		"/opt/microsoft/msedge/msedge"
+	];
+
+	public static async Task<IBrowser?> LaunchBrowserAsync( bool noSandbox = false ) {
+		string? browserPath = GetPathToBrowser();
+		if( browserPath is null ) {
+			return null;
+		}
+
+		var launchOptions = new LaunchOptions {
+			Headless = true,
+			ExecutablePath = browserPath,
+			Args = noSandbox ? ["--no-sandbox"] : []
+		};
+
+		return await Puppeteer.LaunchAsync( launchOptions );
+	}
+
+	private static string? GetPathToBrowser() {
+		string? browser = null;
+		if( OperatingSystem.IsWindows() ) {
+			foreach( string windowsPartialPath in WindowsPartialPaths ) {
+				foreach( string environmentVariable in WindowsEnvironmentVariables ) {
+					string? prefix = Environment.GetEnvironmentVariable( environmentVariable );
+					if( prefix is not null ) {
+						string path = prefix + windowsPartialPath;
+						if( File.Exists( path ) ) {
+							return path;
+						}
+					}
+				}
+			}
+		} else if( OperatingSystem.IsMacOS() ) {
+			return MacPaths.First( File.Exists );
+		} else if( OperatingSystem.IsLinux() ) {
+			return LinuxPaths.First( File.Exists );
+		}
+		return browser;
+	}
+}
--- a/src/D2L.Bmx/Browser.cs
+++ b/src/D2L.Bmx/Browser.cs
@@ -14,15 +14,15 @@ public static class Browser {
 	// https://github.com/microsoft/playwright/blob/6763d5ab6bd20f1f0fc879537855a26c7644a496/packages/playwright-core/src/server/registry/index.ts#L457-L459
 	private static readonly string[] WindowsPartialPaths = [
 		"\\Microsoft\\Edge\\Application\\msedge.exe",
-		"\\Google\\Chrome\\Application\\chrome.exe"
+		"\\Google\\Chrome\\Application\\chrome.exe",
 	];
 	private static readonly string[] MacPaths = [
 		"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
-		"/Applications/Microsoft Edge.app/Contents/MacOS/Microsoft Edge"
+		"/Applications/Microsoft Edge.app/Contents/MacOS/Microsoft Edge",
 	];
 	private static readonly string[] LinuxPaths = [
 		"/opt/google/chrome/chrome",
-		"/opt/microsoft/msedge/msedge"
+		"/opt/microsoft/msedge/msedge",
 	];
 
 	public static async Task<IBrowser?> LaunchBrowserAsync( bool noSandbox = false ) {
@@ -32,7 +32,6 @@ public static class Browser {
 		}
 
 		var launchOptions = new LaunchOptions {
-			Headless = true,
 			ExecutablePath = browserPath,
 			Args = noSandbox ? ["--no-sandbox"] : []
 		};
@@ -41,7 +40,6 @@ public static class Browser {
 	}
 
 	private static string? GetPathToBrowser() {
-		string? browser = null;
 		if( OperatingSystem.IsWindows() ) {
 			foreach( string windowsPartialPath in WindowsPartialPaths ) {
 				foreach( string environmentVariable in WindowsEnvironmentVariables ) {
@@ -59,6 +57,6 @@ public static class Browser {
 		} else if( OperatingSystem.IsLinux() ) {
 			return LinuxPaths.First( File.Exists );
 		}
-		return browser;
+		return null;
 	}
 }
--- a/src/D2L.Bmx/Browser.cs
+++ b/src/D2L.Bmx/Browser.cs
@@ -32,7 +32,7 @@ public class Browser {
 		}
 
 		var launchOptions = new LaunchOptions {
-			Headless = Environment.GetEnvironmentVariable( "BMX_DEBUG" ) != "1",
+			Headless = true,
 			ExecutablePath = browserPath,
 			Args = noSandbox ? ["--no-sandbox"] : []
 		};
--- a/src/D2L.Bmx/Browser.cs
+++ b/src/D2L.Bmx/Browser.cs
@@ -2,7 +2,7 @@ using PuppeteerSharp;
 
 namespace D2L.Bmx;
 
-public class Browser {
+public static class Browser {
 
 	// https://github.com/microsoft/playwright/blob/6763d5ab6bd20f1f0fc879537855a26c7644a496/packages/playwright-core/src/server/registry/index.ts#L630
 	private static readonly string[] WindowsEnvironmentVariables = [
--- a/src/D2L.Bmx/Browser.cs
+++ b/src/D2L.Bmx/Browser.cs
@@ -13,8 +13,8 @@ public static class Browser {
 
 	// https://github.com/microsoft/playwright/blob/6763d5ab6bd20f1f0fc879537855a26c7644a496/packages/playwright-core/src/server/registry/index.ts#L457-L459
 	private static readonly string[] WindowsPartialPaths = [
-		"\\Microsoft\\Edge\\Application\\msedge.exe",
-		"\\Google\\Chrome\\Application\\chrome.exe",
+		"Microsoft\\Edge\\Application\\msedge.exe",
+		"Google\\Chrome\\Application\\chrome.exe",
 	];
 	private static readonly string[] MacPaths = [
 		"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
@@ -45,7 +45,7 @@ public static class Browser {
 				foreach( string environmentVariable in WindowsEnvironmentVariables ) {
 					string? prefix = Environment.GetEnvironmentVariable( environmentVariable );
 					if( prefix is not null ) {
-						string path = prefix + windowsPartialPath;
+						string path = Path.Join( prefix, windowsPartialPath );
 						if( File.Exists( path ) ) {
 							return path;
 						}
@@ -53,9 +53,9 @@ public static class Browser {
 				}
 			}
 		} else if( OperatingSystem.IsMacOS() ) {
-			return MacPaths.First( File.Exists );
+			return Array.Find( MacPaths, File.Exists );
 		} else if( OperatingSystem.IsLinux() ) {
-			return LinuxPaths.First( File.Exists );
+			return Array.Find( LinuxPaths, File.Exists );
 		}
 		return null;
 	}
--- a/src/D2L.Bmx/LoginHandler.cs
+++ b/src/D2L.Bmx/LoginHandler.cs
@@ -6,14 +6,22 @@ internal class LoginHandler(
 	public async Task HandleAsync(
 		string? org,
 		string? user,
-		bool experimental
+		bool experimental,
+		bool? passwordless
 	) {
 		if( !File.Exists( BmxPaths.CONFIG_FILE_NAME ) ) {
 			throw new BmxException(
 				"BMX global config file not found. Okta sessions will not be saved. Please run `bmx configure` first."
 			);
 		}
-		await oktaAuth.AuthenticateAsync( org, user, nonInteractive: false, ignoreCache: true, experimental: experimental );
+		await oktaAuth.AuthenticateAsync(
+			org,
+			user,
+			nonInteractive: false,
+			ignoreCache: true,
+			experimental: experimental,
+			passwordless: passwordless
+		);
 		Console.WriteLine( "Successfully logged in and Okta session has been cached." );
 	}
 }
--- a/src/D2L.Bmx/LoginHandler.cs
+++ b/src/D2L.Bmx/LoginHandler.cs
@@ -6,7 +6,7 @@ internal class LoginHandler(
 	public async Task HandleAsync(
 		string? org,
 		string? user,
-		bool experimental
+		bool experimentalBypassBrowserSecurity
 	) {
 		if( !File.Exists( BmxPaths.CONFIG_FILE_NAME ) ) {
 			throw new BmxException(
@@ -18,7 +18,7 @@ internal class LoginHandler(
 			user,
 			nonInteractive: false,
 			ignoreCache: true,
-			experimentalBypassBrowserSecurity: experimental
+			experimentalBypassBrowserSecurity: experimentalBypassBrowserSecurity
 		);
 		Console.WriteLine( "Successfully logged in and Okta session has been cached." );
 	}
--- a/src/D2L.Bmx/LoginHandler.cs
+++ b/src/D2L.Bmx/LoginHandler.cs
@@ -6,8 +6,7 @@ internal class LoginHandler(
 	public async Task HandleAsync(
 		string? org,
 		string? user,
-		bool experimental,
-		bool? passwordless
+		bool experimental
 	) {
 		if( !File.Exists( BmxPaths.CONFIG_FILE_NAME ) ) {
 			throw new BmxException(
@@ -19,8 +18,7 @@ internal class LoginHandler(
 			user,
 			nonInteractive: false,
 			ignoreCache: true,
-			experimental: experimental,
-			passwordless: passwordless
+			experimentalBypassBrowserSecurity: experimental
 		);
 		Console.WriteLine( "Successfully logged in and Okta session has been cached." );
 	}
--- a/src/D2L.Bmx/LoginHandler.cs
+++ b/src/D2L.Bmx/LoginHandler.cs
@@ -5,14 +5,15 @@ internal class LoginHandler(
 ) {
 	public async Task HandleAsync(
 		string? org,
-		string? user
+		string? user,
+		bool experimental
 	) {
 		if( !File.Exists( BmxPaths.CONFIG_FILE_NAME ) ) {
 			throw new BmxException(
 				"BMX global config file not found. Okta sessions will not be saved. Please run `bmx configure` first."
 			);
 		}
-		await oktaAuth.AuthenticateAsync( org, user, nonInteractive: false, ignoreCache: true );
+		await oktaAuth.AuthenticateAsync( org, user, nonInteractive: false, ignoreCache: true, experimental: experimental );
 		Console.WriteLine( "Successfully logged in and Okta session has been cached." );
 	}
 }
--- a/src/D2L.Bmx/LoginHandler.cs
+++ b/src/D2L.Bmx/LoginHandler.cs
@@ -6,7 +6,7 @@ internal class LoginHandler(
 	public async Task HandleAsync(
 		string? org,
 		string? user,
-		bool experimentalBypassBrowserSecurity
+		bool bypassBrowserSecurity
 	) {
 		if( !File.Exists( BmxPaths.CONFIG_FILE_NAME ) ) {
 			throw new BmxException(
@@ -18,7 +18,7 @@ internal class LoginHandler(
 			user,
 			nonInteractive: false,
 			ignoreCache: true,
-			experimentalBypassBrowserSecurity: experimentalBypassBrowserSecurity
+			bypassBrowserSecurity: bypassBrowserSecurity
 		);
 		Console.WriteLine( "Successfully logged in and Okta session has been cached." );
 	}
--- a/src/D2L.Bmx/Okta/OktaClient.cs
+++ b/src/D2L.Bmx/Okta/OktaClient.cs
@@ -23,7 +23,7 @@ internal interface IOktaAnonymousClient {
 
 internal interface IOktaAuthenticatedClient {
 	Task<OktaApp[]> GetAwsAccountAppsAsync();
-	Task<OktaSession> GetSessionExpiryAsync();
+	Task<OktaSession> GetCurrentOktaSessionAsync();
 	Task<string> GetPageAsync( string url );
 }
 
@@ -188,7 +188,7 @@ internal class OktaAuthenticatedClient( HttpClient httpClient ) : IOktaAuthentic
 				?? throw new BmxException( "Error retrieving AWS accounts from Okta." );
 	}
 
-	async Task<OktaSession> IOktaAuthenticatedClient.GetSessionExpiryAsync() {
+	async Task<OktaSession> IOktaAuthenticatedClient.GetCurrentOktaSessionAsync() {
 		OktaSession? session;
 		try {
 			session = await httpClient.GetFromJsonAsync(
--- a/src/D2L.Bmx/Okta/OktaClient.cs
+++ b/src/D2L.Bmx/Okta/OktaClient.cs
@@ -23,7 +23,7 @@ internal interface IOktaAnonymousClient {
 
 internal interface IOktaAuthenticatedClient {
 	Task<OktaApp[]> GetAwsAccountAppsAsync();
-	Task<DateTimeOffset> GetSessionExpiryAsync();
+	Task<OktaSession> GetSessionExpiryAsync();
 	Task<string> GetPageAsync( string url );
 }
 
@@ -188,7 +188,7 @@ internal class OktaAuthenticatedClient( HttpClient httpClient ) : IOktaAuthentic
 				?? throw new BmxException( "Error retrieving AWS accounts from Okta." );
 	}
 
-	async Task<DateTimeOffset> IOktaAuthenticatedClient.GetSessionExpiryAsync() {
+	async Task<OktaSession> IOktaAuthenticatedClient.GetSessionExpiryAsync() {
 		OktaSession? session;
 		try {
 			session = await httpClient.GetFromJsonAsync(
@@ -198,7 +198,7 @@ internal class OktaAuthenticatedClient( HttpClient httpClient ) : IOktaAuthentic
 			throw new BmxException( "Request to retrieve session expiry from Okta failed.", ex );
 		}
 
-		return session?.ExpiresAt ?? throw new BmxException( "Error retrieving session expiry from Okta." );
+		return session ?? throw new BmxException( "Error retrieving session expiry from Okta." );
 	}
 
 	async Task<string> IOktaAuthenticatedClient.GetPageAsync( string url ) {
--- a/src/D2L.Bmx/Okta/OktaClient.cs
+++ b/src/D2L.Bmx/Okta/OktaClient.cs
@@ -23,6 +23,7 @@ internal interface IOktaAnonymousClient {
 
 internal interface IOktaAuthenticatedClient {
 	Task<OktaApp[]> GetAwsAccountAppsAsync();
+	Task<DateTimeOffset> GetSessionExpiryAsync();
 	Task<string> GetPageAsync( string url );
 }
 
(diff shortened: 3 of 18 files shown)


src/D2L.Bmx/ParameterDescriptions.cs
```diff
@@ -18,4 +18,6 @@ internal static class ParameterDescriptions {
 		Write BMX command to AWS profile, so that AWS tools & SDKs using the profile will source credentials from BMX.
 		See https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html.
 		""";
+
+	public const string Experimental = "Enables experimental features";
```
gord5500, line 22: More detail here?

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +136,112 @@ private bool TryAuthenticateFromCache(
...
+
+		Console.WriteLine( "Attempting to automatically login using DSSO." );
+		var cancellationTokenSource = new CancellationTokenSource( TimeSpan.FromSeconds( 10 ) );
+		var sessionIdTaskProducer = new TaskCompletionSource<string?>( TaskCreationOptions.RunContinuationsAsynchronously );
+		var userEmailTaskProducer = new TaskCompletionSource<string?>( TaskCreationOptions.RunContinuationsAsynchronously );
+		string? sessionId;
+		string? userEmail;
+
+		try {
+			var page = await browser.NewPageAsync().WaitAsync( cancellationTokenSource.Token );
```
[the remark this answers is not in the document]
gord5500: I'm going to try testing it more though so we can be more confident that won't happen again

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +136,112 @@ private bool TryAuthenticateFromCache(
...
+			cancellationTokenSource.Dispose();
+			browser.Dispose();
+		}
+
+		if( sessionId is null || userEmail is null ) {
+			return null;
+		} else if( !OktaUserMatchesProvided( userEmail, user ) ) {
+			consoleWriter.WriteWarning(
+				"WARNING: Could not create Okta session using DSSO as "
+				+ $"provided Okta user '{user}' does not match user '{userEmail}'." );
```
gord5500, line 227: Potentially confusing if `user` doesn't have the `@desire2learn.com` included even though we are not including that in our comparison.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +136,112 @@ private bool TryAuthenticateFromCache(
 		return true;
 	}
 
+	private async Task<IOktaAuthenticatedClient?> TryAuthenticateWithDSSOAsync(
```
gord5500, line 139: Not throwing any exceptions here since I think we'd always want to fall back on prompting for password. 

Wondering if maybe there are some setups where the warning messages could be annoying for people. Jumpbox is probably one of them? Was thinking we could put a new value in the config that disables passwordless which by default would be false?

gord5500: Experimental would stay as an arg only option

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +136,114 @@ private bool TryAuthenticateFromCache(
...
+			}
+
+			async Task GetOktaUserEmailAsync(
+				IResponse response
+			) {
+				if( response.Url.Contains( $"{baseAddress}enduser/api/v1/home" ) ) {
+					string content = await response.TextAsync();
+					var home = JsonSerializer.Deserialize( content, JsonCamelCaseContext.Default.OktaHomeResponse );
+					if( home is not null ) {
+						userEmailTaskProducer.SetResult( home.Login );
```
gord5500, line 197: We can also get this from `sessions/me` route. Was thinking as a design pov we shouldn't create the authenticated client object unless we know it's the same person. But if that doesn't matter this can go and we just create the client then verify the user email afterwards

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +136,114 @@ private bool TryAuthenticateFromCache(
...
+			await page.GoToAsync( baseAddress );
+			sessionId = await sessionIdTaskProducer.Task.WaitAsync( cancellationTokenSource.Token );
+			userEmail = await userEmailTaskProducer.Task.WaitAsync( cancellationTokenSource.Token );
+
+			async Task GetSessionCookieAsync() {
+				var url = new Uri( page.Url );
+				if( url.Host == $"{org}.okta.com" ) {
+					string title = await page.GetTitleAsync();
+					if( title.Contains( "sign in", StringComparison.OrdinalIgnoreCase ) ) {
+						if( attempt < 3 && url.AbsolutePath == "/" ) {
```
gord5500, line 175: `==` is different than the poc `!=`. I found that for our sign in page at least that the absolute path was `/`

src/D2L.Bmx/OktaAuthenticator.cs
gord5500: I suppose we have to add this as otherwise jumpbox users will be forced to wait for the timeout

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +143,121 @@ private bool TryAuthenticateFromCache(
...
+			await page.GoToAsync( baseAddress );
+			sessionId = await sessionIdTaskProducer.Task.WaitAsync( cancellationTokenSource.Token );
+			userEmail = await userEmailTaskProducer.Task.WaitAsync( cancellationTokenSource.Token );
+
+			async Task GetSessionCookieAsync() {
+				var url = new Uri( page.Url );
+				if( url.Host == $"{org}.okta.com" ) {
+					string title = await page.GetTitleAsync();
+					if( title.Contains( "sign in", StringComparison.OrdinalIgnoreCase ) ) {
+						if( attempt < 3 ) {
```
gord5500, line 182: poc had a check for absolute path not being `/`. I ran in to scenarios where it was both `/` and some long value

gord5500: Maybe possible I had vpn off when that happened. I'm adding it back for now

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +143,121 @@ private bool TryAuthenticateFromCache(
...
+					if( home is not null ) {
+						userEmailTaskProducer.SetResult( home.Login );
+					}
+				}
+			}
+
+		} catch( TaskCanceledException ) {
+			consoleWriter.WriteWarning(
+				$"WARNING: Timed out when trying to create {org} Okta session through DSSO. Check if org is correct."
+					+ " If running BMX with admin privileges, rerun the command with the '--experimental' flag."
```
gord5500, line 212: This maybe also needs some mention about being on the vpn. I don't know how to make that org agnostic though.

Maybe `If your Okta org requires a VPN connection, ensure it's active`? There's already a lot of text here

src/D2L.Bmx/ParameterDescriptions.cs
gord5500: And potentially being dangerous

src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -23,6 +23,7 @@ string challengeResponse
 
 internal interface IOktaAuthenticatedClient {
 	Task<OktaApp[]> GetAwsAccountAppsAsync();
+	Task<DateTimeOffset> GetSessionExpiryAsync();
```
gord5500, line 26: Looking through the chrome traffic I couldn't find a value for this

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +143,121 @@ private bool TryAuthenticateFromCache(
...
+		string user,
+		IOktaClientFactory oktaClientFactory,
+		bool experimental
+	) {
+		await using IBrowser? browser = await Browser.LaunchBrowserAsync( experimental );
+		if( browser is null ) {
+			return null;
+		}
+
+		Console.WriteLine( "Attempting to automatically login using DSSO." );
```
gord5500, line 157: Maybe I should replace all the places I use `DSSO` with `passwordless` since people probably won't know what DSSO means

gord5500: For any of the println messages

boarnoah:
> Time to log in is kind of slow. I timed a few and they were between 5-9 seconds. I can definitely type in my password faster than that but it's probably still something people would prefer.

Not too fussed, its a once a day activity + anything that encourages people to not type their password into more places than they really need to is a win.

src/D2L.Bmx/Browser.cs
```diff
@@ -0,0 +1,64 @@
...
+	];
+
+	public static async Task<IBrowser?> LaunchBrowserAsync( bool noSandbox = false ) {
+		string? browserPath = GetPathToBrowser();
+		if( browserPath is null ) {
+			return null;
+		}
+
+		var launchOptions = new LaunchOptions {
+			Headless = Environment.GetEnvironmentVariable( "BMX_DEBUG" ) != "1",
```
boarnoah, line 35: So when you run it with ``BMX_DEBUG = 1`` wouldn't the chromium instance automatically get stopped once puppeteer is done with getting sessions.

Any situations when it would remain open long enough to do any real investigation?

gord5500: > wouldn't the chromium instance automatically get stopped once puppeteer

Yea it would close after it gets the session or the 15 second timeout

> Any situations when it would remain open long enough to do any real investigation?

Yea I'm not sure how valuable it could be. Knowing where it got stuck in the login process if it was the login screen or some other page could maybe give some insight

gord5500: Just the fact that currently there are no exceptions being thrown I feel like it might be hard to know what went wrong. I could check the env variable again in that generic exception catch though and also print the exception

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,101 @@ private bool TryAuthenticateFromCache(
...
+		IOktaClientFactory oktaClientFactory,
+		bool experimentalBypassBrowserSecurity
+	) {
+		await using IBrowser? browser = await Browser.LaunchBrowserAsync( experimentalBypassBrowserSecurity );
+		if( browser is null ) {
+			return null;
+		}
+
+		Console.WriteLine( "Attempting to automatically login using DSSO." );
+		string normalizedOrg = org.Replace( ".okta.com", "" );
```
gord5500, line 155: Similar amount of effort to how we do it in the [okta clients](https://github.com/Brightspace/bmx/blob/551f6cf7b1e0a3bdce7671706ceb102b848a5ed1/src/D2L.Bmx/Okta/OktaClient.cs#L55-L59)

cfbao: They are not equivalent.
Note: https://d2l.slack.com/archives/G4P099831/p1726600460156129?thread_ts=1726597940.542299&cid=G4P099831

cfbao: Ideally we can reuse the same logic in both places instead of writing it twice.

gord5500: Yea I didn't realize what you meant by that comment initially. I switched it

src/D2L.Bmx/Browser.cs
```diff
@@ -0,0 +1,64 @@
+using PuppeteerSharp;
+
+namespace D2L.Bmx;
+
+public class Browser {
```
boarnoah, line 5: nit: make it clear this is for utility stuff with no internal state

```suggestion
public static class Browser {
```

gord5500:
Time from beginning to open the browser to having the authenticated client with headless is 3-5 seconds for me

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
...
+				with elevated privileges. Rerun the command with the '--experimental-bypass-browser-security' flag.
+				"""
+			);
+			return null;
+		} catch( Exception ) {
+			consoleWriter.WriteWarning( "WARNING: Unknown error while trying to authenticate with Okta using DSSO." );
+			return null;
+		} finally {
+			cancellationTokenSource.Dispose();
+			browser.Dispose();
```
boarnoah, line 212: nit: Browser is being initialized here with using statement syntax https://github.com/Brightspace/bmx/blob/40171e4acf81ad2da538ffe6a1e6f396a8e3186f/src/D2L.Bmx/OktaAuthenticator.cs#L149

The whole point of which is to ensure it will get disposed of properly automatically. Doing this shouldn't be necessary

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
...
+				var cookies = await page.GetCookiesAsync( baseAddress );
+				if( Array.Find( cookies, c => c.Name == "sid" )?.Value is string sid ) {
+					sessionIdTaskProducer.SetResult( sid );
+				}
+			}
+		} catch( TaskCanceledException ) {
+			consoleWriter.WriteWarning( $"""
+				WARNING: Timed out when trying to create Okta session through DSSO.
+				Check if the org '{org}' is correct. If running BMX with elevated privileges,
+				rerun the command with the '--experimental-bypass-browser-security' flag
```
boarnoah, line 196: Sorry I might have missed this detail in the PoCs, Where if you launch chromium while elevated, it could be either one of these exceptions (this or ``TargetClosedException``) ?

gord5500: Yea it can be either and I'm not sure what scenario causes each to occur. The `TargetClosedException` I've only seen happen when using admin + not including `--no-sandbox` and nothing else

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
...
+
+		if( sessionId is null ) {
+			return null;
+		}
+
+		var oktaAuthenticatedClient = oktaClientFactory.CreateAuthenticatedClient( org, sessionId );
+		var sessionExpiry = ( await oktaAuthenticatedClient.GetCurrentOktaSessionAsync() ).ExpiresAt;
+		// We can expect a 404 if the session does not belong to the user which will throw an exception
+		try {
+			string userResponse = await oktaAuthenticatedClient.GetPageAsync( $"users/{user}" );
```
cfbao, line 223: https://d2l.slack.com/archives/G4P099831/p1726600460156129?thread_ts=1726597940.542299&cid=G4P099831
meant to say that we're probably better off with the "@" stripping logic

src/D2L.Bmx/Browser.cs
```diff
@@ -0,0 +1,64 @@
...
+			Headless = true,
+			ExecutablePath = browserPath,
+			Args = noSandbox ? ["--no-sandbox"] : []
+		};
+
+		return await Puppeteer.LaunchAsync( launchOptions );
+	}
+
+	private static string? GetPathToBrowser() {
+		string? browser = null;
```
cfbao, line 44: nit: this isn't really used. you can just `return null` at the end
```suggestion
```

src/D2L.Bmx/LoginHandler.cs
```diff
@@ -5,14 +5,21 @@ OktaAuthenticator oktaAuth
...
 				"BMX global config file not found. Okta sessions will not be saved. Please run `bmx configure` first."
 			);
 		}
-		await oktaAuth.AuthenticateAsync( org, user, nonInteractive: false, ignoreCache: true );
+		await oktaAuth.AuthenticateAsync(
+			org,
+			user,
+			nonInteractive: false,
+			ignoreCache: true,
+			experimentalBypassBrowserSecurity: experimentalBypassBrowserSecurity
```
cfbao, line 21: nit: the CLI flag starts with `--experimental-` but the internal variable/parameter names can maybe be shortened to not include "experimental" everywhere?

src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -187,6 +188,19 @@ async Task<OktaApp[]> IOktaAuthenticatedClient.GetAwsAccountAppsAsync() {
...
 	}
 
+	async Task<OktaSession> IOktaAuthenticatedClient.GetCurrentOktaSessionAsync() {
+		OktaSession? session;
+		try {
+			session = await httpClient.GetFromJsonAsync(
+				"sessions/me",
+				JsonCamelCaseContext.Default.OktaSession );
+		} catch( Exception ex ) {
+			throw new BmxException( "Request to retrieve session expiry from Okta failed.", ex );
```
cfbao, line 198:
```suggestion
			throw new BmxException( "Request to retrieve session from Okta failed.", ex );
```
Similarly below.
Also consider rewording message to not say "request", which is a relatively technical term.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
 		return true;
 	}
 
+	private async Task<IOktaAuthenticatedClient?> TryAuthenticateWithDSSOAsync(
```
cfbao, line 143: This doesn't look idiomatic to me.
I'd expect a `Try...` method to return a Boolean or something like a "result" object. Using "nullness" to indicate failure isn't the best API design.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
...
+		string user,
+		IOktaClientFactory oktaClientFactory,
+		bool experimentalBypassBrowserSecurity
+	) {
+		await using IBrowser? browser = await Browser.LaunchBrowserAsync( experimentalBypassBrowserSecurity );
+		if( browser is null ) {
+			return null;
+		}
+
+		Console.WriteLine( "Attempting to automatically login using DSSO." );
```
cfbao, line 154: * Informational message should be written to stderr. Writing to stdout messes up `bmx print`
* We shouldn't write any informational message when `--non-interactive` is passed, otherwise credential process can fail (depending on the AWS client)

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
...
+		var cancellationTokenSource = new CancellationTokenSource( TimeSpan.FromSeconds( 15 ) );
+		var sessionIdTaskProducer = new TaskCompletionSource<string?>( TaskCreationOptions.RunContinuationsAsynchronously );
+		string? sessionId;
+
+		try {
+			var page = await browser.NewPageAsync().WaitAsync( cancellationTokenSource.Token );
```
cfbao, line 160: `IPage` is an `IDisposable`. A `using` statement/declaration would be appropriate.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
...
+			string baseAddress = org.Contains( '.' )
+				? $"https://{org}/"
+				: $"https://{org}.okta.com/";
+			var baseUrl = new Uri( baseAddress );
```
cfbao, line 164: https://github.com/Brightspace/bmx/pull/480#discussion_r1764035285
I was thinking we'd construct the Okta base URL once in OktaAuthenticator and pass it to the oktaClientFactory and the browser, instead of (re)constructing it from the user-supplied "org" value every time.

src/D2L.Bmx/OktaAuthenticator.cs
cfbao: nit: we don't have an agreed upon convention, but personally I'm of the opinion that acronyms & initialisms should be formatted like regular words in PascalCase & camelCase, i.e. `TryAuthenticateWithDssoAsync`.

See also names in .NET standard libraries:
[`HttpClient`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-8.0), [`TcpClient`](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.tcpclient?view=net-8.0), [`Dns`](https://learn.microsoft.com/en-us/dotnet/api/system.net.dns?view=net-8.0) [`AesGcm`](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.aesgcm?view=net-8.0), [`Utf8`](https://learn.microsoft.com/en-us/dotnet/api/system.text.unicode.utf8?view=net-8.0)

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
...
+		bool experimentalBypassBrowserSecurity
+	) {
+		await using IBrowser? browser = await Browser.LaunchBrowserAsync( experimentalBypassBrowserSecurity );
+		if( browser is null ) {
+			return null;
+		}
+
+		Console.WriteLine( "Attempting to automatically login using DSSO." );
+		var cancellationTokenSource = new CancellationTokenSource( TimeSpan.FromSeconds( 15 ) );
+		var sessionIdTaskProducer = new TaskCompletionSource<string?>( TaskCreationOptions.RunContinuationsAsynchronously );
```
cfbao, line 156: "Producer" isn't really the most appropriate name.
I'd expect a "producer" to be like a factory which can create more and more items.
"TaskCompletionSource" isn't too niche a thing in .NET. They're often just named "tcs" in .NET source code
```suggestion
		var sessionIdTcs = new TaskCompletionSource<string?>( TaskCreationOptions.RunContinuationsAsynchronously );
```

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
...
+			var page = await browser.NewPageAsync().WaitAsync( cancellationTokenSource.Token );
+			string baseAddress = org.Contains( '.' )
+				? $"https://{org}/"
+				: $"https://{org}.okta.com/";
+			var baseUrl = new Uri( baseAddress );
+			int attempt = 1;
+
+			page.Load += ( _, _ ) => _ = GetSessionCookieAsync();
+			await page.GoToAsync( baseAddress );
+			sessionId = await sessionIdTaskProducer.Task.WaitAsync( cancellationTokenSource.Token );
```
cfbao, line 169: Passing a cancellation token to `WaitAsync` doesn't actually cancel the original task.
We should always cancel the original thing properly whenever possible.
In this case, we should pass the cancellation token to the `GetSessionCookieAsync` function.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
...
+
+			async Task GetSessionCookieAsync() {
+				var url = new Uri( page.Url );
+				if( url.Host == baseUrl.Host ) {
+					string title = await page.GetTitleAsync();
+					// DSSO can sometimes takes more than one attempt.
+					// If the path is '/', it means DSSO is not available and we should stop retrying.
```
cfbao, line 176: should clarify that it's only when we're at `/` _**and**_ we see the sign in page that we know DSSO isn't available.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +140,103 @@ private bool TryAuthenticateFromCache(
...
+			string userResponse = await oktaAuthenticatedClient.GetPageAsync( $"users/{user}" );
+		} catch( Exception ) {
+			consoleWriter.WriteWarning(
+				$"WARNING: Failed to create Okta session through DSSO as created session does not belong to {user}." );
+			return null;
+		}
+		if( File.Exists( BmxPaths.CONFIG_FILE_NAME ) ) {
+			CacheOktaSession( user, org, sessionId, sessionExpiry );
+		} else {
+			consoleWriter.WriteWarning( """
```
cfbao, line 232: nit: can we consolidate this with the other place where we have the same logic into a shared private method?

src/D2L.Bmx/ParameterDescriptions.cs
```diff
@@ -18,4 +18,7 @@ internal static class ParameterDescriptions {
 		Write BMX command to AWS profile, so that AWS tools & SDKs using the profile will source credentials from BMX.
 		See https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html.
 		""";
+
+	public const string ExperimentalBypassBrowserSecurity = "Enables experimental features";
```
cfbao, line 22: need more specific description

src/D2L.Bmx/ParameterDescriptions.cs
```diff
@@ -18,4 +18,7 @@ internal static class ParameterDescriptions {
...
+	public const string Passwordless = "Use Okta DSSO to attempt to authenticate without providing a password";
```
cfbao, line 23: Definitely don't say "DSSO" in any user-facing message. Nobody knows what it means.

cfbao (changes requested):

src/D2L.Bmx/OktaAuthenticator.cs
cfbao: yeah
https://github.com/Brightspace/bmx/pull/480#discussion_r1765459191

src/D2L.Bmx/Okta/OktaClient.cs
cfbao: ah I see. Nvm then. I'd rather we worded it differently, but no need to bother right now.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -191,11 +196,20 @@ async Task GetSessionCookieAsync() {
...
 					}
 				}
-				var cookies = await page.GetCookiesAsync( orgBaseAddress ).WaitAsync( cancellationTokenSource.Token );
+				var cookies = await page.GetCookiesAsync( orgBaseAddress );
 				if( Array.Find( cookies, c => c.Name == "sid" )?.Value is string sid ) {
 					sessionIdTcs.SetResult( sid );
 				}
 			}
+			async Task GetOktaUserEmailAsync( IResponse response ) {
+				if( response.Url.Contains( $"{orgBaseAddress}enduser/api/v1/home" ) ) {
```
cfbao, line 203: we still shouldn't use enduser endpoints though.
the sessions/me endpoint gives you the same login name, right?

cfbao: I don't think it hurts to create an authenticated client first before verifying the username

src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -23,20 +23,21 @@ string challengeResponse
 
 internal interface IOktaAuthenticatedClient {
 	Task<OktaApp[]> GetAwsAccountAppsAsync();
+	Task<OktaSession> GetCurrentOktaSessionAsync();
 	Task<string> GetPageAsync( string url );
 }
 
 internal class OktaClientFactory : IOktaClientFactory {
 	IOktaAnonymousClient IOktaClientFactory.CreateAnonymousClient( string org ) {
```
cfbao, line 31: nit: now that you're changing the meaning of the `org` parameter, best to rename it appropriately too, e.g. `orgUrl`
Can also consider changing the type to `Uri` to better enforce this.

src/D2L.Bmx/Okta/OktaClient.cs
```diff
@@ -51,10 +52,8 @@ IOktaAuthenticatedClient IOktaClientFactory.CreateAuthenticatedClient( string or
 		return new OktaAuthenticatedClient( httpClient );
 	}
 
-	private static Uri GetBaseAddress( string org ) {
-		return org.Contains( '.' )
-			? new Uri( $"https://{org}/api/v1/" )
-			: new Uri( $"https://{org}.okta.com/api/v1/" );
+	private static Uri GetApiBaseAddress( string org ) {
+		return new Uri( $"{org}api/v1/" );
```
cfbao, line 56: Would prefer a safer way to concat URL parts.
Here you're implicitly requiring that the `org` passed in ends in `/`

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -130,6 +143,129 @@ private bool TryAuthenticateFromCache(
...
+		} else if( !OktaUserMatchesProvided( userEmail, user ) ) {
+			consoleWriter.WriteWarning(
+				"WARNING: Could not create Okta session using Desktop Single Sign-On as "
+				+ $"provided Okta user '{StripLoginDomain( user )}' does not match user '{StripLoginDomain( userEmail )}'." );
+			return null;
+		}
+
+		var oktaAuthenticatedClient = oktaClientFactory.CreateAuthenticatedClient( orgBaseAddress, sessionId );
+		var expiresAt = ( await oktaAuthenticatedClient.GetCurrentOktaSessionAsync() ).ExpiresAt;
+		TryCacheOktaSession( user, org, sessionId, expiresAt );
```
cfbao, line 243: The only reason you passed in the `org` parameter in addition to `orgBaseAddress` is to be able to cache based on `org`.

I wonder if we'd be better off with caching by org domain/URL in general.

I don't imagine people actually doing this, but theoretically there's no harm in caching `d2l` and `d2l.okta.com` as the same thing, right?

boarnoah: > The only reason you passed in the org parameter in addition to orgBaseAddress is to be able to cache based on org.

This seems reason enough to cache by orgBaseAddress

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -114,26 +119,142 @@ No config file found. Your Okta session will not be cached.
...
+					Consider running `bmx configure` if you own this machine.
+					""" );
+		return false;
+	}
+
 	private void CacheOktaSession( string userId, string org, string sessionId, DateTimeOffset expiresAt ) {
 		var session = new OktaSessionCache( userId, org, sessionId, expiresAt );
 		var sessionsToCache = ReadOktaSessionCacheFile();
-		sessionsToCache = sessionsToCache.Where( session => session.UserId != userId && session.Org != org )
+		sessionsToCache = sessionsToCache.Where( session => session.UserId != userId || session.Org != org )
```
gord5500, line 257: Spotted we don't support caching same username for multiple orgs. Did we want this?

src/D2L.Bmx/Okta/OktaClient.cs
gord5500: Okay I tried to move everything to `Uri`

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -114,26 +119,142 @@ No config file found. Your Okta session will not be cached.
...
+				var cookies = await page.GetCookiesAsync( orgUrl.AbsoluteUri ).WaitAsync( cancellationTokenSource.Token );
+				if( Array.Find( cookies, c => c.Name == "sid" )?.Value is string sid ) {
+					sessionIdTcs.SetResult( sid );
+				}
+			}
+		} catch( TaskCanceledException ) {
+			consoleWriter.WriteWarning( $"""
+				WARNING: Timed out when trying to create Okta session through Desktop Single Sign-On.
+				Check if the org '{orgUrl}' is correct. If running BMX with elevated privileges,
+				rerun the command with the '--experimental-bypass-browser-security' flag
```
cfbao, line 198: Please avoid mentioning DSSO.

I'd also rather not direct people to use `--experimental-bypass-browser-security` without more qualifications, e.g. 

> if you _have to_ run BMX with elevated privileges and you aren't concerned with the security of {orgUrl}, consider ...

Also, to confirm - admin mode without --no-sandbox triggers `TaskCanceledException`?
That seems surprising.

cfbao: You mentioned before that it's straightforward to detect elevated privileges in all OSes. Can we do that?
Can be in a separate PR though.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -114,26 +119,142 @@ No config file found. Your Okta session will not be cached.
...
+				"""
+			);
+			return null;
+		} catch( TargetClosedException ) {
+			consoleWriter.WriteWarning( """
+				WARNING: Failed to create Okta session through Desktop Single Sign-On as BMX is likely being run
+				with elevated privileges. Rerun the command with the '--experimental-bypass-browser-security' flag.
```
cfbao, line 205: nit: not a fan of hard wrapping text at places that are not natural sentence break point (e.g. period, comma)
If someone runs BMX in a narrower terminal window, it could look weird
e.g.
```
WARNING: Failed to create Okta session through Desktop Single Sign-On
as BMX is likely being run
with elevated privileges. Rerun the command with the '--experimental
-bypass-browser-security' flag.
```

Ideally I'd use shorter sentences, and break the message up at natural places.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -114,26 +119,142 @@ No config file found. Your Okta session will not be cached.
...
+			return null;
+		} catch( Exception ) {
+			consoleWriter.WriteWarning(
+				"WARNING: Unknown error while trying to authenticate with Okta using Desktop Single Sign-On." );
+			return null;
+		}
+
+		if( sessionId is null ) {
+			return null;
+		}
```
cfbao, line 217: nit: all these `return null` in all the `catch` blocks seem unnecessary?
`if( sessionId is null ) { return null; }` should handle them all, right?

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -114,26 +119,142 @@ No config file found. Your Okta session will not be cached.
...
+				+ $"'{StripLoginDomain( user )}' does not match user '{StripLoginDomain( oktaSession.Login )}'." );
+			return null;
+		}
+
+		TryCacheOktaSession( user, orgUrl.Host, sessionId, oktaSession.ExpiresAt );
+		return oktaAuthenticatedClient;
+	}
+
+	private static string StripLoginDomain( string email ) {
+		return email.Contains( '@' ) ? email.Split( '@' )[0] : email;
```
cfbao, line 233: `return email.Split( '@' )[0]` should be all you need, right?

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -114,26 +119,142 @@ No config file found. Your Okta session will not be cached.
...
+	}
+
+	private static bool OktaUserMatchesProvided( string oktaLogin, string providedUser ) {
```
cfbao, line 236: naming nit:
I'd call this
`bool AreEquivalentOktaUsernames( string username1, string username2 )`
* less unnecessary business context in a simple method
* starts with verb as is convention for method names

src/D2L.Bmx/OktaAuthenticator.cs
cfbao: don't think it was intentional

cfbao:
There are still quite a few mentions of DSSO in user-facing messages.

cfbao:
yeah... I don't think "Desktop Single Sign-On" means anything to most people either.

src/D2L.Bmx/OktaAuthenticator.cs
gord5500: I saw some stackoverflow posts about it but never tried implementing. I can try after yea

gord5500: > Also, to confirm - admin mode without --no-sandbox triggers TaskCanceledException?

Yea it can hang on `browser.NewPageAsync()` and sometimes would throw the other

cfbao (dismissed):

cfbao:
Please update PR description too

gord5500:
@cfbao the review became stale. Do you mind looking over again please?

src/D2L.Bmx/ParameterDescriptions.cs
```diff
@@ -18,4 +18,7 @@ internal static class ParameterDescriptions {
 		Write BMX command to AWS profile, so that AWS tools & SDKs using the profile will source credentials from BMX.
 		See https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html.
 		""";
+
+	public const string ExperimentalBypassBrowserSecurity
+		= "Disable chromium sandbox when running with elevated permissions for Okta Desktop Single Sign-On";
```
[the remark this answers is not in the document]
gord5500: > this flag doesn't (currently) care about elevated permission itself

I found the [WindowsPrincipal](https://learn.microsoft.com/en-us/dotnet/api/system.security.principal.windowsprincipal?view=net-8.0) class that can tell us possibly if they are admin. Going to try that in a separate pr

Linux seems to be based on euid (effective user id) being equal to 0. There is a package that seems to include the syscall but hasn't been updated in awhile. The other approaches look to run a bash process but I don't see anything native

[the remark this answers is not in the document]
cfbao: Sounds good as long as the implementation is simple.
Don't dig too deep if this proves to be troublesome though. This is entirely optional.

cfbao (approved):

PR #481, opened 2024-09-20
ci: add arm builds

gord5500:
### Why

Turns out we don't need the beta github runners for arm64 builds. Following [this](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/cross-compile#linux) ms doc, mac and windows were already solved. If arm64 runners are cheaper they should also just work out of the box to compile to x86

Excluding linux here as there were issues getting it setup with amazonlinux2. We're going to build amazonlinux2 arm64 manually as part of our testing for a new release for now.

### Ticket

VUL-71


gord5500:
Made a release tag and I need to fix the paths because it's zip -> arm64 > BMX.exe

Dockerfile.ubuntu-arm64
```diff
@@ -0,0 +1,49 @@
+FROM ubuntu:22.04 AS build
```
cfbao, line 1: Touched on this in Slack - we should build in AL2
(and even if we do use Ubuntu, it should be the lowest version possible in order to support the most distros & versions we can, not 22.04 which is the latest LTS).

Given the trouble you've already experienced with cross compiling in Linux containers and the consideration that we'll manually build a Linux arm64 build on AL2 when releasing anyway, I'd rather we _don't_ include an automated Ubuntu arm64 build at all.
I'm a little concerned that including this here would incentivise whoever's releasing to skip building & testing an AL2 arm64 build properly.

cfbao (approved):

PR #482, opened 2024-09-24
Check for elevated permissions with automatic login

gord5500:
### Why

The `--experimental-bypass-browser-security` flag is only useful when running as admin. 

I'm not sure though if we would prefer to not completely kill the process as is happening here or just skip over the automatic login area. I lean to killing it so people don't get use to having this flag included regularly

### Ticket

VUL-453

### DIFF
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -154,11 +154,11 @@ internal class OktaAuthenticator(
 		}
 
 		if( !nonInteractive ) {
-			Console.Error.WriteLine( "Attempting to automatically login using Okta Desktop Single Sign-On." );
+			Console.Error.WriteLine( "Attempting to automatically sign in to Okta." );
 		}
 		using var cancellationTokenSource = new CancellationTokenSource( TimeSpan.FromSeconds( 15 ) );
 		var sessionIdTcs = new TaskCompletionSource<string?>( TaskCreationOptions.RunContinuationsAsynchronously );
-		string? sessionId;
+		string? sessionId = null;
 
 		try {
 			using var page = await browser.NewPageAsync().WaitAsync( cancellationTokenSource.Token );
@@ -180,7 +180,7 @@ internal class OktaAuthenticator(
 							await page.GoToAsync( orgUrl.AbsoluteUri ).WaitAsync( cancellationTokenSource.Token );
 						} else {
 							consoleWriter.WriteWarning(
-								"WARNING: Could not authenticate with Okta using Desktop Single Sign-On." );
+								"WARNING: Failed to authenticate with Okta when trying to automatically sign in" );
 							sessionIdTcs.SetResult( null );
 						}
 						return;
@@ -193,23 +193,22 @@ internal class OktaAuthenticator(
 			}
 		} catch( TaskCanceledException ) {
 			consoleWriter.WriteWarning( $"""
-				WARNING: Timed out when trying to create Okta session through Desktop Single Sign-On.
-				Check if the org '{orgUrl}' is correct. If running BMX with elevated privileges,
-				rerun the command with the '--experimental-bypass-browser-security' flag
+				WARNING: Timed out when trying to automatically sign in to Okta. Check if the org '{orgUrl}' is correct.
+				if you have to run BMX with elevated privileges, and aren't concerned with the security of {orgUrl.Host},
+				consider running the command again with the '--experimental-bypass-browser-security' flag.
 				"""
 			);
-			return null;
 		} catch( TargetClosedException ) {
 			consoleWriter.WriteWarning( """
-				WARNING: Failed to create Okta session through Desktop Single Sign-On as BMX is likely being run
-				with elevated privileges. Rerun the command with the '--experimental-bypass-browser-security' flag.
+				WARNING: Failed to automatically sign in to Okta as BMX is likely being run with elevated privileges.
+				if you have to run BMX with elevated privileges, and aren't concerned with the security of {orgUrl.Host},
+				consider running the command again with the '--experimental-bypass-browser-security' flag.
 				"""
 			);
 			return null;
 		} catch( Exception ) {
 			consoleWriter.WriteWarning(
-				"WARNING: Unknown error while trying to authenticate with Okta using Desktop Single Sign-On." );
-			return null;
+				"WARNING: Unknown error occurred while trying to automatically sign in with Okta." );
 		}
 
 		if( sessionId is null ) {
@@ -220,7 +219,7 @@ internal class OktaAuthenticator(
 		var oktaSession = await oktaAuthenticatedClient.GetCurrentOktaSessionAsync();
 		if( !OktaUserMatchesProvided( oktaSession.Login, user ) ) {
 			consoleWriter.WriteWarning(
-				"WARNING: Could not create Okta session using Desktop Single Sign-On as provided Okta user "
+				"WARNING: Could not automatically sign in to Okta as provided Okta user "
 				+ $"'{StripLoginDomain( user )}' does not match user '{StripLoginDomain( oktaSession.Login )}'." );
 			return null;
 		}
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -1,6 +1,5 @@
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
-using System.Text.Json;
 using D2L.Bmx.Okta;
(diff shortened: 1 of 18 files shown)


src/D2L.Bmx/Program.cs
```diff
@@ -21,8 +23,28 @@
 // allow no-sandbox argument for chromium to for passwordless auth with elevated permissions
 var bypassBrowserSecurityOption = new Option<bool>(
 	name: "--experimental-bypass-browser-security",
+	getDefaultValue: () => false,
```
gord5500, line 26: The validator wasn't getting called without this if the flag wasn't included

src/D2L.Bmx/Program.cs
```diff
@@ -282,3 +304,8 @@
 	} )
 	.Build()
 	.InvokeAsync( args );
+
+internal static partial class NativeMethods {
```
gord5500, line 308: Top Level Statement L here?

src/D2L.Bmx/Program.cs
```diff
@@ -282,3 +304,8 @@
...
+	[LibraryImport( "libc", EntryPoint = "geteuid" )]
+	public static partial uint GetPosixEuid();
```
boarnoah, line 310: Do you know if this works on Alpine with musl, AFAICS the musl implementation of libc, is called ``libc`` as far as imports in C etc... are concerned. 

So this might work?

src/D2L.Bmx/Program.cs
```diff
@@ -21,8 +23,28 @@
...
+		isElevated = NativeMethods.GetPosixEuid() == 0;
+	}
+	if( isElevated && !bypass ) {
+		result.ErrorMessage = """
+			Running BMX with elevated permissions is only allowed with the '--experimental-bypass-browser-security' flag.
+			Only include this if you aren't concerned with the security of the host for your Okta organization.
+			""";
+	} else if( !isElevated && bypass ) {
+		result.ErrorMessage
+			= "BMX is not running with elevated permissions, so --experimental-bypass-browser-security must be set to false";
```
boarnoah, line 44: Isn't it better to say, something along the lines of: don't provide ``-experimental-bypass-browser-security``

cfbao: I'd say this is more than a nit

src/D2L.Bmx/Program.cs
```diff
@@ -21,8 +23,28 @@
...
+	bool isElevated = false;
+	if( OperatingSystem.IsWindows() ) {
+		isElevated = new WindowsPrincipal( WindowsIdentity.GetCurrent() ).IsInRole( WindowsBuiltInRole.Administrator );
+	} else if( OperatingSystem.IsLinux() || OperatingSystem.IsMacOS() ) {
+		isElevated = NativeMethods.GetPosixEuid() == 0;
+	}
+	if( isElevated && !bypass ) {
+		result.ErrorMessage = """
+			Running BMX with elevated permissions is only allowed with the '--experimental-bypass-browser-security' flag.
+			Only include this if you aren't concerned with the security of the host for your Okta organization.
```
boarnoah, line 40: nit: I would really like it if we could articulate that:
1. Using BMX to do AWS stuff is perfectly  fine while elevated
2. You only need this flag when you want to do the equivalent of ``bmx login`` as admin
    a. Right call in most situations is to, run ``bmx login`` on your computer once a day without elevated privs
    b. If you insist in running BMX with elevated privs as the first thing of the day, that is when you need the flag.
    
 Not sure how to fit all that in one sentence :P

src/D2L.Bmx/Program.cs
gord5500: Sure I'll reword

src/D2L.Bmx/Program.cs
cfbao: I'd rather we wrap this along with the IsWindows, IsLinux, IsMacOS logic all in a separate file & class.

cfbao: There's a lot happening in Program.cs already

gord5500: Sure

src/D2L.Bmx/Program.cs
cfbao: our existing libc calls work in alpine

src/D2L.Bmx/Program.cs
gord5500: Okay so this checking for permission logic should go in the okta authenticator class then

cfbao: There's also the consideration that... what if this flag is used with AWS credential process, and what if that may be triggered automatically sometimes elevated sometimes not?
This is a real concern for the LMS.

cfbao: in other words, I'm not too sure that we should always fail BMX when this flag is present but the user isn't elevated

gord5500: Yup I'll just move it to the okta authenticator and divert to password if it's not provided and they are admin

cfbao:
@boarnoah
yeah I agree with Adipa here.
BMX should never use `--no-sandbox` when not elevated, but forcing not using the flag when not elevated onto our user might be detrimental to UX.

src/D2L.Bmx/Program.cs
gord5500: Okay I moved it to the authenticator where if bmx is running as admin and no flag provided, we fall back to password. If not admin but flag provided, we set the flag to false so '--no-sandbox' doesn't get included

gord5500: I think that should also address the issue with wording you had Adipa since it's only on attempting to login now

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -144,9 +144,24 @@ private bool TryAuthenticateFromCache(
...
 		bool nonInteractive,
-		bool experimentalBypassBrowserSecurity
+		bool bypassBrowserSecurity
 	) {
-		await using IBrowser? browser = await Browser.LaunchBrowserAsync( experimentalBypassBrowserSecurity );
+
+		bool hasElevatedPermissions = UserPrivileges.HasElevatedPermissions();
+		if( hasElevatedPermissions && !bypassBrowserSecurity ) {
+			consoleWriter.WriteWarning( $"""
+				BMX is being run with elevated privileges and is unable to automatically sign in to Okta.
```
cfbao, line 153: Not in scope, but - some warnings messages are prefixed with "WARNING:" while others are not.
Would be nice to be consistent on this.
I lean towards no prefix, because we already use a different colour.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -144,9 +144,24 @@ private bool TryAuthenticateFromCache(
...
+				If you have to run BMX with elevated privileges, and aren't concerned with the security of {orgUrl.Host},
+				consider running the command again with the '--experimental-bypass-browser-security' flag.
```
cfbao, line 155: Nit:
When this happens, user might just enter their Okta password and move on, so this message is just an "FYI for next time", and they won't want to run this same command again right now even if they have to run BMX elevated and aren't concerned with the security of the Okta site.

Maybe a better message should say "consider using this flag" instead of "consider running the command again".

Another thing that bugs me and it concerns other related warning messages too - if a user can't use DSSO at all, either because their Okta org doesn't support it, or their computer or network condition isn't right, then ideally BMX shouldn't show any DSSO related warning messages at all. I don't think the current implementation does this.
No rush to fix this right now, but I'd like us to do something about this before releasing.

cfbao (dismissed):

src/D2L.Bmx/OktaAuthenticator.cs
gord5500: Hmm yea you're right. I started with including the prefix since they were in the bmx config provider class but have been inconsistent. I see we use `WriteWarning` in other places though without it. I think I'll remove the prefix for this class

src/D2L.Bmx/OktaAuthenticator.cs
gord5500: > I don't think the current implementation does this.
Yea if there was no DSSO implementation for an org, they would end up always seeing some warning messages about it

I think readding the `passwordless` flag could solved this where it is just defaulted to true.

cfbao (approved):

src/D2L.Bmx/OktaAuthenticator.cs
cfbao: more like... `--disable-passwordless` that's default false.
or another flag/config name.
We should probably still rephrase some messages though.
later

boarnoah (approved):

PR #485, opened 2024-10-15
reword user-facing messages to say "passwordless"

cfbao:
### Why

We're going to brand the new feature "passwordless" rather than "auto login".
Good to use consistent language everywhere.

### Ticket

https://desire2learn.atlassian.net/browse/VUL-522

### DIFF
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -150,14 +150,14 @@ internal class OktaAuthenticator(
 		bool hasElevatedPermissions = UserPrivileges.HasElevatedPermissions();
 		if( hasElevatedPermissions && !bypassBrowserSecurity ) {
 			consoleWriter.WriteWarning( $"""
-				BMX is being run with elevated privileges and is unable to automatically sign in to Okta.
-				If you want to automatically sign in, and aren't concerned with the security of {orgUrl.Host},
-				consider using '--experimental-bypass-browser-security' flag.
+				BMX is being run with elevated privileges and is unable to use Okta passwordless authentication.
+				If you want passwordless authentication, and aren't concerned with the security of {orgUrl.Host},
+				consider using the '--experimental-bypass-browser-security' flag.
 				"""
 			);
 			return null;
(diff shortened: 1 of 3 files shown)


src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -205,22 +205,9 @@ async Task GetSessionCookieAsync() {
...
-				Timed out when trying to automatically sign in to Okta. Check if the org '{orgUrl}' is correct.
-				If you have to run BMX with elevated privileges, and aren't concerned with the security of {orgUrl.Host},
-				consider running the command again with the '--experimental-bypass-browser-security' flag.
-				"""
-			);
-		} catch( TargetClosedException ) {
-			consoleWriter.WriteWarning( """
-				Failed to automatically sign in to Okta as BMX is likely being run with elevated privileges.
-				If you have to run BMX with elevated privileges, and aren't concerned with the security of {orgUrl.Host},
-				consider running the command again with the '--experimental-bypass-browser-security' flag.
```
cfbao, line 218: since https://github.com/Brightspace/bmx/pull/482, we check for elevated permission beforehand and short circuit the logic, so messages here are unnecessary

scowing (approved):

src/D2L.Bmx/Program.cs
```diff
@@ -18,7 +18,7 @@
 	name: "--user",
 	description: ParameterDescriptions.User );
 
-// allow no-sandbox argument for chromium to for passwordless auth with elevated permissions
+// allow no-sandbox argument for Chromium to for passwordless auth with elevated permissions
```
boarnoah, line 21: This message seems a bit weirdly written? 

Probably should say something along the lines of how: we need to set explicitly acknowledge we have no sandboxing in elevated permissions by passing this flag

boarnoah (approved):

src/D2L.Bmx/Program.cs
cfbao: ha, I was actually trying to fix this comment, that's why I touched "Chromium" here, but didn't end up completing the job.
will leave it be for now.

PR #486, opened 2024-10-17
extract password auth into a dedicated method

cfbao:
One nit I have about the current `OktaAuthenticator` class is that it puts DSSO and password auth at different abstraction levels.

This PR extracts password auth into its own dedicated private method, similar to DSSO auth.
The main public method now just shows the simple flow of "try cache -> try DSSO -> try password".

### DIFF
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -55,67 +55,27 @@ internal class OktaAuthenticator(
 		}
 
 		var orgUrl = GetOrgBaseAddress( org );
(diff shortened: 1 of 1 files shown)


src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -55,67 +55,27 @@ bool bypassBrowserSecurity
 		}
 
 		var orgUrl = GetOrgBaseAddress( org );
-		var oktaAnonymous = oktaClientFactory.CreateAnonymousClient( orgUrl );
```
cfbao, line 58: a side benefit of wrapping things in dedicated private method is that it exposes and prevents when we declare/define variables too early, like here

gord5500 (approved):

PR #487, opened 2024-10-18
tweak DSSO messages for UX

cfbao:
### Ticket

https://desire2learn.atlassian.net/browse/VUL-506

### DIFF
--- a/src/D2L.Bmx/Browser.cs
+++ b/src/D2L.Bmx/Browser.cs
@@ -25,12 +25,7 @@ public static class Browser {
 		"/opt/microsoft/msedge/msedge",
 	];
 
-	public static async Task<IBrowser?> LaunchBrowserAsync( bool noSandbox = false ) {
-		string? browserPath = GetPathToBrowser();
-		if( browserPath is null ) {
-			return null;
-		}
-
+	public static async Task<IBrowser> LaunchBrowserAsync( string browserPath, bool noSandbox = false ) {
 		var launchOptions = new LaunchOptions {
 			ExecutablePath = browserPath,
 			Args = noSandbox ? ["--no-sandbox"] : []
@@ -39,7 +34,7 @@ public static class Browser {
 		return await Puppeteer.LaunchAsync( launchOptions );
 	}
 
-	private static string? GetPathToBrowser() {
+	public static string? GetPathToBrowser() {
 		if( OperatingSystem.IsWindows() ) {
 			foreach( string windowsPartialPath in WindowsPartialPaths ) {
 				foreach( string environmentVariable in WindowsEnvironmentVariables ) {
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -2,7 +2,6 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using D2L.Bmx.Okta;
 using D2L.Bmx.Okta.Models;
-using PuppeteerSharp;
 
 namespace D2L.Bmx;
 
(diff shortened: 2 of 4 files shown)


src/D2L.Bmx/Browser.cs
```diff
@@ -39,24 +35,27 @@ public static class Browser {
 		return await Puppeteer.LaunchAsync( launchOptions );
 	}
 
-	private static string? GetPathToBrowser() {
+	public static bool TryGetPathToBrowser( [NotNullWhen( returnValue: true )] out string? path ) {
+		path = null;
```
cfbao, line 39: Changes in this method are not needed for the purpose of this task. Just a style change to make it more conventionally C#.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -60,14 +59,21 @@ bool bypassBrowserSecurity
...
-			orgUrl,
-			user,
-			nonInteractive,
-			bypassBrowserSecurity
-		);
-		if( oktaAuthenticated is not null ) {
-			return new( Org: org, User: user, Client: oktaAuthenticated );
+		if( TryGetDssoAuthOptions( nonInteractive, bypassBrowserSecurity, out var dssoAuthOptions ) ) {
+			if( !nonInteractive ) {
+				Console.Error.WriteLine( "Attempting Okta passwordless authentication..." );
```
cfbao, line 64: `TryGetDssoAuthOptions` contains early validation for DSSO/browser process. It's split out mainly to better control when we print the "attempting..." and "falling back to..." messages.

Some points in consideration:

* When we do launch a browser, it takes a couple seconds, so the "attempting passwordless..." message is worthwhile to print
* If DSSO fails after a browser is launched and the "attempting" message is printed, it makes sense to print another message indicating that we're now switching to password auth
* It only makes sense to print the "fallback" message if the "attempting" message is printed
* It only makes sense to print the "attempting" message if we're actually launching a browser, because deciding whether to launch the browser is very fast (synchronous code)
* Therefore, it makes sense to control when we print both of these messages in the same code area

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -99,35 +105,64 @@ private bool TryAuthenticateFromCache(
...
+		bool bypassBrowserSecurity,
+		[NotNullWhen( returnValue: true )] out DssoAuthOptions? options
 	) {
+		options = null;
+
+		bool printDebugMessage =
+			BmxEnvironment.IsDebug
+			|| (
+				// if the user passed --experimental-bypass-browser-security, they've explicitly expressed interest in the browser/DSSO process
+				bypassBrowserSecurity
```
cfbao, line 121: Aside:
I'm actually wondering if we can afford to not have this flag at all, and just always use `--no-sandbox` when elevated.
BMX has always trusted the org Okta URL. I mean, it sends the user's Okta password (a super sensitive piece of data) to that URL after all, and it never had any warnings about that.
Seems weird that we're now asking users to confirm that they trust their own Okta org when we're just visiting the org homepage.

gord5500: It's an additional risk to the user's machine, right? Seems like it's worth keeping because of that

gord5500 (approved):

src/D2L.Bmx/OktaAuthenticator.cs
cfbao: The question is more that - under what threat model & scenario is it not ok/too risky to load the URL in an unsandboxed browser, but at the same time ok to send the user's Okta password to that very same URL (modulo path difference)?

I think the flag only makes sense if we can find a non-null answer to the question above, and I'm having trouble finding a reasonable threat model & scenario that meets the criteria.

boarnoah: Coming around to the idea, especially since it reduces the mental model re: number of states BMX can be in.

PR #489, opened 2024-10-21
properly time out when waiting on the browser

cfbao:
If Okta redirects us to an unexpected page, we'd be stuck there waiting forever, because the current timeout logic is flawed.


cfbao:
Ideally, we'd also have a per page timeout that's much shorter than the total timeout, e.g. we should never be sitting on an page for 3 seconds without anything happening.

I'll see if this can be implemented easily.

scowing (approved):

PR #492, opened 2024-10-22
add per-page timeout for passwordless auth

cfbao:
We have an overall timeout of 15 seconds for passwordless auth, but also we should not get stuck on a single page for anywhere near that long.

* https://github.com/Brightspace/bmx/pull/492/commits/253142f632cb13c986516480e760baf01d35086f pure refactor
* https://github.com/Brightspace/bmx/pull/492/commits/19d8da990e4f4becccbc1065271ba308f8429267 add per-page timeout

### DIFF
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -114,51 +114,10 @@ internal class OktaAuthenticator(
 		string user,
 		string browserPath
 	) {
-		await using var browser = await browserLauncher.LaunchAsync( browserPath );
-
-		using var cancellationTokenSource = new CancellationTokenSource( TimeSpan.FromSeconds( 15 ) );
-		var sessionIdTcs = new TaskCompletionSource<string?>( TaskCreationOptions.RunContinuationsAsynchronously );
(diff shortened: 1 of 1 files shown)


src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -189,6 +148,65 @@ The provided Okta user '{providedLogin}' does not match the system configured pa
...
+
+		page.Load += ( _, _ ) => _ = OnPageLoadAsync();
+		await page.GoToAsync( orgUrl.AbsoluteUri ).WaitAsync( cancellationTokenSource.Token );
+		return await sessionIdTcs.Task;
+
+		async Task OnPageLoadAsync() {
+			// reset the 3-sec per-page timer on every page load
+			lock( pageTimer ) {
+				pageTimer.Stop();
+				pageTimer.Start();
```
boarnoah, line 176: Don't follow why ``.stop`` needs to be called here?

cfbao: Calling `.Start()` on an already started timer does nothing - it will still fire at the originally scheduled time.
We want to restart the timer here, so `.Stop()` must be called first to reset it.

gord5500 (approved):

PR #493, opened 2024-10-23
don't try passwordless on Linux

cfbao:
Okta doesn't support DSSO on Linux anyway, and in my testing BMX always runs into the "unknown browser error" territory when I run it on Linux with Chrome installed.
Might as well remove the bloat and noise.

### DIFF
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -59,7 +59,7 @@ internal class OktaAuthenticator(
 			return new( Org: org, User: user, Client: oktaAuthenticated );
 		}
 
-		if( browserLauncher.TryGetPathToBrowser( out string? browserPath ) ) {
+		if( !OperatingSystem.IsLinux() && browserLauncher.TryGetPathToBrowser( out string? browserPath ) ) {
 			if( !nonInteractive ) {
 				Console.Error.WriteLine( "Attempting Okta passwordless authentication..." );
 			}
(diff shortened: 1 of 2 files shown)


cfbao:
🤞 I'm hoping the compiler can trim PuppeteerSharp away entirely.

boarnoah (approved):

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -59,7 +59,12 @@ bool ignoreCache
 			return new( Org: org, User: user, Client: oktaAuthenticated );
 		}
 
-		if( browserLauncher.TryGetPathToBrowser( out string? browserPath ) ) {
+		if(
+			// `TryGetPathToBrowser` does return `false` for Linux, but excluding Linux earlier here helps
+			// the compiler trim more unused code (e.g. all of PuppeteerSharp)
+			!OperatingSystem.IsLinux()
```
boarnoah, line 65: I get how that works in theory, how the compiler can suss out that during build time that "hey this check with a static will always be false here, no need to include code only used in this branch" 

It is still mind blowing :P

Not relevant here but, do you know if there is any ways to guarantee that certain things are trimmed or else to throw errors etc...? AFAIK its a bit of a problem with code that relies on compiler optimizations via clever syntax, where it would silently start using the slower path after relatively innocuous looking changes.

gord5500 (approved):

src/D2L.Bmx/OktaAuthenticator.cs
cfbao: I don't know of a way to do that in general.

In the specific case of OS dependent code though, using [preprocessor directives](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives) is a common way to do it, e.g.
https://stackoverflow.com/questions/30153797/c-sharp-preprocessor-differentiate-between-operating-systems
The .NET runtime and core libraries themselves use this a lot.

I tried this a bit in the early days of .NET BMX, but couldn't get the dev experience I wanted.

AnimalXing (approved):

PR #495, opened 2024-10-28
disable passwordless for macOS

cfbao:


gord5500 (approved):

boarnoah (approved):
Unfortunate :(

> Also reconsidering adding --disable-passwordless-auth as suggested before.

Why? Are you thinking in terms of, the fact its not setup in OSX is mostly a D2L thing?

PR #497, opened 2024-12-02
Use spectre.console lib to manage ANSI escape code

AnimalXing:
### Why
Avoid using hardcoded ANSI escape codes to make it more readable.

test

### Ticket

VUL-424

### DIFF
--- a/src/D2L.Bmx/ConsoleWriter.cs
+++ b/src/D2L.Bmx/ConsoleWriter.cs
@@ -1,3 +1,5 @@
+using Spectre.Console;
+
 namespace D2L.Bmx;
 
 internal interface IConsoleWriter {
@@ -11,7 +13,6 @@ internal interface IConsoleWriter {
 // if stdout is redirected (e.g. typical use case for `bmx print`), we won't get any coloured text on stderr.
 // See https://github.com/dotnet/runtime/issues/83146.
 // Furthermore, ANSI escape codes give us greater control over the spread of custom background colour.
-// TODO: use a library to manage ANSI codes and NO_COLOR?
 internal class ConsoleWriter : IConsoleWriter {
 	// .NET runtime subscribes to the informal standard from https://no-color.org/. We should too.
 	// https://github.com/dotnet/runtime/blob/v9.0.0-preview.6.24327.7/src/libraries/Common/src/System/Console/ConsoleUtils.cs#L32-L34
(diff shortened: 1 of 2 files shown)


boarnoah:
Looks nice, can you verify things still function with a native AOT build

build it similar to how we do it in CI (AFAIK PRs don't generate artifacts for u test with)
https://github.com/Brightspace/bmx/blob/845153255c70ad0acad31801ec8d760d5eb05231/.github/workflows/publish.yml#L97-L102

something like
```
          dotnet publish `
            -r win-x64 `
            -o build/x64
```

AnimalXing:
@boarnoah 
> ```
> dotnet publish -r win-x64 -o build/
> ```

Ran this command and tested with its build. It looked fine to me. However, it had some trimming warnings just like from [CI](https://github.com/Brightspace/bmx/actions/runs/12125377971/job/33805206145):

`Warning: /home/runner/work/puppeteer-sharp/puppeteer-sharp/lib/PuppeteerSharp/BrowserFetcher.cs(177): warning IL3000: PuppeteerSharp.BrowserFetcher.GetBrowsersLocation(): 'System.Reflection.Assembly.Location.get' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'. [D:\a\bmx\bmx\src\D2L.Bmx\D2L.Bmx.csproj]` etc.

Is this something to worry about?

boarnoah:
That is a known warning with our choice of puppeter sharp (and present prior to this PR), doesn't look relevant to the new lib you are introducing. 

What we do need to verify is that the coloured text still prints as expected when invoking the native AoT build of BMX which you just built.

===

It isn't unlikely to be an issue in this particular case, but in the past for example we've had issues where the application works as expected when doing ``dotnet run``.

However later when we try the native build (which has had code trimming done as part of building for Native AoT) the application no longer functions etc..

Native AoT stuff has difficulty understanding code is actually used in some cases and isn't dead code that can be trimmed, so best to always check things are functional after introducing a new library.

AnimalXing:
> It isn't likely to be an issue in this particular case, but in the past for example we've had issues where the application works as expected when doing `dotnet run`.
> 
> However later when we try the native build (which has had code trimming done as part of building for Native AoT) the application no longer functions etc..

Thank you for the info! Just verified that the coloured text worked as expected with the Native AoT build, both when the `NO_COLOR` flag was on and off.

cfbao (approved):

PR #502, opened 2025-01-31
build for Linux ARM64 in CI

cfbao:
https://desire2learn.atlassian.net/browse/VUL-648

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -63,18 +63,20 @@ jobs:
     needs: create_release
     if: ${{ !failure() }}
     strategy:
+      fail-fast: false
       matrix:
         include:
           - machine: windows-latest
-            platform: win
+            os: win
             file_name: bmx.exe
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -63,18 +63,20 @@ jobs:
     needs: create_release
     if: ${{ !failure() }}
     strategy:
+      fail-fast: false
       matrix:
         include:
           - machine: windows-latest
-            platform: win
+            os: win
```
cfbao, line 70: nit: "platform" means os + arch

Dockerfile.alpine
```diff
@@ -1,4 +1,6 @@
-FROM amd64/alpine:3.15 AS build
+FROM alpine:3.15 AS build
```
cfbao, line 1: amd64/alpine doesn't have arm64 images

gord5500 (approved):

PR #511, opened 2025-09-26
Write errors to log file

andreeag1:
### Why

Log file added that shows all the exceptions along with their timestamp to make debugging easier.

### Ticket

HOD-3557


src/D2L.Bmx/WriteHandler.cs
```diff
@@ -73,7 +73,8 @@ bool useCredentialProcess
 			?? throw new BmxException( "Invalid output path" );
 		Directory.CreateDirectory( outputFolder );
 		if( !File.Exists( output ) ) {
-			using( File.Create( output ) ) { };
+			using( File.Create( output ) ) { }
+			;
```
andreeag1, line 77: Formatting CI failure again

cfbao: we should remove this semicolon entirely

src/D2L.Bmx/MessageWriter.cs
```diff
@@ -41,13 +48,42 @@ void IConsoleWriter.WriteUpdateMessage( string text ) {
...
+			if( !OperatingSystem.IsWindows() ) {
+				fileOptions.UnixCreateMode = UnixFileMode.UserRead | UnixFileMode.UserWrite;
+			}
+
+			using var stream = new FileStream( _logFilePath, fileOptions );
+			using var writer = new StreamWriter( stream );
+			writer.WriteLine( $"{DateTime.UtcNow:yyyy-MM-dd HH:mm:ss} {text}" );
+			writer.Flush();
+		} catch( Exception ex ) {
+			Console.WriteLine( $"Error writing to log file: {ex.Message}" );
```
boarnoah, line 86:
```suggestion
		string color = _noColor ? "default" : "red";
		_ansiConsole.MarkupLine( $"[{color}] Error writing to log file:  {ex.Message}[/]" );
```


MessageWriter can be invoked in ``bmx print`` or ``write``, in there we can't write misc. text to stdout.

Because we typically use BMX like: ``bmx print | Invoke-Expression`` etc..., we need to make sure we only [pipe](https://www.geeksforgeeks.org/linux-unix/piping-in-unix-or-linux/) text like 
```
$env:AWS_SESSION_TOKEN=123
$env:AWS_ACCESS_KEY_ID=123
$env:AWS_SECRET_ACCESS_KEY=123
```
to the next program.

So we have to write any text intended for the human user elsewhere (to stderr), it will still be visible to the user's terminal but won't get piped.

boarnoah: You may want to double check if that code suggestion I provided works, just copy pasted it from above, may have got syntax wrong

PR #521, opened 2026-02-17
deps: update node.js version from 20 to 24

kspaans:
### Ticket

https://desire2learn.atlassian.net/browse/VUL-1118


gord5500 (approved):

gord5500:
Oops I think you've set the wrong ticket parent though. I think we left `VUL` just for OpenSearch stuff now and `HOD` is everything else (or at least BMX is definitely under it)

boarnoah:
my hot take is we should have gone with VUL, its confusing :P @gord5500 make sure to merge this, since it requires codeowner

PR #524, opened 2026-05-12
check for ACTIVE session id when doing passwordless authentication

gord5500:
### Why

Turning off zscaler gives a bit of time where DSSO will return a session id but not one with status `ACTIVE` and so it is useless 

```
C:\code\bmx\src\D2L.Bmx\bin\Release\net9.0\win-x64\publish [main ≡ +1 ~2 -0 !]> bmx login
Okta org or domain name: d2l (from config file)
Okta username: lgordon (from config file)
Attempting Okta passwordless authentication...
Launching browser: C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe
Creating new browser tab
Navigating to https://d2l.okta.com/
Browser loaded https://d2l.okta.com/
Browser loaded
https://d2l.kerberos.okta.com/login/agentlessDsso/interact?precheck=false&interactionHandle=008f3771y2gka-vRMCmgq3qHDLsB
FHfxLabbDc1wXJ
Browser loaded
https://d2l.okta.com/login/second-factor?fromURI=%2Fapp%2FUserHome%3Fiss%3Dhttps%253A%252F%252Fd2l.okta.com%26login_hint
%3Dlgordon%2540desire2learn.com
Okta passwordless authentication failed.
An active session ID was not returned from Okta.
Falling back to Okta password authentication...
Okta password:
```

### Ticket

[HOD-4575 - BMX > fix passwordless auth](https://desire2learn.atlassian.net/browse/HOD-4575)

### DIFF
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -138,10 +138,9 @@ internal class OktaAuthenticator(
 		var oktaAuthenticatedClient = oktaClientFactory.CreateAuthenticatedClient( orgUrl, sessionId );
 		var oktaSession = await oktaAuthenticatedClient.GetCurrentOktaSessionAsync();
 		if( oktaSession.Status != "ACTIVE" ) {
-			messageWriter.WriteWarning( """
-				Okta passwordless authentication failed.
-				Okta did not provide an expected response.
-				""" );
+			if( BmxEnvironment.IsDebug ) {
+				messageWriter.WriteWarning( "Okta passwordless authentication failed" );
+			}
 			return null;
 		}
 
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -137,6 +137,14 @@ internal class OktaAuthenticator(
 
 		var oktaAuthenticatedClient = oktaClientFactory.CreateAuthenticatedClient( orgUrl, sessionId );
 		var oktaSession = await oktaAuthenticatedClient.GetCurrentOktaSessionAsync();
+		if( oktaSession.Status != "ACTIVE" ) {
+			messageWriter.WriteWarning( """
+				Okta passwordless authentication failed.
+				An active session ID was not returned from Okta.
+				""" );
+			return null;
+		}
+
 		string sessionLogin = oktaSession.Login.Split( "@" )[0];
 		string providedLogin = user.Split( "@" )[0];
 		if( !sessionLogin.Equals( providedLogin, StringComparison.OrdinalIgnoreCase ) ) {
--- a/src/D2L.Bmx/OktaAuthenticator.cs
+++ b/src/D2L.Bmx/OktaAuthenticator.cs
@@ -140,7 +140,7 @@ internal class OktaAuthenticator(
 		if( oktaSession.Status != "ACTIVE" ) {
 			messageWriter.WriteWarning( """
 				Okta passwordless authentication failed.
-				An active session ID was not returned from Okta.
+				Okta did not provide an expected response.
 				""" );
 			return null;
 		}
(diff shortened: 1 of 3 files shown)


gord5500:
> PR #525 created to format .cs files.

Unrelated so I wonder how that got in there in the first place

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -137,6 +137,14 @@ string browserPath
 
 		var oktaAuthenticatedClient = oktaClientFactory.CreateAuthenticatedClient( orgUrl, sessionId );
 		var oktaSession = await oktaAuthenticatedClient.GetCurrentOktaSessionAsync();
+		if( oktaSession.Status != "ACTIVE" ) {
+			messageWriter.WriteWarning( """
+				Okta passwordless authentication failed.
+				An active session ID was not returned from Okta.
+				""" );
```
cfbao: confirming - is this behaviour consistent with when passwordless auth fails for other reasons (e.g. browser asking for password)? like we're printing the same style of warning messages?

I also wonder if "An active session ID was not returned from Okta" is too detailed and unhelpful. Maybe we can reuse a generic failure message for passwordless auth failures (mostly due to users' network environment)

cfbao:
> > PR #525 created to format .cs files.
> 
> Unrelated so I wonder how that got in there in the first place

probably .NET SDK updates

src/D2L.Bmx/OktaAuthenticator.cs
gord5500: There is this [one](https://github.com/Brightspace/bmx/blob/1d9d72a634decff19f6eb92df67829a3553e8698/src/D2L.Bmx/OktaAuthenticator.cs#L152-L153) but the rest are kind of generic and don't give a specific reason and I can see why this wouldn't be super helpful for the user - especially since it doesn't tell them it's probably related to just closing zscaler but that seems weird to add in since it's specific to us. 

I don't think we have a generic output message but I could just reword it to `Okta did not provide expected response`

cfbao: I'm more thinking of these:
https://github.com/Brightspace/bmx/blob/1d9d72a634decff19f6eb92df67829a3553e8698/src/D2L.Bmx/OktaAuthenticator.cs#L220-L224

cfbao: I think "Okta passwordless authentication failed" is appropriate here.
also they're behind the debug flag, so this should be the same.

cfbao: There's no real difference between this inactive session failure and the "browser asking for password" failure, so they should behave the same

gord5500: Ohh okay I can go with that then

cfbao (dismissed):

cfbao:
can you merge in the format change too

cfbao:
also need to merge in main to make vulcan-approval work

gord5500:
Sorry, @cfbao , merging in the autoformat dismissed your review if you don't mind approving again thanks

cfbao (approved):

boarnoah:
Did you observe the session status value with experiments @gord5500 ? 

I wonder what the purpose of a non active session even is, their docs on this part of the API is non-existent AFAICS

gord5500:
> Did you observe the session status value with experiments @gord5500 ?
> 
> I wonder what the purpose of a non active session even is, their docs on this part of the API is non-existent AFAICS

@boarnoah I looked at the docs Chenfeng listed in the ticket https://developer.okta.com/docs/api/openapi/okta-management/management/tags/session/other/getsession

An interesting thing I found and makes sense, is when you use bmx login with zscaler off and it caches the bad session id, if you then turn zscaler back on and run bmx print, the cached session will be usable once you turn zscaler back off

PR #529, opened 2026-06-03
add passwordless timeout config value

gord5500:
### Why

Zscaler can be slow for some people. Bump the default timeout from 15 to 30 seconds and add an optional config value to lower timeout or just disable passwordless entirely. Updating the tests in vulcan-scratch repo too to cover a few scenarios

### Ticket

[HOD-4334 - BMX: support longer timeouts](https://desire2learn.atlassian.net/browse/HOD-4334)

### DIFF
--- a/src/D2L.Bmx/BmxConfig.cs
+++ b/src/D2L.Bmx/BmxConfig.cs
@@ -1,11 +1,5 @@
 namespace D2L.Bmx;
 
-internal static class PasswordlessTimeoutDefaults {
-	public const int Min = 5;
-	public const int Max = 30;
-	public const int Default = 60;
-}
-
 internal record BmxConfig(
 	string? Org,
 	string? User,
--- a/src/D2L.Bmx/BmxConfig.cs
+++ b/src/D2L.Bmx/BmxConfig.cs
@@ -3,7 +3,7 @@ namespace D2L.Bmx;
 internal static class PasswordlessTimeoutDefaults {
 	public const int Min = 5;
 	public const int Max = 30;
-	public const int Default = 30;
+	public const int Default = 60;
 }
 
 internal record BmxConfig(
--- a/src/D2L.Bmx/BmxConfig.cs
+++ b/src/D2L.Bmx/BmxConfig.cs
@@ -1,10 +1,17 @@
 namespace D2L.Bmx;
 
+internal static class PasswordlessTimeoutDefaults {
+	public const int Min = 5;
+	public const int Max = 30;
+	public const int Default = 30;
+}
+
 internal record BmxConfig(
 	string? Org,
 	string? User,
 	string? Account,
 	string? Role,
 	string? Profile,
-	int? Duration
+	int? Duration,
+	int? PasswordlessTimeout
 );
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -47,13 +47,10 @@ internal class BmxConfigProvider(
 		int? passwordlessTimeout = null;
 		if( !string.IsNullOrEmpty( data.Global["passwordlessTimeout"] ) ) {
 			if( !int.TryParse( data.Global["passwordlessTimeout"], out int configTimeout )
-				|| configTimeout < 0
-				|| ( configTimeout > 0 && configTimeout < PasswordlessTimeoutDefaults.Min )
-				|| configTimeout > PasswordlessTimeoutDefaults.Max ) {
+				|| configTimeout < 0 ) {
 				throw new BmxException(
 					"Invalid passwordlessTimeout in config."
-					+ $" Must be 0 (disabled) or between {PasswordlessTimeoutDefaults.Min}"
-					+ $" and {PasswordlessTimeoutDefaults.Max} seconds." );
+					+ " Must be 0 (to disable passwordless authentication) or a positive number of seconds." );
 			}
 			passwordlessTimeout = configTimeout;
 		}
--- a/src/D2L.Bmx/BmxConfigProvider.cs
+++ b/src/D2L.Bmx/BmxConfigProvider.cs
@@ -44,13 +44,28 @@ internal class BmxConfigProvider(
 			duration = configDuration;
 		}
 
+		int? passwordlessTimeout = null;
+		if( !string.IsNullOrEmpty( data.Global["passwordlessTimeout"] ) ) {
+			if( !int.TryParse( data.Global["passwordlessTimeout"], out int configTimeout )
+				|| configTimeout < 0
+				|| ( configTimeout > 0 && configTimeout < PasswordlessTimeoutDefaults.Min )
+				|| configTimeout > PasswordlessTimeoutDefaults.Max ) {
+				throw new BmxException(
+					"Invalid passwordlessTimeout in config."
+					+ $" Must be 0 (disabled) or between {PasswordlessTimeoutDefaults.Min}"
+					+ $" and {PasswordlessTimeoutDefaults.Max} seconds." );
+			}
+			passwordlessTimeout = configTimeout;
+		}
+
 		return new BmxConfig(
 			Org: data.Global["org"],
 			User: data.Global["user"],
 			Account: data.Global["account"],
 			Role: data.Global["role"],
 			Profile: data.Global["profile"],
-			Duration: duration
+			Duration: duration,
+			PasswordlessTimeout: passwordlessTimeout
 		);
 	}
 
@@ -74,6 +89,9 @@ internal class BmxConfigProvider(
 		}
 		if( config.Duration.HasValue ) {
 			data.Global["duration"] = $"{config.Duration}";
+		}
+		if( config.PasswordlessTimeout.HasValue ) {
+			data.Global["passwordlessTimeout"] = $"{config.PasswordlessTimeout}";
 		}
 
 		fs.Position = 0;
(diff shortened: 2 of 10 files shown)


src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -158,20 +171,21 @@ The provided Okta user '{providedLogin}' does not match the system configured pa
...
-		using var cancellationTokenSource = new CancellationTokenSource( TimeSpan.FromSeconds( 15 ) );
+		// cancel if the total time exceeds the configured timeout, including all page loads and retries
+		using var cancellationTokenSource = new CancellationTokenSource( TimeSpan.FromSeconds( timeoutSeconds ) );
 		cancellationTokenSource.Token.Register( () => sessionIdTcs.TrySetCanceled() );
 
-		// cancel if we can't load the first page for 6 seconds
-		using var pageTimer = new System.Timers.Timer( TimeSpan.FromSeconds( 6 ) ) { AutoReset = false };
+		// cancel if we can't load the first page within a derived timeout
+		using var pageTimer = new System.Timers.Timer(
+			TimeSpan.FromSeconds( Math.Min( 6, timeoutSeconds / 2 ) ) ) { AutoReset = false };
```
cfbao: 6 is the min we should wait. Here should be `Math.Max`

cfbao: there's also a 3 sec wait somewhere below that need to be bumped up

src/D2L.Bmx/BmxConfig.cs
```diff
@@ -1,10 +1,17 @@
 namespace D2L.Bmx;
 
+internal static class PasswordlessTimeoutDefaults {
+	public const int Min = 5;
+	public const int Max = 30;
```
cfbao: I wouldn't necessarily cap max at 30. If people want to wait a minute I wouldn't stop them

src/D2L.Bmx/OktaAuthenticator.cs
cfbao: I don't believe the timeout that people actually hit is the 15 total timeout, but rather the 6 and 3 page load timeout.

src/D2L.Bmx/BmxConfig.cs
gord5500: Yeaaa I was wondering what to do for this. I'll bump it

scowing (dismissed):

src/D2L.Bmx/ParameterDescriptions.cs
```diff
@@ -14,6 +14,10 @@ internal static class ParameterDescriptions {
 	public const string NonInteractive = "Run non-interactively without showing any prompts";
 	public const string CacheAwsCredentials =
 		"Enables Cache for AWS tokens. Implied if '--use-credential-process' is supplied";
+	public static readonly string PasswordlessTimeout =
+		"Timeout for Okta passwordless (DSSO) authentication in seconds"
```
cfbao: Do we ever say "DSSO" in user facing messages?

src/D2L.Bmx/ParameterDescriptions.cs
```diff
@@ -14,6 +14,10 @@ internal static class ParameterDescriptions {
...
+		+ $" (0 to disable, {PasswordlessTimeoutDefaults.Min}-{PasswordlessTimeoutDefaults.Max},"
```
cfbao: message not clear to me: 0 to disable the timeout control or to disable passwordless auth?

src/D2L.Bmx/BmxConfig.cs
cfbao: you updated the default not the max?

cfbao (changes requested):

src/D2L.Bmx/BmxConfig.cs
cfbao: also, nit, and I'm not too sure either, but this seems more like "constants" rather than "config"?

cfbao: we already have a constants class

src/D2L.Bmx/BmxConfigProvider.cs
```diff
@@ -44,13 +44,28 @@ public BmxConfig GetConfiguration() {
 			duration = configDuration;
 		}
 
+		int? passwordlessTimeout = null;
+		if( !string.IsNullOrEmpty( data.Global["passwordlessTimeout"] ) ) {
+			if( !int.TryParse( data.Global["passwordlessTimeout"], out int configTimeout )
+				|| configTimeout < 0
+				|| ( configTimeout > 0 && configTimeout < PasswordlessTimeoutDefaults.Min )
```
cfbao: only need a single equality check with 0?

src/D2L.Bmx/BmxConfig.cs
cfbao: was just looking at the config parsing logic - I'm not sure there's much value in setting a non-zero min either?

src/D2L.Bmx/BmxConfigProvider.cs
cfbao: consider also https://github.com/Brightspace/bmx/pull/529/changes#r3352148966

src/D2L.Bmx/ConsolePrompter.cs
```diff
@@ -63,6 +64,25 @@ string IConsolePrompter.PromptPassword() {
 		return duration;
 	}
 
+	int? IConsolePrompter.PromptPasswordlessTimeout() {
+		Console.Error.Write(
+			"Okta passwordless (DSSO) timeout in seconds"
```
cfbao: https://github.com/Brightspace/bmx/pull/529/changes#r3352100232

cfbao: should use `ParameterDescriptions` here?

src/D2L.Bmx/BmxConfig.cs
cfbao: I guess what I'm leaning towards is no min or max.
We can keep the "0 = disable" behaviour (which is a logical conclusion from the config) and check timeout >=0 as a sanity check.
But I'm not seeing value in a non-zero min or any max.
Keep things simpler.

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -158,20 +171,21 @@ The provided Okta user '{providedLogin}' does not match the system configured pa
...
-		using var cancellationTokenSource = new CancellationTokenSource( TimeSpan.FromSeconds( 15 ) );
+		// cancel if the total time exceeds the configured timeout, including all page loads and retries
+		using var cancellationTokenSource = new CancellationTokenSource( TimeSpan.FromSeconds( timeoutSeconds ) );
 		cancellationTokenSource.Token.Register( () => sessionIdTcs.TrySetCanceled() );
 
-		// cancel if we can't load the first page for 6 seconds
-		using var pageTimer = new System.Timers.Timer( TimeSpan.FromSeconds( 6 ) ) { AutoReset = false };
+		// cancel if we can't load the first page within a derived timeout
+		using var pageTimer = new System.Timers.Timer(
+			TimeSpan.FromSeconds( Math.Max( 6, timeoutSeconds / 2 ) ) ) { AutoReset = false };
```
cfbao: minor, but might be better to avoid integer arithmetic and truncation behaviour
e.g.
5/2 = 2
vs
5/2.0 = 2.5
25% difference

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -194,7 +208,7 @@ async Task OnPageLoadAsync() {
 			lock( pageTimer ) {
 				pageTimer.Stop();
 				// we give the first page 6 sec to load, but 3 sec is probably enough for subsequent pages
-				pageTimer.Interval = 3000;
+				pageTimer.Interval = Math.Max( 3, timeoutSeconds / 4 ) * 1000;
```
cfbao: I'm maybe paranoid, but I wonder if any page might load really slow and 7 sec isn't enough... especially like, what if the first page isn't the slowest one, and it's the 3 page that would take bulk of the time??

thinking - remove the distinction between first page load vs subsequent page load, and just always use `timeoutSeconds / 2.0` for each page load.

src/D2L.Bmx/BmxConfig.cs
gord5500: Sure I can remove the limits. For me I just never saw it succeed below 5 seconds

gord5500: Okay I removed the limits and tried rewording the prompt for setting the value to 0 means disabling passwordless auth. Also the timing changes to always be timeout / 2.0

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -194,7 +208,7 @@ async Task OnPageLoadAsync() {
 			lock( pageTimer ) {
 				pageTimer.Stop();
 				// we give the first page 6 sec to load, but 3 sec is probably enough for subsequent pages
-				pageTimer.Interval = 3000;
+				pageTimer.Interval = Math.Max( 3, timeoutSeconds / 2.0 ) * 1000;
```
cfbao: thinking about these page timeouts again...
I wonder if the 6 vs 3 timeouts minimums are even meaningful now.
Can maybe just set a single page that's half the total timeout when the `Timer` is created, and not change the `Interval` here?

gord5500: Sure I've got it now to just always be set to half the total timeout

cfbao (approved):

PR #532, opened 2026-09-18
use identity engine cookie for passwordless

gord5500:
### Ticket

[VUL-3855 - BMX > Stopgap for auto login failure](https://desire2learn.atlassian.net/browse/VUL-3855)


src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -210,38 +228,42 @@ async Task OnPageLoadAsync() {
...
-				// If the path is '/' with 'sign in' in the title, it means DSSO is not available and we should stop retrying.
-				if( title.Contains( "sign in", StringComparison.OrdinalIgnoreCase ) ) {
-					if( attempt < 3 && url.AbsolutePath != "/" ) {
+			if(
+				url.Host == orgUrl.Host
+				&& title.Contains( "sign in", StringComparison.OrdinalIgnoreCase )
+			) {
+				// DSSO can sometimes take more than one attempt. Both this terminal page and the
+				// transient Agentless DSSO page have "Sign In" titles, so only retry from this path.
+				if( url.AbsolutePath.Equals( "/oauth2/v1/authorize", StringComparison.OrdinalIgnoreCase ) ) {
```
cfbao: In my browser dev tool I'm not seeing this URL as the terminal page.

I suspect you were seeing some race conditions. IIUC, page loads/refreshes don't wait for `await` in this event handler `OnPageLoadAsync`, so you could be seeing "sign in" title from a later loaded page but with URL from an earlier page?

cfbao: `/oauth2/v1/authorize` returns a 302 with no body, so it can't have a title at all.
the terminal page is `/login/agentlessDsso/idx?stateToken=...`

gord5500: Hmm yea I guess we are maybe killing the attempt too early then since when we see that url we retry? I can see where it ends up. That probably would only help with debugging though since this url (seemingly?) only gets hit off vpn? 

Probably a small amount of time but is there an other reason to not exit as early as we could if we know the attempt won't work? I didn't do this deliberately though since that original text check was already there so I'm down for whatever. I am not sure about the title though - maybe it's a Puppeteer thing?

```
Attempting Okta passwordless authentication...
Launching browser: C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe
Creating new browser tab
Navigating to https://d2l.okta.com/
Browser loaded https://d2l.okta.com/app/UserHome with title ''
Browser loaded https://d2l.okta.com/oauth2/v1/authorize with title 'D2L Corporation - Sign In'
Attempt 1 failed. Retry...
Browser loaded https://d2l.okta.com/oauth2/v1/authorize with title 'D2L Corporation - Sign In'
Attempt 2 failed. Retry...
Browser loaded https://d2l.okta.com/oauth2/v1/authorize with title 'D2L Corporation - Sign In'
Okta passwordless authentication is not available
```

src/D2L.Bmx/OktaAuthenticator.cs
```diff
@@ -210,38 +231,53 @@ async Task OnPageLoadAsync() {
 				pageTimer.Start();
 			}
 
+			string title = await page.GetTitleAsync().WaitAsync( cancellationTokenSource.Token );
+			var url = new Uri( page.Url );
 			if( BmxEnvironment.IsDebug ) {
-				messageWriter.WriteWarning( $"Browser loaded {page.Url}" );
+				messageWriter.WriteWarning(
+					$"Browser loaded {url.GetLeftPart( UriPartial.Path )} with title '{title}'"
```
cfbao, line 238: I assume this is to hide the query string which can contain somewhat sensitive parameters?
would be good to include a comment explaining this

cfbao (dismissed):

cfbao (approved):

